PR #1, opened 2023-10-30
Stricter utf8 library validation

zeux:
[Rendered link](https://github.com/luau-lang/rfcs/blob/zeux-utf8-strict/docs/behavior-stricter-utf8-library.md)

### DIFF
--- a/docs/behavior-stricter-utf8-library.md
+++ b/docs/behavior-stricter-utf8-library.md
@@ -1,0 +1,46 @@
+# Stricter utf8 library validation
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


Dekkonot:
> Because this technically changes behavior, due to Hyrum's law we might break some code that really wants to encode surrogates in UTF-8 despite the fact that only a couple functions support this.

I imagine nobody is encoding them deliberately since it's bad behavior. If they are, I'd love to hear from them and talk about what lead to this choice.

vegorov-rbx (approved):

PR #3, opened 2023-10-31
Add RFC for syntax for property access

asajeffrey:


asajeffrey:
Rendered: https://github.com/asajeffrey/rfcs/blob/syntax-property-access-modifiers/docs/syntax-property-access-modifiers.md

asajeffrey:
Moved from https://github.com/luau-lang/luau/pull/284

andyfriesen:
I have a PR in the works that implements the following syntax:

```lua
type A = {x: number}                        -- Read-write
type B = {read x: number}                   -- Read only
type C = {write x: number}                  -- Write only
type D = {read x: string, write x: "hello"} -- Read type and write types are not the same
```

It should be quite rare that a person actually has to write the syntax for `D`.

PR #5, opened 2023-10-31
Add RFC for shared self

asajeffrey:


asajeffrey:
Rendered: https://github.com/asajeffrey/rfcs/blob/shared-self-types/docs/shared-self-types.md

asajeffrey:
Moved from: https://github.com/luau-lang/luau/pull/863

andyfriesen (approved):

PR #6, opened 2023-11-01
RFC: New Require By String Semantics

menarulalam:
[Rendered](https://github.com/Roblox/luau/blob/RFC-for-req-by-string-relative-paths/rfcs/require-by-string-relative-paths.md)

Note: this is one of the two parts that the previous require by string https://github.com/Roblox/luau/pull/969 was split into.

Transferred from https://github.com/luau-lang/luau/pull/1060


docs/require-by-string-relative-paths.md
```diff
@@ -0,0 +1,169 @@
...
+To require a Luau module under the current implementation, we must require it either by relative or absolute path:
+
+#### Relative paths
+
+Modules can be required relative to the requiring file's location in the filesystem (note, this is different from the current implementation, which evaluates all relative paths in relation to the current working directory).
+
+If we are trying to require a module called `MyModule.luau` in `C:/MyLibrary`:
+```lua
+-- From C:/MyLibrary/SomeModule.luau
+local MyModule = require("MyModule")
```
alexmccord, line 52: In the other RFC for require by alias, there are bits of examples that says `local Roact = require("Roact")` which resolves to the package `Roact`, but then you say this is a relative require.

I propose that we don't support this either. Require an explicit `./` or `../`.

If you have some file named `Foo` for which you say `require("Foo")`, and you happen to install some library named `Foo`, you now have to define the _order_ of which one should be required. Is it the file, or the library?

Similarly, if you already have the package named `Bar`, and you added a file named `Bar`, you might end up changing the other `require("Bar")` in the sibling files to resolve to the wrong module.

By demanding an explicit `./`/`../`, you avoid the above problem.

docs/require-by-string-relative-paths.md
```diff
@@ -0,0 +1,169 @@
...
+local MyModule = require("../MyLibrary/MyModule")
+```
+
+Relative paths can begin with `./` or `../`, which denote the directory of the requiring file and its parent directory, respectively. When a relative path does begin with one of these prefixes, it will only be resolved relative to the requiring file. If these prefixes are not provided, path resolution will fallback to checking the paths in the `paths` configuration variable, as described later.
+
+When a require statement is executed directly in a REPL input prompt (not in a file), relative paths will be evaluated in relation to the pseudo-file `stdin`, located in the current working directory. If the code being executed is not tied to a file (e.g. using `loadstring`), executing any require statements in this code will result in an error.
+
+#### Absolute paths
+
+Absolute paths will no longer be supported in `require` statements, as they are unportable. The only way to require by absolute path will be through a explicitly defined paths or aliases defined in configuration files, as described in another [RFC](https://github.com/Roblox/luau/pull/1061).
```
alexmccord, line 70: So we're abolishing absolute requires, which I wasn't expecting to see! I think it's a very positive thing.

The name of this RFC only very specifically talks about `require by string with relative paths`. Maybe rename it to something like `New require by string semantics`?

docs/require-by-string-relative-paths.md
```diff
@@ -0,0 +1,169 @@
...
+
+```lua
+-- Beginning of /Users/JohnDoe/LuauModules/Math/math.luau
+local sqrt = require("../MathHelperFunctions/sqrt")
+```
+
+If we then launched the Luau CLI from the directory `/Users/JohnDoe/Projects/MyCalculator` and required `math.luau` as follows:
+
+```lua
+> local math = require("/Users/JohnDoe/LuauModules/Math/math")
```
alexmccord, line 27:
```suggestion
local math = require("/Users/JohnDoe/LuauModules/Math/math")
```

docs/require-by-string-relative-paths.md
menarulalam: Done!

docs/require-by-string-relative-paths.md
menarulalam: Yep, that makes sense to me. It's also just more readable even if it is unambiguous.

docs/require-by-string-relative-paths.md
alexmccord, line 1: Don't forget to rename file too.

alexmccord (approved):

docs/require-by-string-relative-paths.md
```diff
@@ -61,7 +57,7 @@ local MyModule = require("../MyModule")
 local MyModule = require("../MyLibrary/MyModule")
 ```
 
-Relative paths can begin with `./` or `../`, which denote the directory of the requiring file and its parent directory, respectively. When a relative path does begin with one of these prefixes, it will only be resolved relative to the requiring file. If these prefixes are not provided, path resolution will fallback to checking the paths in the `paths` configuration variable, as described later.
+Relative paths must begin with `./` or `../`, which denote the directory of the requiring file and its parent directory, respectively. When a relative path does begin with one of these prefixes, it will only be resolved relative to the requiring file. If these prefixes are not provided, path resolution will fallback to checking the paths in the `paths` configuration variable, as described later.
```
alexmccord: Oh, sorry! One more thing. We probably should include the rationale that motivated this change.

menarulalam: Added it.

docs/new-require-by-string-semantics.md
```diff
@@ -57,7 +57,7 @@ local MyModule = require("../MyModule")
 local MyModule = require("../MyLibrary/MyModule")
 ```
 
-Relative paths must begin with `./` or `../`, which denote the directory of the requiring file and its parent directory, respectively. When a relative path does begin with one of these prefixes, it will only be resolved relative to the requiring file. If these prefixes are not provided, path resolution will fallback to checking the paths in the `paths` configuration variable, as described later.
+Relative paths must begin with `./` or `../`, which denote the directory of the requiring file and its parent directory, respectively. Disallowing implicit relative requires (`require("Foo")` resolves to `require("./Foo")`) disambiguates how requires will run. When a relative path does begin with one of these prefixes, it will only be resolved relative to the requiring file. If these prefixes are not provided, path resolution will fallback to checking the paths in the `paths` configuration variable, as described later.
```
alexmccord, line 60:
```suggestion
Relative paths must begin with `./` or `../`, which denote the directory of the requiring file and its parent directory, respectively. Disallowing implicit relative requires (`require("Foo")` resolves to `require("./Foo")`) disambiguation of how requires will run. When a relative path does begin with one of these prefixes, it will only be resolved relative to the requiring file. If these prefixes are not provided, path resolution will fallback to checking the paths in the `paths` configuration variable, as described later.
```

alexmccord (approved):

docs/require-by-string-relative-paths.md
BradSharp: I understand the concern with the ambiguity but I think it is at the cost of usability. My preference would be to use prefixes to disambiguate requires but have no-prefix available as a short-hand with a defined order that we perform the lookup in.

Regardless, I don't think this matters for this RFC. We can always start more restrictive and relax things later. The alias RFC can cover this.

JohnnyMorganz:
Little bit late to the party, but a few questions that I don't think were answered by this RFC:

- The file extension is omitted in all examples. Does this mean it will auto lookup for a `.luau` file extension?
- If so, is `.lua` file extension supported? And if yes, what's the precedence? (I assume `.luau` over `.lua`)
- Can you explicitly use a file extension in the path?

menarulalam:
> Little bit late to the party, but a few questions that I don't think were answered by this RFC:
> 
> * The file extension is omitted in all examples. Does this mean it will auto lookup for a `.luau` file extension?
> * If so, is `.lua` file extension supported? And if yes, what's the precedence? (I assume `.luau` over `.lua`)
> * Can you explicitly use a file extension in the path?

So to answer your question, when we require a file or directory, we check these suffixes: `".luau", ".lua", "/init.luau", "/init.lua"`. I believe this also answers your question about including directories. In the future, we could allow for having paths end in `.luau` or `.lua`.

vegorov-rbx:
Looks like the 'path resolution' section was dropped during the split, it defined which files are looked up: https://github.com/luau-lang/luau/pull/969/files#diff-34bfad7a4b99122bb5f52f25fae03ea19422997b0677571b1855e8713abe1d6dR80-R89

PR #7, opened 2023-11-01
RFC: Require by string with aliases

menarulalam:
[Rendered](https://github.com/Roblox/luau/blob/RFC-for-req-by-string-aliases/rfcs/require-by-string-aliases.md)

Note: this is one of the two parts that the previous require by string https://github.com/Roblox/luau/pull/969 was split into.

Migrated from https://github.com/luau-lang/luau/pull/1061

### DIFF
--- a/docs/require-by-string-aliases.md
+++ b/docs/require-by-string-aliases.md
@@ -1,0 +1,251 @@
+# Require by String with Aliases
+
+## Summary
+
+We need to add intuitive alias and paths functionality to facilitate the grouping together of related Luau files into libraries and allow for future package managers to be developed and integrated easily. 
+
+## Motivation
+
+The Roblox engine does not currently support require-by-string. One motivation for this RFC is to consolidate require syntax and functionality between the Roblox engine and the broader Luau language itself.
+
+Luau itself currently supports a basic require-by-string syntax that allows for requiring Luau modules by relative or absolute path. Unfortunately, the current implementation does not support aliases.
+
+## Design
+
+#### Aliases
+
+Aliases can be used to bind an absolute or relative path to a convenient, case-sensitive name that can be required directly.
+
+```json
+"aliases": {
+    "Roact": "C:/LuauModules/Roact-v1.4.2"
+}
+```
+
+Based on the alias map above, you would be able to require Roact directly:
+
+```lua
+local Roact = require("Roact")
+```
+
+Or even a sub-module:
+
+```lua
+local createElement = require("Roact/createElement")
+```
+
+Aliases are overrides. Whenever the first component of a path exactly matches a pre-defined alias, it will be replaced before the path is resolved to a file.
+
+##### Versioning
+
+Aliases are simple bindings and aren't concerned with versioning. The intention is to allow package management software to leverage aliases by automatically adding and updating the alias map to reflect a package's dependencies. For manual versioning, a user could always "version" their aliases: `@MyModule-v1`, `@MyModule-v2`, etc.
+
+##### Library root alias
+
+In the past, it has been proposed to define an alias (e.g. "`@`") to represent the root directory of a file's encapsulating library.
+- However, the concept of a "Luau library" and its root directory is not yet rigorously defined in Luau, in terms of folder/file structure.
+- In the future, we may add a `package.json` file or something similar that marks the root directory of a library, but this is outside of the scope of this RFC, which primarily focuses on improving require-by-string.
+- For the time being, this functionality will remain unimplemented for this reason. The alias "`@`" will remain reserved for now, meaning it cannot be overridden.
+
+Of course, users can still use the alias map to explicitly define this behavior with a named alias:
+
+```json
+"aliases": {
+    "src": "../"
+}
+```
+
+#### Paths
+
+Similar to [paths in TypeScript](https://www.typescriptlang.org/tsconfig#paths), we will introduce a `paths` array that can be configured in `.luaurc` files. Whenever a path is passed to `require` and does not begin with `./` or `../`, the path will first be resolved relative to the requiring file. If this fails, we will attempt to resolve paths relative to each path in the `paths` array.
+
+The `paths` array can contain absolute paths, and relative paths are resolved relative to `.luaurc` file in which they appear.
+
+##### Example Definition
+
+With the given `paths` definition (`.luaurc` file located in `/Users/johndoe/Projects/MyProject/src`):
+```json
+"paths": [
+    "../dependencies",
+    "/Users/johndoe/MyLuauLibraries",
+    "/Users/johndoe/MyOtherLuauLibraries",
+]
+```
+
+If `/Users/johndoe/Projects/MyProject/src/init.luau` contained the following code:
+```lua
+local graphing = require("graphing")
+```
+We would search the following directories, in order:
+- `/Users/johndoe/Projects/MyProject/src`
+- `/Users/johndoe/Projects/MyProject/dependencies`
+- `/Users/johndoe/MyLuauLibraries`
+- `/Users/johndoe/MyOtherLuauLibraries`
+
+### Paths array
+
+The `paths` configuration variable provides convenience and allows Luau developers to build complex, well-organized libraries. Imagine the following project structure:
+
+```
+luau-paths-project
+├── .luaurc
+├── dependencies
+│   └── dependency.luau
+└── src
+    └── module.luau
+```
+
+If `.luaurc` contained the following `paths` array:
+```json
+{
+    "paths": ["./dependencies"]
+}
+```
+
+Then, `module.luau` could simply require `dependency.luau` like this:
+```lua
+local dependency = require("dependency")
+
+-- Instead of: require("../dependencies/dependency")
+```
+
+Using the `paths` array allows Luau developers to organize their projects however they like without compromising code readability.
+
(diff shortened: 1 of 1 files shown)


docs/require-by-string-aliases.md
```diff
@@ -0,0 +1,251 @@
...
+
+#### Defining paths/aliases directly in the requiring file
+
+Rather than defining paths/alias maps in an external configuration file, we could alternatively define paths/aliases directly in the files that require them. For example, this could manifest itself through an extension of the `--!` comment syntax or introduce new syntax like `--@<ALIAS> = @<PATH>`.
+```lua
+--@"Roact" = @"C:/LuauModules/Roact-v1.4.2"
+local Roact = require("@Roact")
+
+-- Same as this:
+local Roact = require("@C:/LuauModules/Roact-v1.4.2")
```
alexmccord, line 240: This doesn't look right. It should be without the `@` right?

docs/require-by-string-aliases.md
```diff
@@ -0,0 +1,251 @@
...
+
+#### Aliases
+
+Aliases can be used to bind an absolute or relative path to a convenient, case-sensitive name that can be required directly.
+
+```json
+"aliases": {
+    "Roact": "C:/LuauModules/Roact-v1.4.2"
+}
+```
```
alexmccord, line 23: hmm. I'm actually not sure about case-sensitivity! Paths on Windows and Mac are not case sensitive, but they are on Linux. What do other languages do?

docs/require-by-string-aliases.md
```diff
@@ -0,0 +1,251 @@
...
+
+Based on the alias map above, you would be able to require Roact directly:
+
+```lua
+local Roact = require("Roact")
```
alexmccord, line 28: Why are we dropping the `@` prefix?

docs/require-by-string-aliases.md
alexmccord: I think we should be case insensitive across all platforms, _including Linux_. If we, or the community, develops a package manager, the name of the package they wish to install are going to be case insensitive, so it makes sense for the require by alias to be insensitive.

Otherwise you run into security issues because you typed in the wrong capitalization.

alexmccord: Similarly, we want the name of the library to match the name everywhere except by case, because otherwise you have problems found with `pip`, e.g. `No module named 'yaml'` -> `pip install yaml` -> `cannot find yaml` -> google the name of the package -> `pip install pyyaml`.

docs/require-by-string-aliases.md
menarulalam: Oh did not realize that section was needed in this part. Whoops, fixing it.

docs/require-by-string-aliases.md
menarulalam: By name do you mean the key in the alias map? If so that's user chosen.

alexmccord: Right, it _can be_ user-chosen, but I imagine that if a package management tool exists, it also needs to modify this alias map, right? There's no other way around it. The point I was trying to get across is, the mismatch of "name of the missing module" and "name of the module you need to install" should be exactly the same, _except_ by case.

menarulalam: 1. Ok if we see this alias map primarily being modified by a package manager then I agree with 1. 
2. I don't see a good way to resolve this. Ideally the package manager assigns it the proper name, but inferring the alias from the path has its problems. Maybe I am not understanding your suggestion. What does "we want the name of the library to match the name everywhere except by case" mean?

menarulalam: For posterity, we have decided to make alias names case insensitive and restrict the character set to letters, numbers, and the characters `._-`

docs/require-by-string-aliases.md
alexmccord: Still need to fix the require strings here.

docs/require-by-string-aliases.md
```diff
@@ -0,0 +1,278 @@
...
+```
+
+Or even a sub-module:
+
+```lua
+local createElement = require("Roact/createElement")
+```
+
+Aliases are overrides. Whenever the first component of a path exactly matches a pre-defined alias, it will be replaced before the path is resolved to a file. Alias names are also restricted to the charset `[A-Za-z0-9.-_]`.
```
alexmccord, line 37: Need to escape `-` here, otherwise it is parsed as a range of the byte of `.` (46) and the byte of `_` (95) which obviously is not what you want.

```suggestion
Aliases are overrides. Whenever the first component of a path exactly matches a pre-defined alias, it will be replaced before the path is resolved to a file. Alias names are also restricted to the charset `[A-Za-z0-9.\-_]`.
```

menarulalam: Oh whoops, thanks!

docs/require-by-string-aliases.md
```diff
@@ -0,0 +1,278 @@
...
+
+The Roblox engine does not currently support require-by-string. One motivation for this RFC is to consolidate require syntax and functionality between the Roblox engine and the broader Luau language itself.
+
+Luau itself currently supports a basic require-by-string syntax that allows for requiring Luau modules by relative or absolute path. Unfortunately, the current implementation does not support aliases.
+
+## Design
+
+#### Aliases
+
+Aliases can be used to bind an absolute or relative path to a convenient, case-insensitive name that can be required directly.
```
alexmccord, line 17: I find myself going back and forth on case-sensitivity of alias names. It's annoying that Linux and Roblox are case sensitive, but Windows and Mac are not.

This is made even worse by the fact that changing from case sensitive to insensitive is not backward compatible, and case insensitive to sensitive is also not backward compatible. So we have to pick one and get it right.

menarulalam: I think the original justification that aliases are typically packages and thus should be case insensitive makes sense.

alexmccord: Okay, yeah I'm convinced. Let's add this justification into the RFC.

menarulalam: Done!

alexmccord (approved):

docs/require-by-string-aliases.md
```diff
@@ -0,0 +1,278 @@
...
+local dependency = require("dependency")
+
+-- Instead of: require("../dependencies/dependency")
+```
+
+Using the `paths` array allows Luau developers to organize their projects however they like without compromising code readability.
+
+### Large-scale projects in Luau
+
+For large-scale Luau projects, we might imagine that every dependency of the project is a Luau project itself. We might use an organizational structure like this to create a clean hierarchy:
```
andyfriesen, line 143: This section looks like a partial duplicate of the section that starts at 220.  Is this here intentionally?

docs/require-by-string-aliases.md
```diff
@@ -0,0 +1,265 @@
...
+
+```json
+"aliases": {
+    "src": "../"
+}
+```
+
+#### Paths
+
+Similar to [paths in TypeScript](https://www.typescriptlang.org/tsconfig#paths), we will introduce a `paths` array that can be configured in `.luaurc` files. Whenever a path is passed to `require` and does not begin with `./` or `../`, the path will first be resolved relative to the requiring file. If this fails, we will attempt to resolve paths relative to each path in the `paths` array.
```
JohnnyMorganz, line 87: > the path will first be resolved relative to the requiring file.

Does this contradict the decision made in #6 ? Or does it mean something different

> Relative paths must begin with `./` or `../`, which denote the directory of the requiring file and its parent directory, respectively. Disallowing implicit relative requires (`require("Foo")` resolves to `require("./Foo")`) disambiguation of how requires will run

Same with the provided example below

docs/require-by-string-aliases.md
```diff
@@ -0,0 +1,265 @@
...
+{
+    "aliases": {
+        "Roact": "/Users/johndoe/LuauLibraries/Roact/src"
+    }
+}
+```
+
+Then, we could simply write the following in `module.luau`, and everything would work as intended:
+```lua
+local Roact = require("@Roact")
```
JohnnyMorganz, line 201: Roact in the alias map is pointing to a directory correct? What does the require of a directory point to - is it an `init.luau` file inside that directory? 

(Maybe this is actually a question for #6)

docs/require-by-string-aliases.md
```diff
@@ -0,0 +1,265 @@
...
+└── src
+    └── module.luau
+```
+
+For example, if we wanted to require `Roact` in `module.luau`, we could add the following alias to `.luaurc`:
+
+```json
+{
+    "aliases": {
+        "Roact": "/Users/johndoe/LuauLibraries/Roact/src"
```
JohnnyMorganz, line 194: Can an alias point to a file? Or only a directory?

docs/require-by-string-aliases.md
menarulalam: I guess technically this is still true. However the implication is if the path at first doesn't resolve  relative to the requiring file, there will be more steps after, which is not the case anymore as per that decision.

docs/require-by-string-aliases.md
menarulalam: Yes you can require directories.

PR #10, opened 2023-11-08
Attributes for functions

Dekkonot:
Rendered view [here](https://github.com/Dekkonot/luau-rfcs/blob/rfc-attributes/docs/syntax-attributes-functions.md).

A proposal for attributes for functions. Attributes take the form of `@identifier` and control the behavior of the language a bit more directly than we currently can. The scope of them is limited to functions right now due to feedback on prior RFCs of this nature.

The syntax as it's proposed doesn't include parameters for attributes, which might upset some people!

My justification is that I started thinking about it and I don't know what the limitations for those need to be right now. The easy answer is "literals and identifiers" but I don't have any reason to feel that way, it's just what I think is the right answer. To avoid making the *wrong* choice, I'm just not proposing anything related to parameters with this RFC. I'm open to discussion about adding them to this RFC, but I'd like that discussion to be based on actual use cases rather than speculative ones.

The good news is that most attributes I can think of work just fine without parameters, so it's no huge tragedy at this moment if they're omitted.

### DIFF
--- a/docs/syntax-attributes-functions.md
+++ b/docs/syntax-attributes-functions.md
@@ -79,12 +79,12 @@ Most of the features attributes can support could be implemented by refining the
 
 Attributes are distinct from decorators in that they are compiler defined and thus cannot be defined during runtime. An alternative use for this syntax may be user-defined decorators as in JavaScript. If the language provided special built-in decorators, this could serve a very similar purpose. This is not rejected on merit, it's simply that attributes have a more obvious path to implementation.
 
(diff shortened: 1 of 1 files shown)


Ukendio:
From the looks of it, the suggestion for attributes only applies to named functions and not anonymous functions which would eliminates any use case where you would like to inline or have native codegen on callbacks that you know will be static or trivially inline-able.

Dekkonot:
That raises a few questions I don't have good answers to. I don't know if anonymous functions can currently be inlined currently, as an example. Also, intellisense based attributes (like `@deprecated` don't make a lot of sense on anonymous functions.

Do you have some examples of languages that have attributes (or decorators, since the basic principle is the same) that can go on whatever their equivalent to anonymous functions are?

aviralg (approved):
We have started working on this.

PR #11, opened 2023-11-30
RFC: Exact Table Types

aatxe:
[Rendered](https://github.com/aatxe/luau-rfcs/blob/exact-table-types/docs/exact-table-types.md)

Migrated from https://github.com/luau-lang/luau/pull/624.

### DIFF
--- a/docs/exact-table-types.md
+++ b/docs/exact-table-types.md
@@ -1,0 +1,250 @@
+# Exact Table Types
+
+## Summary
+
+In Luau, we currently support two kinds of table types: sealed, inexact table types (the kind you
+use as parameter types) and unsealed, exact table types (the kind the type inference engine uses
+when checking table literals and subsequent assignments to them in the same scope). The RFCs for
+[unsealed table types][unsealed-tables] use language like _precision_, but is unclear about their
+exact meaning. This RFC aims to clarify the language and syntax around different facets of table
+types, and proposes the introduction of sealed, exact table types and a migration plan to make them
+the default table type.
+
+## Motivation
+
+In the current design of table types, we have conflated two axes of properties, namely whether
+tables are sealed vs. unsealed and whether tables are exact vs. inexact. In the interest of both
+making the language clearer and in making typechecking behavior easier to implement and explain, we
+want to disentangle these concepts. For the sake of precision, this RFC aims to always specify both
+facets, but in practice, we believe programmers will only really be concerned with whether or not a
+table type is exact.
+
+The first distinction, sealed vs. unsealed, deals with state during typechecking. Namely, a _sealed_
+table type is one that will not be changed during typechecking. So, if we have a binding with a
+sealed table type, we know that the type of the binding itself will not change. By contrast,
+_unsealed_ table types are stateful, and an assignment to a binding can grow the set of properties
+on that type. All unsealed tables are sealed when they leave the scope they were created in, so
+unsealed tables should be thought of as a technique used for typechecking imperative-style table
+initialization code.
+
+The second distinction, exact vs. inexact, deals with how the table type relates to other table
+types. In particular, an exact table type indicates that the table has _only_ the properties listed
+in its type, while an inexact table type indicates that the table has _at least_ the properties
+listed in its type. This corresponds to whether or not _width subtyping_ is allowed on the type.
+
+With this language, we currently have two table types which are determined contextually, and have no
+syntactic difference. When a table type is used as the type of a function parameter, e.g. `function
+getx(t : {x: number}) return x end`, we treat is as _sealed_ and _inexact_ meaning the type of the
+parameter will _not_ change, but call sites to the function will permit tables that have _more_
+properties than the ones listed. Within the body of the function however, the typechecker should
+only permit the use of properties explicitly listed in the type since it was annotated. If the type
+was instead inferred, a dual condition should be true --- the properties explicitly listed in the
+type are precisely the ones whose existence the function depends on. Meanwhile, when a table type is
+used for a local definition, it is given an _unsealed_ and _exact_ type meaning we always know
+_precisely_ the set of properties it has, but that the type itself is stateful and assignments are
+allowed to grow the set of properties the table has.
+
+This alludes to a gap of two sorts of table types: __sealed, exact tables_ and _unsealed, inexact
+tables_. We believe the latter is undesirable in the sense that there does not appear to be a
+compelling use case for _forgetting_ some of the structure of the table you're building statefully.
+The main consequence of doing so would be that you would be unable to make further assignments to a
+property after forgetting it was present. The former, however, is the novel proposal for this RFC.
+
+## Design
+
+A _sealed, exact table_ type is effectively the most strict we could be about a particular table. It
+says that the table does not allow width subtyping, and therefore must have _exactly_ the properties
+described by the type, and further says that the type itself is final meaning assignments to missing
+properties would subsequently be rejected. To imagine why this is useful, we can consider a small
+fragment of a hypothetical vector math library:
+
+```lua
+type Vec2 = { x: number, y: number }
+type Vec3 = { x: number, y: number, z: number }
+
+function length(pt: Vec2)
+  return math.sqrt(pt.x^2 + pt.y^2)
+end
+```
+
+If the table type `Vec2` here is `inexact`, we would allow `length` to be applied to a table of type
+`Vec3`. However, in this case, our domain knowledge tells us that `length`'s behavior is incorrect
+for `Vec3` since it does not take into account the `z`-component. While we could design a function
+that instead worked for both, it may be reasonable to instead say that this function only operates
+on the _exact_ table. While this particular example is admittedly a bit contrived,
+[similar proposals][ts-exact-types] for other languages like TypeScript have collected many examples
+where API designers wanted to be able specify exact types. One general pattern of use cases is 
+state machines where you might define a number of types representing individual states and then a
+union over all such types. With only inexact types, the type system permits undesirable combinations
+that can introduce unwanted behavior in code using the state machine.
+
+```luau
+type StartState = {
+  state: "Start",
+  startData: number,
+}
+
+type OtherState = {
+  state: "Other",
+  otherData: string,
+}
+
+type State = StartState | OtherState
+
+-- but we can construct a `State` that is neither a `StartState` nor an `OtherState`.
+local notRight: State = { state: "Start", startData: 42, otherData: "foo" }
+```
+
+Since this proposal would involve having two sorts of sealed table types, _exact_ and _inexact_, and
+their difference is not something we can distinguish contextually in general (as was the case was
+_sealed_ and _unsealed_), we have to propose a syntax for them. This proposal posits that the best
+path forward is to make sealed, exact tables the default table that you get when you write, e.g. `{
+x: number, y: number }` as an annotation on a parameter or in a type alias. To annotate a table as
+inexact and therefore open to having additional properties, we would include an ellipsis in the
+type, e.g. `{ x: number, y: number, ...}`. This was originally proposed as an alternative design in
+the [width subtyping][width-subtyping] proposal.
+
+We propose these syntaxes (`{field: type}` and `{field: type, ...}`) for exact and inexact tables
+respectively because they are fairly lightweight, support anonymous usage, and represent a fairly
+mild change from the syntax that we already have. Alternative syntax options like `interface` in
+TypeScript would represent a bigger deviation from our current syntax of types.
+
+There are a few apparent benefits to this approach. First, it means that a programmer designing an
+API has to make a conscious decision that "yes, in fact, it is sensible to treat all of the unlisted
+properties of this parameter as being unrelated to the behavior of this function." The type checker
+should prefer approaches that, by default, rule out more possible bugs in API usage while keeping
+explicit acceptance of the looser behavior natural and low friction. The motivation for taking this
+approach really being that, as a programmer, you are _forced_ to notice when a program you want to
+succeed produces an error, but you can remain blissfully unaware when programs you want to be
+rejected succeed silently.
+
+Secondly, though relatedly, making all table types inexact-by-default seems to exacerbate confusions
+when considering both the meaning of and the subtyping relations between table types. Consider, for
+example, the inexact table type `{...}` (currently written `{}` in the inexact-by-default world).
+Right away, the syntax of the type matches the syntax of the value form for the empty table. This
+makes it tempting to think of the type as something like "the empty table type" which one might
+expect has a single inhabiting value, namely the empty table. However, with inexact-by-default, this
+`{}` type is really the type of _every single table_ and it functions as a sort of top type for
+tables. If you do mistakenly read it (even momentarily) as the empty table type, you can then be
+surprised by the fact that it is _not_ a subtype of any more specific table, such as `{ x: number,
+y: number }`. This makes sense with the understanding in mind that it is the type of _all_ tables,
+rather than just the empty table, but then you may be surprised that it _is_ a subtype of `{[any]:
+any}`, but is not a subtype of, say, `{[any]: number}`. These sorts of surprises and confusions have
+come up in internal discussions of subtyping for Luau, and the syntax appears to not be doing us any
+favors here.
+
+Thirdly, we argue that it gives the most coherent and intuitive syntax for distinguishing exact vs.
+inexact tables since we believe that being able to distinguish between them and choose the
+appropriate one is important. Ellipses have an already established connotation of omission which
+maps exactly to how sealed, inexact types are _omitting_ some of the properties that the tables
+actually have (since tables with arbitrarily many additional properties can be accepted at an
+inexact table type). The author is unaware of existing syntax that carries the same strong
+connotation, but for exactness. One proposal discussed in the alternatives would be to follow Flow
+in using `{| x: number, y: number |}` to mean a sealed, exact table.
+
+Though we are proposing to make the default interpretation of table _annotations_ sealed and exact,
+we maintain that the sensible default inference behavior is to default to inferring inexact table
+types for unannotated table parameters. Consider the following example:
+
+```lua
+function dostuff(t)
+  ...
+  t.x = 4
+  ...
+  t.y = 5
+  ...
+end
+```
+
+In this case, we expect that `t` will be given the inferred type `{x: number, y: number, ...}`.
+However, if we also passed `t` to another function that requires an exact table, we can propagate
+this requirement to `dostuff`.
+
+```lua
+function dostuff(t)
+  ...
+  t.x = 4
+  ...
+  t.y = 5
+  ...
+  dootherstuff(t)
+  ...
+end
+
+function dootherstuff(pt: {x: number, y: number})
+  ...
+end
+```
+
+So, in this case, `t` would be given the exact table type `{x: number, y: number}` during inference.
+Note that this is still true even if `dostuff` did not refer to some or all of the properties of `t`
+outside of passing it into `dootherstuff`. That is, the type will be the same even if the code was
+as follows:
+
+```lua
+function dostuff(t)
+  return dootherstuff(t)
+end
+
+function dootherstuff(pt: {x: number, y: number})
+  ...
+end
+```
+
+If instead the table is passed to a function requiring an exact table type with _different_
+properties than the ones used in `dostuff`, we will emit an error indicating that they are in
+conflict.
+
+```lua
+function dostuff(t)
+  ...
+  t.z = 4
+  ...
+  ...
+  dootherstuff(t)
+  ...
+end
+
+function dootherstuff(pt: {x: number, y: number})
+  ...
+end
+```
+
+In this case, `t` has two constraints: one indicating that it should be the exact table type
+`{x: number, y: number}` and the other indicating that it should have the property `z`. These
+constraints cannot be reconciled, and so we will get an error at the call site of `dootherstuff`.
+
+## Drawbacks
+
+The main drawback of doing this is that the existing RFC for [sealed table
+subtyping][width-subtyping] was already implemented as of March 1st, meaning that this is ultimately
+a breaking change. As such, it would be important to manage the change as a gradual rollout to
+mitigate problems. Since Luau currently only has sealed, inexact table types we can start by adding
+the syntax for explicit inexact tables (e.g. `{x: number, y: number, ...}`) and implement a warning
+that explains the default behavior will change in the future and to change it to the explicit
+inexact syntax if the programmer would like to keep the current behavior. At a later date, we can
+enable exact-by-default, but should provide an explicit suggestion in the error message to tell the
+programmer that if they _intended_ to allow tables with additional properties, they can change the
+signature of the function to include the `...`.
+
+## Alternatives
+
+There are two primary alternatives to this proposal. First, we could stick with the decision that
+was already made in the [sealed table subtyping][width-subtyping] proposal and leave sealed, inexact
+tables as the default table type. We then have either the choice of doing nothing else or adding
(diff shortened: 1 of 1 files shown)


vegorov-rbx:
What will be the mechanism for the warnings?
Usually it means the linter, but I'm not sure it's powerful enough to find this kind of issue.

aatxe:
> What will be the mechanism for the warnings? Usually it means the linter, but I'm not sure it's powerful enough to find this kind of issue.

It might require making the linter more type-aware, which is perhaps more work than I had anticipated it being when I first wrote this.

karl-police:
This RFC is interesting though, and pursues something I pursued, with setting sealed tables and unsealed tables.

alexmccord:
One more use case for this: iterator of some inexact tables results in `...unknown`

```lua
function f(t: { p: string })
  for i, v in t do -- i : unknown, v : unknwon
  end
end

f({ p = "a", q = 7 }) -- q : number is reachable to the iterator, so v : string does not hold
```

karl-police:
Other use case is to support Module stuff e.g. react-lua

If you create a custom extended property, you'll notice that everyone says that if you do use require, all tables become unmutable... though... you see the issue e.g. when you want ``props`` to be custom and mutable _(for the autocomplete)_.

karl-police:
I really wanted to figure out what happens in the new solver right now. That I wanted to test it.

The result is actually cool, because it seems to be working.


image
image

karl-police:
if you do this

```lua
local test = typeof({}) & typeof({})
```
it actually creates an Intersection and the ConstraintSolver will skip it **hence why the autocomplete never knows**. Which seems to be intended.

Free Types are an exception but they break.

Aslong the tables aren't put into one type, I don't really know if this would work.

andyfriesen:
`{ ... }` and `{}` are great syntaxes for inexact and exact tables, but doing so would break quite a lot of code.

Is there a backwards-compatible syntax that we could use instead?

karl-police:
with ``...`` if Luau ever wanted to have ``...`` "operation" feature from JavaScript, could it be an issue?

karl-police:
> _All unsealed tables are sealed when they leave the scope they were created in, so unsealed tables should be thought of as a technique used for typechecking imperative-style table initialization code._

What if the scope is extended or passed within a sub-scope?

aatxe:
> What if the scope is extended or passed within a sub-scope?

The lexical scope of a binding cannot be extended, so I'm not sure what you mean with the former, but scopes nested within the initializing scope are still within the scope the table was created.

aatxe:
> `{ ... }` and `{}` are great syntaxes for inexact and exact tables, but adopting them would break quite a lot of code.
> 
> 
> 
> Is there a backwards-compatible syntax that we could use instead?

The main alternative is the bars syntax that Flow used for exact tables, but a lot of my hope (and it's seemed validated by the past two years of seeing users use Luau) is that people are by-and-large already expecting table types to be exact, not inexact.

karl-police:
> > What if the scope is extended or passed within a sub-scope?
> 
> The lexical scope of a binding cannot be extended, so I'm not sure what you mean with the former, but scopes nested within the initializing scope are still within the scope the table was created.

Would this go for the example below as well? It's just like the above except the table was compressed into ``require("./name.luau")``

>```lua
>local name = require("./name.luau") -- returns the table, except it's just through "require".
>
>local name2 = name
>function name2:test2() end -- runtime wise this works, type check wise though?
>
>local obj = name2.new()
>```

aatxe:
No, that would not typecheck because the type of `name` is coming from another scope (via require), and `name2` is just aliasing it.

karl-police:
Are there more examples for the "Design" section?

e.g.

```lua
-- this part here is an old solver or new solver bug or something
type T1<E> = { prop1: E }
local tbl = {} :: T1<typeof({})>

-- here I fill in the table with what I want
tbl.prop1 = {}
tbl.prop1.e = "e"
```
In one of the solvers, this currently does something.

It just says that ``prop1`` can include a custom table that is input into it.

How would this be written as with Exact Table Types?

andyfriesen:
Something that might surprise people quite a lot is that intersections of exact tables don't behave the same way at all.

```lua
-- These are all exact tables
type A = {x: number}
type B = {y: string}
type C = {x: number, y: string}

-- D is inexact
type D = {x: number, ...}


type X = A & B -- never!  No value exists that simultaneously satisfies "has x as its sole property" and "has y as its sole property"
type Y = A & C -- Also never
type Z = A & D -- Same as A
type W = C & D -- Same as C
```

PhoenixWhitefire:
That's actually what I expected to happen initially, based on my understanding of intersections between other types.

docs/exact-table-types.md
```diff
@@ -0,0 +1,250 @@
...
+parameter will _not_ change, but call sites to the function will permit tables that have _more_
+properties than the ones listed. Within the body of the function however, the typechecker should
+only permit the use of properties explicitly listed in the type since it was annotated. If the type
+was instead inferred, a dual condition should be true --- the properties explicitly listed in the
+type are precisely the ones whose existence the function depends on. Meanwhile, when a table type is
+used for a local definition, it is given an _unsealed_ and _exact_ type meaning we always know
+_precisely_ the set of properties it has, but that the type itself is stateful and assignments are
+allowed to grow the set of properties the table has.
+
+This alludes to a gap of two sorts of table types: __sealed, exact tables_ and _unsealed, inexact
```
vegorov-rbx, line 47:
```suggestion
This alludes to a gap of two sorts of table types: _sealed, exact tables_ and _unsealed, inexact
```

docs/exact-table-types.md
```diff
@@ -0,0 +1,315 @@
...
+exactness available. If we add a new syntax for sealed, exact tables instead, the most sensible
+syntax would likely follow Flow in including pipes, so `{| x: number, y: number |}` would be the
+type of a sealed, exact table. This approach does not require a migration path, but would likely
+reflect the belief that the added safety of exact table types is only rarely necessary. The author
+believes it is worth considering, however, that Flow has also been migrating to
+[exact-by-default][flow-exact-by-default] for their object types.
+
+[unsealed-tables]: https://github.com/Roblox/luau/blob/master/rfcs/unsealed-table-literals.md
+[ts-exact-types]: https://github.com/microsoft/TypeScript/issues/12936
+[width-subtyping]: https://github.com/Roblox/luau/blob/master/rfcs/sealed-table-subtyping.md
```
nnullcolumn: this link is to an internal rfc

docs/exact-table-types.md
```diff
@@ -0,0 +1,315 @@
...
+orthogonal to the function using that type, but of course, this assumption is sometimes wrong. It is
+possible still that inexactness is so often necessary that we see little safety benefit from making
+exactness available. If we add a new syntax for sealed, exact tables instead, the most sensible
+syntax would likely follow Flow in including pipes, so `{| x: number, y: number |}` would be the
+type of a sealed, exact table. This approach does not require a migration path, but would likely
+reflect the belief that the added safety of exact table types is only rarely necessary. The author
+believes it is worth considering, however, that Flow has also been migrating to
+[exact-by-default][flow-exact-by-default] for their object types.
+
+[unsealed-tables]: https://github.com/Roblox/luau/blob/master/rfcs/unsealed-table-literals.md
```
nnullcolumn: re: internal rfc

docs/exact-table-types.md
```diff
@@ -0,0 +1,315 @@
...
+
+```lua
+function dostuff(t)
+  ...
+  t.x = 4
+  ...
+  t.y = 5
+  ...
+end
+```
```
nnullcolumn, line 157: while sensible for function arguments, i think it actually makes more sense for runtime table declarations to be exact-by-default. That's a really big thing that I'd like to see no matter how the default annotation semantics turn out.

docs/exact-table-types.md
```diff
@@ -0,0 +1,315 @@
...
+makes it tempting to think of the type as something like "the empty table type" which one might
+expect has a single inhabiting value, namely the empty table. However, with inexact-by-default, this
+`{}` type is really the type of _every single table_ and it functions as a sort of top type for
+tables. If you do mistakenly read it (even momentarily) as the empty table type, you can then be
+surprised by the fact that it is _not_ a subtype of any more specific table, such as `{ x: number,
+y: number }`. This makes sense with the understanding in mind that it is the type of _all_ tables,
+rather than just the empty table, but then you may be surprised that it _is_ a subtype of `{[any]:
+any}`, but is not a subtype of, say, `{[any]: number}`. These sorts of surprises and confusions have
+come up in internal discussions of subtyping for Luau, and the syntax appears to not be doing us any
+favors here.
```
nnullcolumn, line 134: > you can then be surprised by the fact that it is _not_ a subtype of any more specific table, such as `{ x: number, y: number }`.

did you mean to describe the inverse relationship here as surprising? hamburger `{| |}` empty table wouldn't be <: `{ x: number, y: number }`, but `{ x: number, y: number }` <: `{ ... }` and `{ x: number, y: number }` </: hamburger

docs/exact-table-types.md
```diff
@@ -0,0 +1,315 @@
...
+
+The existing RFC for [sealed table subtyping][width-subtyping] was already implemented. This is
+therefore a breaking change. As such, it would be important to manage the change as a gradual rollout to
+mitigate problems. Since Luau currently only has sealed, inexact table types we can start by adding
+the syntax for explicit inexact tables (e.g. `{x: number, y: number, ...}`) and implement a warning
+that explains the default behavior will change in the future and to change it to the explicit
+inexact syntax if the programmer would like to keep the current behavior. At a later date, we can
+enable exact-by-default, but should provide an explicit suggestion in the error message to tell the
+programmer that if they _intended_ to allow tables with additional properties, they can change the
+signature of the function to include the `...`.
```
nnullcolumn, line 293: this section should mention user-defined type functions. i think they would become significantly more confusing to author.

> The coexistence of these two systems in the same language is going to add some mental burden to the developer.

i'm not sure this is necessary to say. is there really a world where we don't end up supporting both (regardless of which becomes the default)?

> Since Luau currently only has sealed, inexact table types we can start by adding the syntax for explicit inexact tables (e.g. `{x: number, y: number, ...}`) and implement a warning that explains the default behavior will change in the future and to change it to the explicit inexact syntax if the programmer would like to keep the current behavior. At a later date, we can enable exact-by-default, but should provide an explicit suggestion in the error message to tell the programmer that if they _intended_ to allow tables with additional properties, they can change the signature of the function to include the `...`.

this migration path isn't great - how do we create exact tables in the meantime? it seems patient and understanding but reads like it'd have to get rushed through in practice.

additionally, exact-by-default feels very inconsistent with the rest of the language. strict's baseline behavior seems to consistently prioritize soundness but as you need precision you **literally** build it from the top down. I think it'd be really good to infer tables with indexers as exact-by-default (if not *always* exact), as well as all runtime declarations. When met with (default) inexact annotations, they'd just promote to inexact.

there's a lot to be said about composability in the language with table annotations today. As someone who has introduced a lot of people to Luau's type system (non-programmers, in some cases), I'd be lying if I said this wasn't a point of confusion sometimes. But I've also seen people pick up a mental model for what table types *actually* mean really quickly, and I think switching the default seriously discourages the learning opportunities that come with that. Are we sure switching the default might not come with a bunch of other (far scarier) mental model / predictability problems for beginners?

docs/exact-table-types.md
```diff
@@ -0,0 +1,315 @@
...
+
+Thirdly, we argue that it gives the most coherent and intuitive syntax for distinguishing exact vs.
+inexact tables since we believe that being able to distinguish between them and choose the
+appropriate one is important. Ellipses have an already established connotation of omission which
+maps exactly to how sealed, inexact types are _omitting_ some of the properties that the tables
+actually have (since tables with arbitrarily many additional properties can be accepted at an
+inexact table type). The author is unaware of existing syntax that carries the same strong
+connotation, but for exactness. One proposal discussed in the alternatives would be to follow Flow
+in using `{| x: number, y: number |}` to mean a sealed, exact table.
+
```
nnullcolumn, line 144: > Ellipses have an already established connotation of omission which maps exactly to how sealed, inexact types are _omitting_ some of the properties that the tables actually have

is "maps exactly" maybe too strong of wording here? personally I think it's usually a bad thing for this syntax that ellipses are so established - in my ideal world, we wouldn't use someone somebody could type at runtime (even by mistake). That's a big reason I like the hamburger syntax, it doesn't translate to anything the language has today which in my experience makes it a lot easier to teach people stuff like "yeah, {| |} means you have exactly those properties". The only case where this isn't true is with the empty table type, but I bet that confusion melts away the first time someone learns about exact syntax in a real script. As someone who teaches luau semi-frequently, I'm not really looking forward to explaining all of the different contexts ellipses are used in and why they mean different things. `T...`, `...T`, `...: T`, `{ T, ... }` (type system), `{ T, ... }` (runtime), `fn(...)` unpacks but `fn(..., foo)` truncates, `...: any` is fine, but `... :: any` isn't, and groups or other following punctuated values truncate, good lord. Admittedly, a lot of this is pre-existing; I've already seen people have a disproportionate amount of issues with ellipses as syntax today. That doesn't mean we should add to that issue.

docs/exact-table-types.md
```diff
@@ -0,0 +1,315 @@
...
+### Subtyping
+
+Width subtyping applies only when the table in the supertype position is exact.
+
+```
+{x: T, y: U, ...} <:  {x: T, ...}
+{x: T, y: U}      <:  {x: T, ...}
+{x: T, y: U}      </: {x: T}
+{x: T, ...}       </: {x: T}
+```
```
nnullcolumn: > Width subtyping applies only when the table in the supertype position is exact.

shouldn't this say "Width subtyping applies only when the table in the subtype position is inexact."?

docs/exact-table-types.md
```diff
@@ -0,0 +1,315 @@
...
+{x: T, ...} </: setmetatable<{}, MT where MT = {__index: MT, x: T, ...}>
+{x: T} </: setmetatable<{...}, MT where MT = {__index: MT, x: T, ...}>
+```
+
+If the metatables match, subtyping works on the table part as usual:
+
+```
+type MT = {__index: MT, x: T}
+setmetatable<{x: T, y: U, ...}, MT> <: setmetatable<{x: T, ...}, MT>
+```
```
nnullcolumn, line 244: i don't see why we should carve something out just for metatables. is there a disadvantage i'm missing to following the regular table rules?

docs/exact-table-types.md
```diff
@@ -0,0 +1,315 @@
...
+type X = A & B -- never.  No value exists that simultaneously satisfies "has x as its sole property" and "has y as its sole property"
+type Y = A & C -- never
+type Z = A & D -- {x: number}
+type W = C & D -- {x: number, y: string}
+```
+
+Luau presently offers no set-theoretic operator that can combine an `{x: number}` with a
+`{y: string}` to build a `{x: number, y: string}`.  Developers can effect this with a type function.
+We could consider a type-level `..` operator for combining disjoint tables.  eg
+`type Combined = {x: number} .. {y: string}`.
```
nnullcolumn, line 268: this seems simple on the surface but feels like it'd have a bunch of other implications in practice. i feel like the design on this should be expanded a lot if its going to be pursued seriously; udtfs, what happens with `{ ... } .. { ... }`, what happens with `{ ... } .. {|  |}`, is it commutative, how do we expose this in any useful way to users of nonstrict, what about intersections/unions/metatables, wouldn't this be a sensible stringification (or potential conflation!) of the `concat<left, right>` builtin type function, etc

###### apologies for this reference, but.. psychic damage.

docs/exact-table-types.md
```diff
@@ -0,0 +1,315 @@
...
+designers to rule out certain classes of bugs like the one presented with vector types. The
+implicit assumption of width subtyping, after all, is that properties not listed in the type are
+orthogonal to the function using that type, but of course, this assumption is sometimes wrong. It is
+possible still that inexactness is so often necessary that we see little safety benefit from making
+exactness available. If we add a new syntax for sealed, exact tables instead, the most sensible
+syntax would likely follow Flow in including pipes, so `{| x: number, y: number |}` would be the
+type of a sealed, exact table. This approach does not require a migration path, but would likely
+reflect the belief that the added safety of exact table types is only rarely necessary. The author
+believes it is worth considering, however, that Flow has also been migrating to
+[exact-by-default][flow-exact-by-default] for their object types.
```
nnullcolumn, line 310: > The author believes it is worth considering, however, that Flow has also been migrating to [exact-by-default][flow-exact-by-default] for their object types.

> This approach does not require a migration path, but would likely reflect the belief that the added safety of exact table types is only rarely necessary.

re: dynamic typing & consistency with other features which are as permissive as they can reasonably be by default in Luau. Especially function types & lambdas. I don't know a lot about flow, but I feel like a lot more users are *writing* annotations in Flow.

docs/exact-table-types.md
```diff
@@ -0,0 +1,315 @@
...
+tables as the default table type. We then have either the choice of doing nothing else or adding
+a new type syntax for sealed, exact tables. If we do nothing, we are reducing the ability for API
+designers to rule out certain classes of bugs like the one presented with vector types. The
+implicit assumption of width subtyping, after all, is that properties not listed in the type are
+orthogonal to the function using that type, but of course, this assumption is sometimes wrong. It is
+possible still that inexactness is so often necessary that we see little safety benefit from making
+exactness available. If we add a new syntax for sealed, exact tables instead, the most sensible
+syntax would likely follow Flow in including pipes, so `{| x: number, y: number |}` would be the
+type of a sealed, exact table. This approach does not require a migration path, but would likely
+reflect the belief that the added safety of exact table types is only rarely necessary. The author
```
nnullcolumn, line 308: > If we add a new syntax for sealed, exact tables instead, the most sensible
syntax would likely follow Flow in including pipes, so `{| x: number, y: number |}` would be the
type of a sealed, exact table.

i find it likely (re: "how do we create exact tables in the meantime") that pipe syntax will be necessary *no matter what is done*. I would think the first step would be to let people explicitly annotate both ways, maybe switch runtime declarations to exact-by-default, and see how people settle into it. If it turns out people *really* do not want to write `{|  |}`, we could do codemod stuff after that. Goal would be that eventually either `{ ... }` or `{|  |}` would eventually become unnecessary depending on how vibes are after that.

> would likely reflect the belief that the added safety of exact table types is only rarely necessary.

I think "exact table types is only rarely necessary" is probably outright untrue, but I don't think we should optimize the syntax around how rare it is either; When I add syntax to types repeatedly, I'm not thinking about "how often is this necessary", I'm thinking about "Is this necessary"? In that sense, I'm very used to the workflow of adding specificity *when it is necessary*.

docs/exact-table-types.md
```diff
@@ -0,0 +1,315 @@
...
+path forward is to make sealed, exact tables the default table that you get when you write, e.g. `{
+x: number, y: number }` as an annotation on a parameter or in a type alias. To annotate a table as
+inexact and therefore open to having additional properties, we would include an ellipsis in the
+type, e.g. `{ x: number, y: number, ...}`. This was originally proposed as an alternative design in
+the [width subtyping][width-subtyping] proposal.
+
+We propose these syntaxes (`{field: type}` and `{field: type, ...}`) for exact and inexact tables
+respectively because they are fairly lightweight, support anonymous usage, and represent a fairly
+mild change from the syntax that we already have. Alternative syntax options like `interface` in
+TypeScript would represent a bigger deviation from our current syntax of types.
```
nnullcolumn, line 110: > Alternative syntax options like `interface` in TypeScript would represent a bigger deviation from our current syntax of types.

Although it isn't really mentioned much here, I think we should *really* delineate these things. Inexact tables remaining in the limbo they're at now "haha, `{}` is sorta <: table" & "haha, string is sorta <: `{}`" (and friends) isn't good regardless of which direction this RFC takes. The problems are connected but we should talk about them separately and give them the respect and space they need.

docs/exact-table-types.md
```diff
@@ -0,0 +1,315 @@
...
+```
+
+Since this proposal would involve having two sorts of sealed table types, _exact_ and _inexact_, and
+their difference is not something we can distinguish contextually in general (as was the case was
+_sealed_ and _unsealed_), we have to propose a syntax for them. This proposal posits that the best
+path forward is to make sealed, exact tables the default table that you get when you write, e.g. `{
+x: number, y: number }` as an annotation on a parameter or in a type alias. To annotate a table as
+inexact and therefore open to having additional properties, we would include an ellipsis in the
+type, e.g. `{ x: number, y: number, ...}`. This was originally proposed as an alternative design in
+the [width subtyping][width-subtyping] proposal.
```
nnullcolumn, line 105: this isn't really a big deal but I wonder if `{ number, + }` might be better (although it is currently in use, that seems trivial to change)

nnullcolumn (commented):
overall, i'm a lot more happy with this rfc than I had expected I would be. This line in particular was very compelling:
> it means that a programmer designing an API has to make a conscious decision that "yes, in fact, it is sensible to treat all of the unlisted properties of this parameter as being unrelated to the behavior of this function." The type checker should prefer approaches that, by default, rule out more possible bugs in API usage while keeping explicit acceptance of the looser behavior natural and low friction.

I do wonder if there isn't some happy middle-ground where we look at some value somebody's passing **directly** into a function, see that *they* have added extra properties, and be like.. dude, are you sure? As a warning kinda thing. Not that it's explicitly unsafe, but kinda how `foo.bar` errors on missing properties in strict even when it could probably just send out 'unknown' (ignoring roblox instances, i guess)

as a final thought, i do wonder how operations would work on exact tables. for example, is `a.foo` allowed when `a` is a union of `{|  |}` and `{| foo: number |}`?

docs/exact-table-types.md
aatxe: no such thing, it's just a dead link because the repos were moved.

docs/exact-table-types.md
aatxe: no such thing, it's just a dead link because the repos were moved.

docs/exact-table-types.md
aatxe: I don't understand how this is related to the quoted section. Can you explain more?

docs/exact-table-types.md
aatxe: I personally find it much uglier, and it has no established precedence for its meaning in any language.

docs/exact-table-types.md
PhoenixWhitefire:
```suggestion
[unsealed-tables]: https://github.com/luau-lang/rfcs/blob/master/docs/unsealed-table-literals.md
```

docs/exact-table-types.md
PhoenixWhitefire:
```suggestion
[width-subtyping]: https://github.com/luau-lang/rfcs/blob/master/docs/sealed-table-subtyping.md
```

PhoenixWhitefire (commented):
Fix dead links to other RFCs.

docs/exact-table-types.md
nnullcolumn: fair enough

docs/exact-table-types.md
nnullcolumn: resolved in private; the vibe i'm getting is that people writing more annotations and less permissive code by default is an explicit goal of doing this.

docs/exact-table-types.md
aatxe: What is a "runtime table declaration?"

docs/exact-table-types.md
andyfriesen: I'm on the fence as well.  It's not clear to me that it's possible to use such an operator to do anything useful.  I think we need some hands-on-keyboard time with the feature to get a better feel for what they're like.

docs/exact-table-types.md
nnullcolumn: i think all tables constructed with {} syntax should start as exact

PR #12, opened 2023-12-01
Add RFC for buffer varint encoding

bmcq-0:
[Rendered](https://github.com/bmcq-0/function-buffer-varints/blob/master/docs/function-buffer-varints.md)

### DIFF
--- a/docs/function-buffer-varints.md
+++ b/docs/function-buffer-varints.md
@@ -50,14 +50,16 @@ The functions above are inefficient and difficult to understand compared to a li
 The `buffer` library will receive 4 new functions:
 
 ```
-buffer.readleb128(b: buffer, offset: number): number
-buffer.readuleb128(b: buffer, offset: number): number
+buffer.readleb128(b: buffer, offset: number): (number, number)
+buffer.readuleb128(b: buffer, offset: number): (number, number)
 
-buffer.writeleb128(b: buffer, offset: number, value: number): ()
-buffer.writeuleb128(b: buffer, offset: number, value: number): ()
+buffer.writeleb128(b: buffer, offset: number, value: number): number
+buffer.writeuleb128(b: buffer, offset: number, value: number): number
 ```
 
 Since other numbers in the buffer library have unsigned and signed implementations, it also makes sense to include both options for varints.
+
+The functions take arguments similar to other read/write functions in the buffer library. However, they differ in that they return the amount of bytes that were read/written. In readleb128/readuleb128, this is the second return value. Having these functions return the count is necessary as it must be known in order for users to keep track of the buffer offset.
 
 ## Drawbacks
 
--- a/docs/function-buffer-varints.md
+++ b/docs/function-buffer-varints.md
@@ -53,7 +53,7 @@ The `buffer` library will receive 4 new functions:
 buffer.readleb128(b: buffer, offset: number): number
 buffer.readuleb128(b: buffer, offset: number): number
 
-buffer.readleb128(b: buffer, offset: number, value: number): ()
+buffer.writeleb128(b: buffer, offset: number, value: number): ()
 buffer.writeuleb128(b: buffer, offset: number, value: number): ()
 ```
 
--- a/docs/function-buffer-varints.md
+++ b/docs/function-buffer-varints.md
@@ -1,0 +1,70 @@
+# Varint Functions for Buffers
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


docs/function-buffer-varints.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Design
+
+The `buffer` library will receive 4 new functions:
+
+```
+buffer.readleb128(b: buffer, offset: number): number
+buffer.readuleb128(b: buffer, offset: number): number
+
+buffer.readleb128(b: buffer, offset: number, value: number): ()
```
TheGreatSageEqualToHeaven, line 56: Small nitpick, this should be `buffer.writeleb128`

docs/function-buffer-varints.md
```diff
@@ -0,0 +1,70 @@
...
+    if byte < 0x80 then
+      break
+    end
+    bits = bits + 7
+  end
+  return result
+end
+```
+
+The functions above are inefficient and difficult to understand compared to a native implementation. Implementations will also be needed for the corresponding signed functions.
```
vegorov-rbx, line 46: Please use 'library implementation' instead of 'built-in' (implies fastcalls) and 'native' (implies codegen).

And if fastcall/native was implied, we are unlikely to accept this.
Reason for that is that almost everything can be faster with a built-in or with native codegen, so it cannot be the only motivation to having it implemented that way.
It has to be supported by real-life use cases that are suffering by performance that regular C library implementation is able to provide or some other good reason.

WheretIB:
What's confusing about these functions is that you have no idea how many bytes were written or how many were read.
So if you have multiple varints to read, you don't know at which offsets each one of them is placed.

docs/function-buffer-varints.md
```diff
@@ -0,0 +1,72 @@
...
+These functions will write and read (U)LEB-128 variable-width integers (varints) from a buffer.
+
+## Motivation
+
+One of the top use cases for buffers is compressing and serializing data as small as possible. Numbers consume 8 bytes in their raw form but can be serialized in smaller sizes using buffers to reduce size. However, writing code to serialize as the smallest size can be tedious and confusing for less-experienced programmers.  
+
+The following are implementations of ULEB-128 reading/writing in Luau:
+
+```lua
+local function writeuleb128(stream, value)
```
vegorov-rbx, line 22: Please update the implementation examples to match what the proposed functions will do with a library implementation.
Right now they don't compile, incorrectly use the `buffer` library functions and don't return the proposed results.
It would be nice to also convert the argument value to integer number using `bit32` library.

vegorov-rbx:
From the support and not that many problems being reported, we are likely to move forward with this proposal (unless new issues are raised).
Having said that, we need to work to make the specification clean and precise.

docs/function-buffer-varints.md
TheGreatSageEqualToHeaven: @bmcq-0 You can update your read implementation with this, adjust accordingly.

```lua
	local function readByte()
		local byte = buffer_readu8(stream, position)
		position = position + 1
		return byte
	end

	local function readVarInt()
		local result = 0

		for i = 0, 4 do
			local value = readByte()
			result = bit32.bor(result, bit32.lshift(bit32.band(value, 0x7F), i * 7))
			if not bit32.btest(value, 0x80) then
				break
			end
		end

		return result
	end
```

Dekkonot:
It's unclear to me, from reading the RFC, what the arguments to each function do. I understand intuitively because I'm a clever guy, but it feels like a bad idea to merge an RFC where intuitive understanding is a requirement.

Dekkonot:
> > It's unclear to me, from reading the RFC, what the arguments to each function do.
> 
> The arguments are the exact same as other read/write functions in the buffer library. Do you mean the return values?

There's a certain level of irony to me calling myself clever in the same message I made that level of mistake, but that is what I meant yes.

To be more specific since that last comment was poorly written: this RFC assumes prior knowledge of LEB128, which is probably a bad thing. A quick rundown of what it is and why you would use it over any other form of variable-length format would be good to include.

sircfenner:
The number of bytes used needs to be known *before* calling the write* functions because you need to know if the varint will fit within the buffer. Every user would have to implement this non-trivial calculation by themselves, which undermines the three main justifications in the proposal (speed, convenience, and not having to know the specifics of LEB128).

As a side note: the example code in the RFC does not do this size check at the necessary point.

PR #13, opened 2023-12-06
create `github:luau-lang/awesome` repository

junaga:
read the RFC commited


AmaranthineCodices:
While a good idea, this isn't something we want to take up. A collection repository like this should be maintained by the community, not by the Luau team.

PR #14, opened 2023-12-07
Updates to require by string RFCs

menarulalam:
The requirement to add explicitly relative paths in the form of `./path/to/file` broke backwards compatibility, so that is being updated.

### DIFF
--- a/docs/require-by-string-aliases.md
+++ b/docs/require-by-string-aliases.md
@@ -34,7 +34,7 @@ Or even a sub-module:
 local createElement = require("@Roact/createElement")
 ```
 
-Aliases are overrides. Whenever the first component of a path exactly matches a pre-defined alias, it will be replaced before the path is resolved to a file. Alias names are also restricted to the charset `[A-Za-z0-9.\-_]`. We restrict the charset and make them case insensitive because we envision alias names to be primarily used as package names, which tend to be case insensitive and alphanumeric. 
+Aliases are overrides. Whenever the first component of a path exactly matches a pre-defined alias, it will be replaced before the path is resolved to a file. Alias names are also restricted to the charset `[A-Za-z0-9.\-_]`. We restrict the charset and make them case insensitive because we envision alias names to be primarily used as package names, which tend to be case insensitive and alphanumeric. Aliases must also begin with an `@`. 
 
 ### Package management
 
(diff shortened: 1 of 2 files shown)


docs/require-by-string-aliases.md
```diff
@@ -131,7 +131,7 @@ If `.luaurc` contained the following `paths` array:
 
 Then, `module.luau` could simply require `dependency.luau` like this:
 ```lua
-local dependency = require("dependency")
+local dependency = require("@dependency")
```
alexmccord, line 134: I don't think this is right, `@dependency` looks through the `aliases` map, and should never undergo resolution through the `paths`, right? So this is an error.

vegorov-rbx: I agree, and not sure why github has hidden my feedback :( https://github.com/luau-lang/rfcs/commit/0643bda137b1dc990cf782dd6e7f99d173c253ff#r134719538

docs/require-by-string-aliases.md
```diff
@@ -34,7 +34,7 @@ Or even a sub-module:
 local createElement = require("@Roact/createElement")
 ```
 
-Aliases are overrides. Whenever the first component of a path exactly matches a pre-defined alias, it will be replaced before the path is resolved to a file. Alias names are also restricted to the charset `[A-Za-z0-9.\-_]`. We restrict the charset and make them case insensitive because we envision alias names to be primarily used as package names, which tend to be case insensitive and alphanumeric. 
+Aliases are overrides. Whenever the first component of a path exactly matches a pre-defined alias, it will be replaced before the path is resolved to a file. Alias names are also restricted to the charset `[A-Za-z0-9.\-_]`. We restrict the charset and make them case insensitive because we envision alias names to be primarily used as package names, which tend to be case insensitive and alphanumeric. They also must be preceded with an `@` symbol.
```
vegorov-rbx, line 37: Where can I see a discussion of '@' as a symbol used for aliases?

docs/require-by-string-aliases.md
menarulalam: Done.

alexmccord (approved):

vegorov-rbx (approved):

PR #15, opened 2023-12-07
 RFC: Syntax for read- and write- access to properties

andyfriesen:
Rendered: https://github.com/luau-lang/rfcs/blob/read-write-table-syntax/docs/syntax-property-access-modifiers.md

Previous PRs:
* https://github.com/luau-lang/rfcs/pull/3
* https://github.com/luau-lang/luau/pull/284

### DIFF
--- a/docs/syntax-property-access-modifiers.md
+++ b/docs/syntax-property-access-modifiers.md
@@ -1,0 +1,179 @@
+# Syntax for table property access modifiers
+
+## Summary
+
+We need syntax to match the semantics of read-only and writed-only modifiers for table properties and indexers.
+
+## Motivation
+
+See the semantic RFCs for motivation:
+
+* [Read-only table properties](https://github.com/luau-lang/rfcs/blob/master/docs/property-readonly.md)
+* [Write-only table properties](https://github.com/luau-lang/rfcs/blob/master/docs/property-writeonly.md)
+
+## Design
+
+We will use the following syntax for describing a read or a write type of a property:
+
+```lua
+type ReadOnly = { read x: number }
+type WriteOnly = { write x: number }
+```
+
+A property will occasionally be both readable and writable, but using different
+types.  The author will have to duplicate the property name in this case:
+
+```lua
+type Foo = {
+    read p: Animal,
+    write p: Dog
+}
+```
+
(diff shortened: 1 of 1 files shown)


docs/syntax-property-access-modifiers.md
```diff
@@ -0,0 +1,179 @@
...
+or the type:
+
+```lua
+  type Vector2 = { read x : number, read y : number }
+  type Vector2 = { x read : number, y read : number }
+  type Vector2 = { x : read number, y : read number }
+  type Vector2 = { x : number read, y : number read }
+```
+
+The east-coast options are not easy-to-reade with names, but are
```
alexmccord, line 143:
```suggestion
The east-coast options are not easy-to-read with names, but are
```

docs/syntax-property-access-modifiers.md
```diff
@@ -0,0 +1,195 @@
+# Syntax for table property access modifiers
+
+## Summary
+
+We need syntax to match the semantics of read-only and writed-only modifiers for table properties and indexers.
```
vegorov-rbx, line 5:
```suggestion
We need syntax to match the semantics of read-only and write-only modifiers for table properties and indexers.
```

docs/syntax-property-access-modifiers.md
```diff
@@ -0,0 +1,195 @@
...
+     x.pet = Dog.new()
+     return a
+   end
+```
+
+If we are adding the modifier to the property name, we can repeat the name, for example
+
+```lua
+  x : { read pet : Animal, write pet : Dog }
+```
```
vegorov-rbx, line 185: While we have types that are written out in a way that can't be parsed today (ToStringResult::invalid), I think it's reasonable to try to print out in a way that can be put back into the parser.

alexmccord (approved):

potatoprank:
Commenting here just as a observer with no conext on the development of this feature.

Can we add to the documentation what this feature actually does?

Also, a little unclear to me the use case and implications for allowing different read/write types. Seems like unnecessary complexity to support a niche use case I'm not familiar with so my first reaction is 👎

potatoprank:
Specifically, it's unclear to me from the document what it means for a property of some type to be read/write only. I may have missed it. I did kind of tldr; the doc.

PR #20, opened 2024-01-10
RFC: Format Specification for String Interpolation

menarulalam:
We are adding string interpolation format specification to Luau with this change. It supports the same formats as `string.format` and is delimited by a `,`. 

[Rendered view](https://github.com/luau-lang/rfcs/blob/string-interpolation-format-specification/docs/string-interpolation-format-specification.md)

EDIT: Update, this is on hold unless we can find a good delimiter. The problem with `,` and `:`  (e.g. print(`{x:.2f}`) is they are both used in other contexts.

### DIFF
--- a/docs/string-interpolation-format-specification.md
+++ b/docs/string-interpolation-format-specification.md
@@ -10,7 +10,7 @@ Most languages allow for format specifiers such as python f-strings or printf in
 
 ## Design
 
-Under the hood, this will just be syntactic sugar in addition to the current implementation of interpolated strings. Before, interpolated strings were passed to `string.format` with no format was speified, but now we will also pass the optional format specifier. We decided on `,` to act as the delimiter for format specification.
+Under the hood, this will just be syntactic sugar in addition to the current implementation of interpolated strings. Before, interpolated strings were passed to `string.format` with no format was specified, but now we will also pass the optional format specifier. We decided on `,` to act as the delimiter for format specification.
 
 To give some examples, this is how it would look like in code:
 
--- a/docs/string-interpolation-format-specification.md
+++ b/docs/string-interpolation-format-specification.md
@@ -1,0 +1,35 @@
+# Format Specification for String Interpolation
+
+## Summary
+
+This feature allows us to give format specifiers when printing interpolated strings such as `.2f` for a float rounded to two decimal places or `x` for printing in hex. It has all of the format specifiers available in `string.format`. It is delimited by a `,` in interpolated strings. 
+
+## Motivation
+
+Most languages allow for format specifiers such as python f-strings or printf in C. Lua also supports `string.format` but we want to make it easier to use with interpolated strings. 
+
+## Design
+
+Under the hood, this will just be syntactic sugar in addition to the current implementation of interpolated strings. Before, interpolated strings were passed to `string.format` with no format was speified, but now we will also pass the optional format specifier. We decided on `,` to act as the delimiter for format specification.
+
+To give some examples, this is how it would look like in code
+
+```lua
+balance = 100.2035
+print(`You have ${balance,.2f} in your bank account`)
+```
+`You have $100.20 in your bank account`
+
+```lua
+number = 12345
+print(`12345 is 0x{number,x} in hex!`)
+```
+`12345 is 0x3039 in hex!`
+
+## Drawbacks
+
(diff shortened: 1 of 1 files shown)


docs/string-interpolation-format-specification.md
```diff
@@ -0,0 +1,37 @@
...
+
+This feature allows us to give format specifiers when printing interpolated strings such as `.2f` for a float rounded to two decimal places or `x` for printing in hex. It has all of the format specifiers available in `string.format`. It is delimited by a `,` in interpolated strings. 
+
+## Motivation
+
+Most languages allow for format specifiers such as python f-strings or printf in C. Lua also supports `string.format` but we want to make it easier to use with interpolated strings. 
+
+## Design
+
+Under the hood, this will just be syntactic sugar in addition to the current implementation of interpolated strings. Before, interpolated strings were passed to `string.format` with no format was speified, but now we will also pass the optional format specifier. We decided on `,` to act as the delimiter for format specification.
```
alexmccord, line 13:
```suggestion
Under the hood, this will just be syntactic sugar in addition to the current implementation of interpolated strings. Before, interpolated strings were passed to `string.format` with no format was specified, but now we will also pass the optional format specifier. We decided on `,` to act as the delimiter for format specification.
```

docs/string-interpolation-format-specification.md
```diff
@@ -0,0 +1,37 @@
...
+
+This will support most additions that could be made to `string.format` in the future as well. 
+
+## Drawbacks
+
+There are no clear drawbacks to this.
+
+## Alternatives
+
+We have also considered allowing arbitrary format specifiers and not just ones supported by `string.format`. We would allow `__tostring` to take a second format argument, and that specifier get passed into it. We determined that we won't do this for now because it could mess up backwards compatibility on existing `__tostring` calls and a performance regression since we can't assume `tostring` will only have one argument anymore. Also for specifiers that work with `string.format` will be slower than just using `string.format`
```
alexmccord, line 37: There are no backward compatibility concerns I'm aware of? Can you provide an example where such code would break?

menarulalam: Theoretically if a 2nd argument is supplied to tostring, it will be used in this case?

menarulalam:
We can't use `,` as a delimiter as this breaks on a list like this:
```
{
    "import": [
        "A",
        "B",
        "C",
    ]
}

```
any suggestions?

docs/string-interpolation-format-specification.md
Kampfkarren:
```lua
local function getString(thing: T, padding: number?)
    local text = thing.name
    if padding ~= nil then
         text = string.rep(' ', padding) .. text
    end
    return text
end

setmetatable(Thing, {
    __tostring = getString,
})

-- Other calls
print(getString(thing, 4))
```

alexmccord: I guess technically you could write `tostring(x, y)` and it'd work just fine and discards `y` and indeed, adding this feature would turn a dormant bug into a breaking bug. I still consider that an extremely unlikely event except in the case of variadics which makes it look inconspicuous at first. Something as weird as this most certainly would change behavior.

```lua
local t = setmetatable({ 1, 2, 3, 5 }, {
  __tostring = table.concat
})

-- before this RFC:
print(tostring(t)) -- 1235
print(tostring(t, ",")) -- 1235
print(`{t,,}`) -- illegal

-- after this RFC:
print(tostring(t)) -- 1235
print(tostring(t, ",")) -- 1,2,3,5 (behavior changed)
print(`{t,,}`) -- 1,2,3,5
```

But it's not backward incompatible via format specifiers in string interpolation syntax because it's syntactically invalid to even write one right now.

alexmccord: One idea is that we allow `__tostring` to have two arguments, but `tostring` can only forward the first argument?

Wunder-Wulfe:
I really like the idea, and although I think the comma delimiter as fine, as syntactically, there is no valid reason to be using tuples/variadics from within the string, however, I would suggest the following possible alternatives:
```
`my variable is {{variable}d}`
`my variable is {variable!d}` (mimics python with !r and !s to call repr() and str())
`my variable is ${variable!d}` 
```

PR #21, opened 2024-01-12
RFC: Trailing commas in calls, function defs, and function type defs

Kampfkarren:
[Rendered](https://github.com/Kampfkarren/rfcs-1/blob/trailing-commas-in-calls/docs/trailing-commas-in-calls.md)

### DIFF
--- a/docs/trailing-commas-in-calls.md
+++ b/docs/trailing-commas-in-calls.md
@@ -1,0 +1,120 @@
+# Trailing comma support in calls and function definitions
+
+## Summary
+
+Luau will support trailing commas in the following places:
+
+```lua
+local function definition(
+    x: number, -- <-- Here, previously a syntax error
+)
+
+call(
+    a,
+    b, -- <-- Here, previously a syntax error
+)
+
+type Fn = (
+    x: number,
+) -> ()
+```
+
+## Motivation
+
+Currently, trailing commas are supported in several places in Luau.
+
+```lua
+local t = {
+    a,
+    b,
+    c,
+}
+
+-- Which even means...
+f {
+    a,
+    b,
+    c,
+}
+
+type Type = {
+    a: number,
+    b: number,
+}
+```
+
+However, they are not supported in function definitions, types, or calls. This is an unintuitive syntax error and it is obvious what the user intended to do.
+
+## Design
+
+We will make these adjustments to the grammar:
+
+```patch
+funcargs =
+-   '(' [explist] ')'
++   '(' {exp ','} [','] ')'
+    | tableconstructor
+    | STRING
+```
+
+This allows `f(a, b, c,)`.
+
+There is no difference between `f(a(),)` and `f(a())`. Meaning, if `a()` returns `1, 2, 3`, then `f(a(),)` will be equivalent to `f(1, 2, 3)`. This is consistent with tables and `{ a(), }`.
+
+```patch
+parlist =
+-    bindinglist [',' '...']
++    (bindinglist [',' | ',' '...']
+    | '...' [':' (Type | GenericTypePack)]
+```
+
+This results in:
+
+```lua
+function f(
+    a,
+    b, -- New!
+) end
+
+function f(
+    a,
+    ... -- The usual
+) end
+
+function f(
+    a,
+    ..., -- NOT allowed
+)
+```
+
+Trailing comma is not allowed after `...` as it is never correct to put more arguments after the ellipsis. This is consistent with:
+- JavaScript - Supports `function f(a, b,)`, but `function f(...a,) {}` results in "Rest parameter must be last formal parameter".
+
+Though this is not universally agreed upon:
+- Rust's unstable C variadics support `unsafe extern "C" fn f(a: u32, ...,) {}`, though this syntax is not finalized.
+- PHP supports `function f($a, ...$args,) {}`
+- Go *requires* a trailing comma after every argument, including variadics, if split over multiple lines.
+
+```patch
+BoundTypeList =
+    [NAME ':'] Type
+        [',' BoundTypeList]
++       [',']
+    | '...' Type
+```
+
+This results in:
+
+```lua
+type Fn = (x: T,) -> () -- Allowed, new!
+type Fn = (x: T, ...U) -> () -- Allowed
+type Fn = (x: T, ...U,) -> () -- Not allowed
+```
+
+## Drawbacks
+
+There are no interesting drawbacks.
+
+## Alternatives
+
+Supporting `function f(...,)` is easily doable if the rules seems unintuitive.
--- a/docs/trailing-commas-in-calls.md
+++ b/docs/trailing-commas-in-calls.md
@@ -88,7 +88,8 @@ function f(
 ```
 
 Trailing comma is not allowed after `...` as it is never correct to put more arguments after the ellipsis. This is consistent with:
-- JavaScript - Supports `function f(a, b,)`, but `function f(...a,) {}` results in "Rest parameter must be last formal parameter".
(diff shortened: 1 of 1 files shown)


docs/trailing-commas-in-calls.md
```diff
@@ -0,0 +1,120 @@
...
+However, they are not supported in function definitions, types, or calls. This is an unintuitive syntax error and it is obvious what the user intended to do.
+
+## Design
+
+We will make these adjustments to the grammar:
+
+```patch
+funcargs =
+-   '(' [explist] ')'
++   '(' {exp ','} [','] ')'
```
alexmccord, line 55: I think this technically says `exp,,` is okay. Obviously you didn't mean that.
```suggestion
+   '(' exp {',' [exp]} ')'
```

docs/trailing-commas-in-calls.md
```diff
@@ -0,0 +1,121 @@
...
+    c,
+}
+
+type Type = {
+    a: number,
+    b: number,
+}
+```
+
+However, they are not supported in function definitions, types, or calls. This is an unintuitive syntax error and it is obvious what the user intended to do.
```
vegorov-rbx, line 46: > This is an unintuitive syntax error and it is obvious what the user intended to do.

I completely disagree, the way they are disallowed in function arguments is completely intuitive to me.
I would even say that I don't know what user might have intended if they left a trailing ',', as in I agree with the current compiler that they likely forgot to type the last argument.

The motivation for this feature is lacking and weak.

alexmccord: There's two use cases.

With formatting, if the line containing the function name + arguments exceeds the line length, reformat into its multiline variant that's still visually pleasing on the eyes.

```lua
function long_function_name(with, really, long, arguments, some: or_all, might: have_types)
end
```
into:
```
function long_function_name(
  with,
  really,
  long,
  arguments,
  some: or_all,
  might: have_types,
)
end
```

With `git diff`, if you add one more argument, then you will only see that line changed, not the line prior to that argument. Say, if you had:

```lua
  function foo(
    x,
    y
  )
```
then if you add a new argument `z`:
```diff
  function foo(
    x,
+   y,
+   z,
  )
```
as opposed to:
```lua
  function foo(
    x,
    y,
  )
```
then if you add a new argument `z`:
```diff
  function foo(
    x,
    y,
+   z,
  )
```

aatxe: Yeah, personally I'm the sort of person who would require it in a style guide when arguments are split across lines (since it makes reordering lines easier, in addition to the git diff problem). So, I'm personally sympathetic to supporting it. I wouldn't be too upset if we rejected it in single-line definitions too, but it's easier to just always allow the trailing comma obviously.

deviaze: I feel trailing commas should be allowed in both single & multiline calls, for consistency reasons. Luau isn't a whitespace-delimited language, so if you're getting a syntax error that disappears the moment you press Enter, it just seems kinda arbitrary. If tables can get an extra comma without needing a newline, why not function calls?

vegorov-rbx: I don't think consistency argument holds here, simply because function calls and tables are different language constructs with different uses.

deviaze: Tables and function calls are pretty different things—agreed, but since this is a syntax thing, I was looking at it from a broader syntax angle: 

In this case I interpret consistency as visual consistency, 

Using the duck method, a function call and a table definition are both visually comma-separated-lists, placed within some form of bracket (round or curly) delimiters. A user might want to use this fact to organize inputs with linebreaks; a contrived example: 
```lua
local o1, o2, o3, o4, o5, o6, o7 = getSomeClassObjectsAsTuple()
local composite123, composite456 = getCompositesFromTriples(
    o1, o2, o3,
    o4, o5, o6, -- arguments
    -- space to put o7 → o9 to make 3x3 matrix of args later
)
-- a similarly formatted table array definition:

local xTriples = {
    o1, o4, o7,
}
```
I feel that from a visual syntax perspective, both look like comma separated lists before they look like tables or function calls specifically, so ∴ IMHO it makes sense both forms of list adopt the same trailing-comma syntax.

deviaze: Also — just realized I might've been a bit vague the other day with the "arbitrary" 

I meant that in response to @aatxe that if we allow a multiline call to end with a `,`
```lua
call(
  arg1,
  arg2,
  arg3,
)
```
Then I believe we should also allow a single-line call to end with a `,`:
`call(arg1, arg2, arg3, )`

Otherwise, you'd get a whitespace-sensitve syntax error that disappears and reappears with a linebreak, and cases like: 
```lua
call(arg1, arg2, arg3,
) -- not cause SyntaxError?

call(arg1,
arg2, arg3,) -- SyntaxError?
```
the easy solution is just allow all of those cases since style guides/conventions should handle the rest 😅.

vegorov-rbx: > Luau isn't a whitespace-delimited language, so if you're getting a syntax error that disappears the moment you press Enter, it just seems kinda arbitrary.

I want to note that this is not true for function calls:
```
-- no error
local f = attachLog(add)(1, 4)

-- error
local g = attachLog
(add)(1, 4)
```

This is not to suggest that we need more cases like this, I would prefer we avoid them.

vegorov-rbx:
So it's been a month, so I plan to reject this proposal.

While it's a small addition and I'm not even strongly against it, it wasn't shown that it's a good addition to the language and RFC was left inactive.

Going over one thing listed in the RFC and many more things suggested in the comments here and in other places:
* the only reason listed is 'currently unintuitive' and that's too subjective. I suspect it's coming from author experience with Rust, but different people have different experiences and the sense of intuitiveness that comes from that
* consistency argument does not apply to things with different meaning, there is a design reason why tables in particular support a trailing comma (datasets and append mode data logs)
* git-cli-diff-centric motivation is weak for two reasons
    * nobody* writes function arguments in multi-line manner, let's be realistic here
    * diff capturing extra lines is more of a problem of diff tools, GUI can display that as an in-line diff and the problem itself is small
* tool-centric motivation to 'skip check whether or not to output last comma' is also weak, i suggest putting a little effort in writing a simple condition

deviaze:
Late reply here (apologies), but I think I'm seeing (and agreeing with) your point a bit better now.

this should 100% be a bug:
```lua
-- some module
function stuff.repeat_n_times(text : string, n : number?)
-- stuff
end

-- otherwhere
s = require("stuff")
s.repeat_n_times("does stuff", )
```
Therefore I now agree with you wrt. the single-liner calls.

I still believe multiline calls should allow an optional trailing comma, for a small number of math purposes, for type hinting a function's parameters. And especially when dealing with overloads. 

But maybe I write a lot more varargs functions than I'm supposed to, because outside a small benefit to type hinting, those are what I've been thinking about here.

Let's say you're a relatively new Luau-ian doing some `Tween`ing or `CFrame`ing on Roblox and you see some fully hinted `CFrame.new(...)` or `TweenService:Create(...)` calls. They're going way off your screen, so you multiline them. In those cases, (especially CFrames!), I think a trailing comma makes sense, for readability. But again, YMMV.

It's not uncommon in other languages either: Python, js, Rust, & some others allow them too (though that's not much justification in itself ik).

alexmccord:
Yeah, I'm not sure about this RFC anymore. I just had a problem in Python a few hours ago where I removed some arguments from a function call and left it unfinished with a trailing comma. Went and removed the positional parameters from the function, and forgot about going back to the function call. Ended up getting a runtime error, which was kinda okay in Python because that already existed. Doing this in Luau would have meant passing `nil` in and possibly cause havoc.

vegorov-rbx:
Closing this one as there are concerns of the feature and the value of trailing comma in these contexts not being as valuable as in table lists.

PR #23, opened 2024-02-08
RFC: If statement initializers

TheGreatSageEqualToHeaven:
[Rendered](https://github.com/TheGreatSageEqualToHeaven/rfcs/blob/master/docs/syntax-if-statements-initializers.md)

### DIFF
--- a/docs/syntax-if-statements-initializers.md
+++ b/docs/syntax-if-statements-initializers.md
@@ -1,0 +1,93 @@
+# Initializers in if statements
+
+# Summary
+
+Introduce an initializer expression that declares and initializes variables in if statements.
+
+# Motivation
+
+Initializers can improve code clarity and reduce scope pollution by allowing developers to declare variables in the conditions of if statements. ⁤⁤ Declaring variables at the point of use in if statements for the scope of the if statement's block simplifies code, leading to better readability and understanding of program logic. By limiting the scope of variables to the condition's pertaining block, the risk of unintended variable reuse and naming conflicts is reduced. ⁤
+
+The reduced scope pollution improves register space in extreme cases (or auto-generated code) where developers have many variables defined and have to work around register limits by reducing the register size. In some scenarios, especially on Roblox, initializing variables within if statements can lead to improved performance and reduced complexity by avoiding unnecessary calls by developers. ⁤ A common paradigm used by Roblox developers is to use `Instance:FindFirstChild` in their condition and then access it afterwards instead of keeping an existing variable around in an existing scope, polluting it.
+
+# Design
+
+If statements with initializers must match (following the Luau grammar) `'if' 'local' bindinglist ['=' explist] 'then'` and `'local' bindinglist ['=' explist] where exp 'then'` syntax. Parenthesis are also allowed around the initializer for consistency with other expressions. The variables declared by the initializer are only available to the block of that condition and will be undefined to the `elseif` conditions and blocks as well as the `else` block; any code after the if statement won't have the variables defined either.
+
+In the former case, the value of the first declared variable will be checked.
+
+Example:
+
+```lua
+local function foo()
+    return true
+end
+
+if local b = foo() then
+   print(b)
+else
+   print("false")
+end
+```
+
+`Output: true`
+
+In the latter case, the `exp` condition is checked rather than the initializer.
+
+Example:
+
+```lua
+local function foo()
+    return true
+end
+
+if local b = foo() where b == false then
+   print(b)
+else
+   print("true")
+end
+```
+
+`Output: true`
+
+When declaring multiple values inside of a condition, only the first condition will be evaluated.
+
+Example:
+
+```lua
+local function foo()
+    return true, false
+end
+
+if local a,b = foo() then
+    print'Hello World, from Luau!'
+end
+```
+
+`Output: Hello World, from Luau!`
+
+# Drawbacks
+
+Parser recovery may be more fragile due to the `local` keyword.
+
+Initializers increase the complexity of the language syntax and may obscure control flow in complex conditions.
+
+# Alternatives
+
+A different keyword or token can be used in place of `where`.
+
+Introducing a new contextual keyword can introduce complexity in the parser, and a different existing keyword like `in` or `do` could be used instead.
+
+```lua
+if local a, b = foo() in b > a then
+  print'Hello World, from Luau!'
+end
+```
+
+While Luau is a verbose language that uses keywords for the majority of its syntax, another approach is using semicolons as a separator. This can work well because statements can use semicolons as a separator, which will retain consistency with the language.
+
+```lua
+if local a, b = foo(); b > a then
+  print'Hello World, from Luau!'
+end
+```


docs/syntax-if-statements-initializers.md
```diff
@@ -0,0 +1,94 @@
...
+
+The reduced scope pollution improves register space in extreme cases (or auto-generated code) where developers have many variables defined and have to work around register limits by reducing the register size. In some scenarios, especially on Roblox, initializing variables within if statements can lead to improved performance and reduced complexity by avoiding unnecessary calls by developers. ⁤ A common paradigm used by Roblox developers is to use `Instance:FindFirstChild` in their condition and then access it afterwards instead of keeping an existing variable around in the new scope, polluting the existing scope.
+
+# Design
+
+If statements with initializers must match (following the Luau grammar) `'if' 'local' bindinglist ['=' explist] 'then'` and `'local' bindinglist ['=' explist] where exp 'then'` syntax. Parenthesis are also allowed around the initializer for consistency with other expressions. The variables declared by the initializer are only available to the block of that condition and will be undefined to the `elseif` conditions and blocks as well as the `else` block; any code after the if statement won't have the variables defined either.
```
vegorov-rbx, line 15: > Parenthesis are also allowed around the initializer for consistency with other expressions.

Can you give an example?

TheGreatSageEqualToHeaven: > Can you give an example? I'm not sure what you mean here and I hope it's not `if (local a = foo()) then`

That is exactly what it means, I felt it would be awkward to not allow `if (local a = foo()) where (a == 15) then`.

vegorov-rbx: That will be a 'no' from me.

TheGreatSageEqualToHeaven: The syntax would be inconsistent but it can be changed if it is somehow a problem, I feel it might be weird if you're writing out a condition and the initializer itself cant be parenthesized.

vegorov-rbx: It will actually be consistent, `()` is part of the expression grammar, local is a part of the if statement grammar.
The ability to write `if (a == 2)` like in other languages is just a coincidence of the ability to write group expressions.

TheGreatSageEqualToHeaven: > It will actually be consistent if it's removed; `()` is part of the expression grammar, local is a part of the if statement grammar. 

I guess that would also be fitting, I updated the RFC to remove that.

vegorov-rbx:
It seems that `elseif local` construct is not proposed?
Is that intentional?

TheGreatSageEqualToHeaven:
> It seems that `elseif local` construct is not proposed? Is that intentional?

Ah I forgot to mention it entirely, it was mentioned on discord and I forgot to mention it here lol. Let me fix that.

docs/syntax-if-statements-initializers.md
```diff
@@ -0,0 +1,114 @@
...
+
+Parser recovery may be more fragile due to the `local` keyword.
+
+Initializers increase the complexity of the language syntax and may obscure control flow in complex conditions.
+
+# Alternatives
+
+A different keyword or token can be used in place of `where`.
+
+Introducing a new contextual keyword can introduce complexity in the parser, and a different existing keyword like `in` or `do` could be used instead.
```
alexmccord, line 100: As a joke, I also thought of using `if` twice, e.g. `if local a, b = foo() if a > b then`. It's not a good idea though in practice.

docs/syntax-if-statements-initializers.md
```diff
@@ -0,0 +1,114 @@
...
+if local b = foo() where b == false then
+   print(b)
+else
+   print("true")
+end
+```
+
+`Output: true`
+
+When declaring multiple values inside of a condition, all of the variables will be evaluated.
```
alexmccord, line 53:
```suggestion
When declaring multiple values inside of a condition, all of the variables will be tested for truthiness.
```

docs/syntax-if-statements-initializers.md
```diff
@@ -0,0 +1,114 @@
...
+```lua
+local function foo()
+    return true
+end
+
+if local b = foo() where b == false then
+   print(b)
+else
+   print("true")
+end
```
alexmccord, line 48: This actually gave me a motivating reason why it might be useful for `b` to be visible in the `else` branch. Say you have some animal:

```lua
type Dog = { tag: "dog", pet: true, name: string }
type Cat = { tag: "cat", pet: true, name: string }
type Sheep = { tag: "sheep", pet: false }
type Animal = Dog | Cat | Sheep

if local animal = getAnimal() where animal.pet then
  print(`Hello {animal.name}!`)
else
  print(`{animal.tag} is not a pet`) -- generalizes to all current and future `animal.tag`s
end
```

docs/syntax-if-statements-initializers.md
goldenstein64, line 53: Just for clarification, how would the variables be combined when being tested? I misread the example at first, so I assumed all variables are combined with `or`, but it actually implies using `and`.

```lua
if local a, b = true, false then
  print("pass")
else
  print("fail")
end

-- same as this?
if local a, b = true, false where a and b then
  print("pass")
else
  print("fail")
end
```

I want to argue that this evaluation is ambiguous, since there is no explicit operation between the variables.

alexmccord:
I would be curious about the interaction with this RFC and key destructuring (https://github.com/luau-lang/rfcs/pull/24). Right now, I believe we should disallow key destructuring in `if local`s.

If we allowed it, then we'd have to decide on the right semantics, which might motivate us to disallow key destructuring in `if local` to begin with because it's too open to many different interpretations.

```lua
if local {.x, .y} = f() then
  print(x, y)
end
```

Which of the following best matches your intuition on what happens?

1. `where x and y`

This is kind of fine, but what happens if `f()` returns `nil` or any value that errors from any attempts to index on it, e.g. `boolean`, `number`? Then this condition blows up.

2. `where (type(t) == "table" or type(t) == "userdata" or type(t) == "vector" or type(t) == "string") and x and y`

This would fix the problem where the condition blows up, which I'd think is an improvement. But now, this means the VM would need to test the type of `t` before testing the rest of the `where` clause, which is possible but requires special handling in the compiler. The `x` and `y` becomes bound to `t.x` and `t.y` only if the type test succeeds.

If we ever allowed key destructuring, I think the second option makes the most sense because the first one just seems like a footgun waiting to happen if you attempt to stimulate the second:

```lua
local t_ok = false

do
  local t = f()
  -- the condition to type test whether `t` can be key destructured
  if type(t) == "table" or type(t) == "userdata" or type(t) == "vector" or type(t) == "string" then
    local {.x, .y} = t
    -- the condition of the `where` clause
    if x and y then
      t_ok = true -- if both type test and where clause succeeds, then `t_ok` is true
      -- ...
    end
  end
end

if not t_ok then -- if either type test or `where` clause failed, this is the `else` branch
  -- ...
end
```

So yeah, it's very easy to start by just not allowing key destructuring if that RFC is ratified.

vegorov-rbx:
Hot take: make `where` part required if more than one local is defined.
This includes `local {.a, .b}` since that RFC says that this defines 'a' and 'b' - two locals.

docs/syntax-if-statements-initializers.md
vegorov-rbx: Another case:
```
if local success, msg = pcall(f) where success then
    print(`all went well`)
else
    print(`f failed with: {msg}`)
end
```

alexmccord:
I also wanted to note that this doesn't mention whether we should extend `if local` to `if` expressions as well, not just `if` statements.

TheGreatSageEqualToHeaven:
> I also wanted to note that this doesn't mention whether we should extend `if local` to `if` expressions as well, not just `if` statements.

Honestly just forgot (like elseif cases), I'm honestly not sure if if expressions should have initializers. The RFC is called "If statement initializers" so it did only define that, what's your opinion on it?

TheGreatSageEqualToHeaven:
> I would be curious about the interaction with this RFC and key destructuring (#24). Right now, I believe we should disallow key destructuring in `if local`s.

> If we allowed it, then we'd have to decide on the right semantics, which might motivate us to disallow key destructuring in `if local` to begin with because it's too open to many different interpretations.

> Hot take: make `where` part required if more than one local is defined. This includes `local {.a, .b}` since that RFC says that this defines 'a' and 'b' - two locals.

This could work, I didn't consider key destructuring (nor am I particularly a fan of it in the first place). Making `where` required for initializers with more than one local does solve the ambiguity of the compiler adding a truthy test.

docs/syntax-if-statements-initializers.md
alexmccord: I've come to agree with @vegorov-rbx on making `where` clause required if more than two locals are involved in the initializer. Good example is found here: https://github.com/luau-lang/rfcs/pull/23/files#r1492770178.

docs/syntax-if-statements-initializers.md
TheGreatSageEqualToHeaven: > This actually gave me a motivating reason why it might be useful for `b` to be visible in the `else` branch. Say you have some animal:
> 
> ```lua
> type Dog = { tag: "dog", pet: true, name: string }
> type Cat = { tag: "cat", pet: true, name: string }
> type Sheep = { tag: "sheep", pet: false }
> type Animal = Dog | Cat | Sheep
> 
> if local animal = getAnimal() where animal.pet then
>   print(`Hello {animal.name}!`)
> else
>   print(`{animal.tag} is not a pet`) -- generalizes to all current and future `animal.tag`s
> end
> ```

> Another case:
> 
> ```
> if local success, msg = pcall(f) where success then
>     print(`all went well`)
> else
>     print(`f failed with: {msg}`)
> end
> ```

Seeing it now allowing `b` to be visible in the `else` branch (and most likely `elseif` branches) would be useful.

docs/syntax-if-statements-initializers.md
```diff
@@ -6,13 +6,13 @@ Introduce an initializer expression that declares and initializes variables in i
...
 
 # Design
 
-If statements with initializers must match (following the Luau grammar) `'if' 'local' bindinglist ['=' explist] 'then'` and `'local' bindinglist ['=' explist] where exp 'then'` syntax. The variables declared by the initializer are only available to the block of that condition and will be undefined to the `elseif` conditions and blocks as well as the `else` block; any code after the if statement won't have the variables defined either.
+If statements with initializers must match (following the Luau grammar) `'if' 'local' bindinglist ['=' explist] 'then'` and `'local' bindinglist ['=' explist] where exp 'then'` syntax. The variables declared by an initializer are only available to the if statement's blocks; any code after the if statement won't have the variables defined.
```
alexmccord, line 15: This grammar has two issues: it lets you omit the expression, and it still allows for multiple bindings without a `where` clause. You probably want (taken from https://luau-lang.org/grammar):

```diff
  stat = varlist '=' explist |
      ...
      'repeat' block 'until' exp |
-     'if' exp 'then' block {'elseif' exp 'then' block} ['else' block] 'end' |
+     'if' cond 'then' block {'elseif' cond 'then' block} ['else' block] 'end' |
      'for' binding '=' exp ',' exp [',' exp] 'do' block 'end' |

- ifelseexp = 'if' exp 'then' exp {'elseif' exp 'then' exp} 'else' exp
+ ifelseexp = 'if' cond 'then' exp {'elseif' cond 'then' exp} 'else' exp

+ cond = 'local' binding '=' exp ['where' exp] |
+     'local' bindinglist '=' explist 'where' exp |
+     exp
```

I also noticed that my ask for adding initializers in `if` expressions earlier (reflected above) also permits this:

```lua
if local x = if local ok, res = pcall(f) where ok then res else nil where x == nil then
  -- ...
end
```

and I think we'd all want to live without seeing that.

alexmccord:
Just ran into something relatively verbose where this RFC would've shined:

```lua
function PlayerControl:GetEntities()
    if self.RemovedSelf then
        return self.Instances
    end

    local entity = self:TryGetEntity()
    if entity == nil then
        return self.Instances
    end

    local index = table.find(self.Instances, entity.Root)
    if index == nil then
        return self.Instances
    end

    table.remove(self.Instances, index)
    self.RemovedSelf = true

    return self.Instances
end
```

Could be compacted into

```lua
function PlayerControl:GetEntities()
    if self.RemovedSelf then
    elseif local entity = self:TryGetEntity() where entity ~= nil then
    elseif local index = table.find(self.Instances, entity.Root) where index ~= nil then
        table.remove(self.Instances, index)
        self.RemovedSelf = true
    end

    return self.Instances
end
```

docs/syntax-if-statements-initializers.md
TheGreatSageEqualToHeaven: I'll update my grammar examples with your change, are you asking for if expressions with initializers too or are you against them?

TheGreatSageEqualToHeaven:
> This would make a great addition to this RFC in the motivation section.

Updated the RFC with this as a motivation.

docs/syntax-if-statements-initializers.md
alexmccord: I think allowing initializers in `if` expressions is ok.

vegorov-rbx: I know that implementation complexity is sometimes completely ignored during design of a language, but I would suggest excluding extension of `if` expressions with this.
Reason being that current compiler model doesn't support local allocations inside expressions (all expression registers are temporary and can be clobbered).
I would understand if there was a great motivation for `if` expressions to rewrite large parts of the compiler, but seeing that this was more of an after-thought and nice-to-have, I suggest excluding them to a separate proposal (and I would probably ask for a compiler prototype in that).

Note that this will also be a reason why `do ... end` cannot become an expression so easily.

Nolan-O:
Could an alternative solution be `andif`? 

```lua
function PlayerControl:GetEntities()
	if not self.RemovedSelf then
		local entity = self:TryGetEntity()
	andif entity ~= nil then
		-- still in the same scope, or has access to previous scope
		local index = table.find(self.Instances, entity.Root)
	andif index then
		table.remove(self.Instances, index)
		self.RemovedSelf = true
	else
		-- false for any andif will go here
		-- convenient warning/error place, could return here too
	end

	return self.Instances
end```

**Comment by TheGreatSageEqualToHeaven** [comment]
Updated the RFC with @alexmccord's grammar example excluding if expressions.

**Comment by TheGreatSageEqualToHeaven** [comment]
> Could an alternative solution be `andif`?

The example you provided isn't an initializer but a completely separate alternative from the proposed syntax.

**Comment by Nolan-O** [comment]
Yes I was commenting an alternative to the proposed syntax, that was not a mistake.

File: docs/syntax-if-statements-initializers.md [hunk 6]
```diff
@@ -52,7 +52,24 @@ end
...
+      ...
+      'repeat' block 'until' exp |
+-     'if' exp 'then' block {'elseif' exp 'then' block} ['else' block] 'end' |
++     'if' cond 'then' block {'elseif' cond 'then' block} ['else' block] 'end' |
+      'for' binding '=' exp ',' exp [',' exp] 'do' block 'end' |
+      ...
+
+- ifelseexp = 'if' exp 'then' exp {'elseif' exp 'then' exp} 'else' exp
++ ifelseexp = 'if' cond 'then' exp {'elseif' cond 'then' exp} 'else' exp
+
```
alexmccord, line 68: This still proposes the `if` expression changes.
```suggestion

```

AsynchronousAI:
Why not use python-like walrus syntax:
```lua
if (randnum := math.random(1,100)) == 25 then
 print("25")
else
 print(`Not 5 ({randnum})`):
end
```

alexmccord:
> Why not use python-like walrus syntax:
> 
> ```lua
> if (randnum := math.random(1,100)) == 25 then
>  print("25")
> else
>  print(`Not 5 ({randnum})`):
> end
> ```

Ignoring that I think `if local` syntax is better and more consistent with the language philosophy, there are technical reasons: https://github.com/luau-lang/rfcs/pull/23#discussion_r1495739113. This would be introducing a local in the midst of an expression, which can't work for the same reasons.

AsynchronousAI:
> > Why not use python-like walrus syntax:
> > ```lua
> > if (randnum := math.random(1,100)) == 25 then
> >  print("25")
> > else
> >  print(`Not 5 ({randnum})`):
> > end
> > ```
> 
> Ignoring that I think `if local` syntax is better and more consistent with the language philosophy, there are technical reasons: [#23 (comment)](https://github.com/luau-lang/rfcs/pull/23#discussion_r1495739113). This would be introducing a local in the midst of an expression, which can't work for the same reasons.

Probably, I also saw your other example. Maybe this syntax:
```lua
elseif entity = self:TryGetEntity() and entity == nil then
-- instead of:
-- elseif local entity = self:TryGetEntity() where entity == nil then
```
I feel like adding a `local` is a bit too extra. But the RFC's idea would be **extremely** helpful.

deviaze:
> Probably, I also saw your other example. Maybe this syntax:
> ```lua
> elseif entity = self:TryGetEntity() and entity == nil then
> -- instead of:
> -- elseif local entity = self:TryGetEntity() where entity == nil then
> ```
> 
> I feel like adding a `local` is a bit too extra. But the RFC's idea would be **extremely** helpful.

The new syntax must be currently invalid (for backwards compatibility) and not be a common error; that example looks like someone forgot a second `=` in `==`, for example. IMHO, it's a lot harder to parse as well (for both humans & parsers), vs. `if local`. 

Since we've been debating syntax (on the Discord at least) for a while, especially about replacing `where` with something shorter like `in`: 

After much testing and deliberation, I find that I much prefer the original `if local identifier = value where expr` syntax.

Why? 
Looking at real usecase examples, `if local ... where` ends up reading a lot cleaner and making sense: 

```lua
-- assuming if not local is in/will get added to the RFC;
-- valid targets: either players not on your team or npcs with "EnemyBot" tag
local function get_valid_target(hit: Instance): (Humanoid?, Player?)
	if not local HitModel = hit.Parent where HitModel:IsA("Model") then
		-- guard clause
		-- should we add special syntax for guard clauses? could be helpful 
	elseif local humanoid = HitModel:FindFirstChildOfClass("Humanoid") 
	where humanoid and humanoid.Health > 0 then
		-- atm if a "where" clause is included then humanoid ~= falsey has to be explicitly checked
		if HitModel:HasTag("EnemyBot") then
			return humanoid, nil
		elseif local HitPlayer = Players:GetPlayerFromCharacter(HitModel)
		where HitPlayer and HitPlayer.Team ~= LocalPlayer.Team then
			return humanoid, HitPlayer
		end
	end
	return nil, nil
end
```
I'm not sure how we'd like to format `where then`s that span into the second line (indent-wise at least), but I often like putting them on the next line when they get kinda long.

AsynchronousAI:
> 
> The new syntax must be currently invalid (for backwards compatibility) and not be a common error; that example looks like someone forgot a second `=` in `==`, for example.
> 
> IMHO, it's a lot harder to glance at and parse as well (for both humans & parsers), vs. `if local`.
> 
> Since we've been debating syntax (on the Discord at least) for a while, especially about replacing `where` with something shorter like `in`:
> 
> After much testing and deliberation, I find that I much prefer the original `if local identifier = value where expr` syntax.
> 
> Why? Looking at real usecase examples, `if local ... where` ends up reading a lot cleaner and making sense:
> 
> ```lua
> -- assuming if not local is in/will get added to the RFC;
> -- valid targets: either players not on your team or npcs with "EnemyBot" tag
> 
> local function get_valid_target(hit: Instance): (Humanoid?, Player?)
> 	if not local HitModel = hit.Parent where HitModel:IsA("Model") then
> 		-- guard clause
> 		-- should we add special syntax for guard clauses? could be helpful 
> 	elseif local humanoid = HitModel:FindFirstChildOfClass("Humanoid") 
> 	where humanoid and humanoid.Health > 0 then
> 		-- atm if a "where" clause is included then humanoid ~= falsey has to be explicitly checked
> 		if HitModel:HasTag("EnemyBot") then
> 			return humanoid, nil
> 		elseif local HitPlayer = Players:GetPlayerFromCharacter(HitModel)
> 		where HitPlayer and HitPlayer.Team ~= LocalPlayer.Team then
> 			return humanoid, HitPlayer
> 		end
> 	end
> 	return nil, nil
> end
> ```
> 
> I'm not sure how we'd like to format `where then`s that span into the second line (indent-wise at least), but I often like putting them on the next line when they get kinda long.

I still believe the syntax should be shorter maybe:
```lua
elseif humanoid as HitModel:FindFirstChildOfClass("Humanoid") and humanoid and humanoid.Health > 0 then
```

deviaze:
> > 
> > The new syntax must be currently invalid (for backwards compatibility) and not be a common error; that example looks like someone forgot a second `=` in `==`, for example.
> > 
> > IMHO, it's a lot harder to glance at and parse as well (for both humans & parsers), vs. `if local`.
> > 
> > Since we've been debating syntax (on the Discord at least) for a while, especially about replacing `where` with something shorter like `in`:
> > 
> > After much testing and deliberation, I find that I much prefer the original `if local identifier = value where expr` syntax.
> > 
> > Why? Looking at real usecase examples, `if local ... where` ends up reading a lot cleaner and making sense:
> > 
> > ```lua
> > -- assuming if not local is in/will get added to the RFC;
> > -- valid targets: either players not on your team or npcs with "EnemyBot" tag
> > 
> > local function get_valid_target(hit: Instance): (Humanoid?, Player?)
> > 	if not local HitModel = hit.Parent where HitModel:IsA("Model") then
> > 		-- guard clause
> > 		-- should we add special syntax for guard clauses? could be helpful 
> > 	elseif local humanoid = HitModel:FindFirstChildOfClass("Humanoid") 
> > 	where humanoid and humanoid.Health > 0 then
> > 		-- atm if a "where" clause is included then humanoid ~= falsey has to be explicitly checked
> > 		if HitModel:HasTag("EnemyBot") then
> > 			return humanoid, nil
> > 		elseif local HitPlayer = Players:GetPlayerFromCharacter(HitModel)
> > 		where HitPlayer and HitPlayer.Team ~= LocalPlayer.Team then
> > 			return humanoid, HitPlayer
> > 		end
> > 	end
> > 	return nil, nil
> > end
> > ```
> > 
> > I'm not sure how we'd like to format `where then`s that span into the second line (indent-wise at least), but I often like putting them on the next line when they get kinda long.
> 
> I still believe the syntax should be shorter maybe:
> ```lua
> elseif humanoid as HitModel:FindFirstChildOfClass("Humanoid") and humanoid and humanoid.Health > 0 then
> ```

This won't work because the parser reads left-to-right. It needs to look for something specific that tells it to parse for an if-statement-initializer, and that's the job of the `local` in `if local`. `humanoid` isn't a restricted keyword or any currently-illegal token so it doesn't do the job. 

This looks like another Pythonism; Luau and Python have very different language philosophies, and we typically don't want to adopt python-like syntax into Luau because it tends to be at odds with ours.

I don't think "as" nor the walrus operator are explicit enough to be used in this case; they're too small (easy to get skipped over in longer Luau conditionals) and don't make the whole "let's initialize a new variable here" part of the if statement initializer very obvious.

AsynchronousAI:
> This won't work because the parser reads left-to-right. It needs to look for something specific that tells it to parse for an if-statement-initializer, and that's the job of the `local` in `if local`. `humanoid` isn't a restricted keyword or any currently-illegal token so it doesn't do the job.
> 
> This looks like another Pythonism; Luau and Python have very different language philosophies, and we typically don't want to adopt python-like syntax into Luau because it tends to be at odds with ours.
> 
> I don't think "as" nor the walrus operator are explicit enough to be used in this case; they're too small (easy to get skipped over in longer Luau conditionals) and don't make the whole "let's initialize a new variable here" part of the if statement initializer very obvious.

Makes sense, in that case I think an `if local` syntax would be fine.

docs/syntax-if-statements-initializers.md
potatoprank: I think where clauses should just always be a requirement 😰

deviaze:
We've extensively discussed this RFC on the OSS Discord lately, and I want to update this page on some major points from those discussions.

To summarize the main changes to the RFC proposal: 

1. Restrict and clarify the primary motivations of the RFC.
2. Restrict the scope of `if/elseif local` bindings to a single branch (no fallthrough).
3. Finalize `in` as the syntactic replacement for the `where` keyword.

## motivations

According to aaron, multiple other luau language engineers, and the majority of active chat participants I see discussing this over there: 
The primary motivations of this RFC should be to
- a) introduce currently impossible-to-write logic to the language,
- b) provide a base for further conditional destructuring syntax,
- c) improve the ergonomics of one of the most common Luau idioms.

At the moment, this RFC can be described as "an ergonomic wrapper around `do end` blocks and nested conditionals".

This is not what the Luau engineers seem to want to implement.

To quote aaron: 
> What I hand *[sic]* in mind for the reasoning is: there is a particular reasonable programming pattern that you cannot write currently without having an unnecessarily wide scope for the binding, and that that's worth addressing because it's generally good programming discipline to limit the scope of bindings. likewise, adding a programming pattern for this gives a direct path to supporting more expressive versions of it that can do destructuring as well. 

In general, the engineers seem extremely interested in implementing `if local` specifically for future conditional destructuring and pattern matching expansions. I'd prefer to let them comment on those features because I'm not very sure what those plans look like atm. 

## fallthrough
The general consensus seems to be "you can already do this with a `do end` block" and "there's no point";
To quote aaron again:
> yeah, no way, the scoping rule should be that the local is bound in the block of that branch, not in the whole if statement 
> it doesn't make any sense, i explained it here before (and should really go write it on the RFC too), but if you want it bound in the other branches... that's already literally how the scoping rules of luau work, so i don't get the point at all 
> ```lua
> do
>     local success, result = pcall(...)
>     if success then
>         ...
>     else
>         ...
>     end
> end
> ```
> if it's just to make the above code more succinct, i don't think it's worth it at all

To quote WheretIB:
> "Wouldn't mind to see [branch fallthrough] go. In C++ that scoping rule is more surprising than useful.
> People are like "what? it works that way?" 
> Later they would say "oh I see, it makes sense for error handling" but I'm yet to see it used in else. It's like a 0.01% of use cases and C++ of course had to accommodate that.

> (tbh, wouldn't mind the whole rfc to not go in either) 

If the language engineers themselves would rather just *not have the feature* than implement it with fallthrough, then it makes the most sense to me to listen to them and figure out what they'd intend it for and why.

### footgun

Additionally, I argue that the currently-proposed fallthrough behavior is actually a footgun: it encourages minimized LOC, never-nesting, and excessive DRY at the expense of code readability. 

I was a pretty vocal proponent for fallthrough until around 2 weeks ago. This is why I've changed my mind. I've spent too many hours these last few months writing and grok-testing real life usecases for `if local` fallthrough. What I've found is more than anything, fallthrough allowed me to eliminate most cases of conditional nesting by "chaining" `if/elseif locals`. The vast majority with hacky guard clauses that didn't seem like intended behavior--it obscured control flow in a way unnatural to Luau's semantic simplicity and scoping conventions. 

Nesting conditionals is not necessarily a bad thing. It allows programmers to easily parse what the control flow is doing at any given point, what variables are in scope and where they're defined and utilized, just by lining up indentation. `if/elseif local` fallthrough chaining destroys this.

On the other hand, allowing `if local`s without fallthrough seemed to provide a pretty significant improvement to readability, general grok-iness, and ergonomics without that footgun potential. In my opinion, this actually leads to *better* and *cleaner* nesting behavior than both the current idiomatic situation & the proposed fallthrough situation. 

Here's a rewrite of my hit validation example which shows how `if local` remains an extremely helpful ergonomic and scoping addition without fallthrough behavior:
(note, this could be written in fewer LOC but it wouldn't be as explicit)
```lua
function Caster:VerifyTargetHit(hit: Instance): (Humanoid?, Player?)
    if local hitModel = hit.Parent in hitModel:IsA("Model") then
        if local humanoid = hitModel:FindFirstChildOfClass("Humanoid") in
            humanoid and humanoid.Health > 0 
        then
            if hitModel:HasTag("EnemyBot") then
                return humanoid, nil
            elseif local player = Players:GetPlayerFromCharacter(hitModel) in
                player and player.Team ~= LocalPlayer.Team
            then
                return humanoid, player
            end
        end
    end
end
```
## new functionality

The RFC, as currently proposed, doesn't properly introduce this new functionality to the language:
```lua
if local result = someExpensiveCall() in result and result.ok then
    -- bind result *only* within this branch
end
```
There is currently no way to express this in Luau:
- create a local binding specific to a branch
- and test it against a condition
- and only call the RHS once.

By limiting the scope of the bound identifier to a single branch, we can introduce this brand new behavior **and** simultaneously improve the ergonomics, readability, and scopeability (is that even a word) of one of the most common Lua/u idioms ever:
```lua
-- Assign a result to an identifier. do something if it's truthy. 
local container = SomeClass:AnnoyinglyLongMethod("meow")
if container then
    -- do things with container
end
-- ignore the now bound container forever
-- or
if SomeClass:AnnoyinglyLongMethod("meow") then
    local container = SomeClass:AnnoyinglyLongMethod("meow")
    -- yay we're now bound only in here! 
    -- hope that wasn't an expensive method call (or caused side effects) cause we just called it twice
    -- do things with container 
end
```
Which can now be written:
```lua
if local container = SomeClass:AnnoyinglyLongMethod("meow") then
  -- do things with container knowing it happily exists
end
-- what container?? never heard of her
```
I might be extremely biased here, but just this basic `if local identifier = value then branch end` syntax right here--an immediate and obvious QoL win for every Luau developer across the whole platform regardless of skill level, and can have immediate benefits on any codebase written in the language¹. But we still want conditions.

## ~~where~~in local bikeshedding
This has been the primary and most vocal point of contention on the server wrt. to this entire RFC. 

There have been many many many discussion points and alternate syntaxes raised for the "where" clause over the months, but the only ones that have survived are: the original `where` (because it's part of this RFC), `in`, and `;`. Of those 3, the only one that's garnered major public support is `in`. 

- `in` is already a keyword.
- nobody wants to add a new keyword to the language
- it makes a lot more sense with future pattern matching & destructuring syntactic expansions.
- it's similar to other languages with a similar feature (Haskell was cited as one)
- to paraphrase aaron et al., "you can't put local a = foo() *where* the expression a > 2 ..? That makes no sense. You put the local a **into** the expression a > 2."
- We want to keep the `local` part visually to the left of/above the conditional part for visual clarity: new identifiers shouldn't appear out of nowhere until you see a `where` clause. (a proposed idea was inverting the order: `if a > 4 where local a = f() then end` which people preferred with the `where` keyword). 
- Luau is not a declarative query language like SQL (wrt. SQL's where clause) therefore `where` is out of place in Luau.
- "`;`" is out of place in a language that uses keywords not keysymbols for control flow.

Basically, it seems we've argued a lot about this and restated the same things multiple times. We all want to keep the conditional clause as a feature; `in` is probably the best option we're going to get here and it seems like the language engineers are all in (pun *in*tended) agreement here.

## conclusion and final thoughts
`if local` is an amazing idea. To me, it's a no-brainer awesome improvement to the most common conditional checking idiom we have in Luau. Even in its first most basic form it'll be useful. It adds behavior we currently can't express in the language while simultaneously improving code organization and developer happiness. We can only imagine the power we may be afforded with this when combined with table destructuring syntax, pattern matching, destructuring in loops, etc. in future expansions to/related to this feature. 

I hope we can finalize this rfc discussion/debate pretty soon and get this LGTMed for implementation ~~before the type solver release?!~~. 

---
This is the type of feature you'll end up typing out subconsciously a month later, then smiling because it exists. Seemingly small but with big future {destructuring, pattern_matching} = potential.

(¹as soon as it gets implemented of course)

docs/syntax-if-statements-initializers.md
```diff
@@ -0,0 +1,167 @@
...
+    return true, false
+end
+
+if local a, b = foo() in a and b then
+else
+    print'Hello World, from Luau!'
+end
+```
+
+`Output: Hello World, from Luau!`
```
aatxe, line 121: With an eye towards consistent behavior with destructuring in the future, I would prefer to see this changed to requiring `a` and `b` be non-nil implicitly, but it's ultimately okay if we air on the conservative side here since this requirement is forwards-compatible with changing it in the future to mean that. In general, I think the behavior we want though is that using `local a = foo()` and `local a, b = foo()` in a condition will only ever take that branch when the results are non-nil.

docs/syntax-if-statements-initializers.md
aatxe: I commented this elsewhere, but I realized this is pretty much a good point to say it in. The desirable behavior IMO is that any variable initialized in an `if local` statement will be tested to be non-`nil`. The direct argument would be that putting the initialization in a _condition_ is an indication that you want it to be bound to some sort of non-`nil` value in the block following it. The forward-looking advantage of this is that it's consistent with how destructuring should behave, e.g. if we allow destructuring tables in local definitions (#24), you could write something like:

```lua
if local {.x, .y, .z} = point then
    -- x, y, z are defined and non-`nil`
end
```

The main downside is that it means the case where you specifically want to test a result for truthiness, rather than non-`nil`ness will require you to write the condition clause, i.e.

```lua
if local cond = bool() then
    -- cond is not `nil`, but could be `true` or `false`
end

if local cond = bool() in cond then
    -- cond is not `nil`, and has been tested `true` by the condition clause
end

if local cond = bool() in not cond then
    -- cond is not `nil`, and has been tested `false` by the condition clause
end
```

I think this is a reasonable decision to make. Requiring `in` is a more conservative choice that is forwards-compatible with this evolution though.

docs/syntax-if-statements-initializers.md
aatxe: IMO `pcall` is the most compelling reason to scope the locals this way, but this scoping behavior is a consistent footgun in C++ where bindings in conditions are scoped to the whole if statement and you can trivially end up dereferencing null pointers accidentally using them in the wrong branch. I think, outside of `pcall`, scoping the bindings to the whole block means making more bugs with accidentally using `nil` possible, and I see little benefit since you can already write the following to have this scoping behavior.
```lua
do local success, msg = pcall(f)
    if success then
        print(`all went well`)
    else
        print(`f failed with: {msg}`)
    end
end
```

docs/syntax-if-statements-initializers.md
deviaze: That makes sense. If you want to allow `b` to either exist or not, you should probably have to say that explicitly. Otherwise the fact that you asked for `local a, b = foo()` probably means that you wanted both `a` and `b` to both exist.

docs/syntax-if-statements-initializers.md
```diff
@@ -0,0 +1,167 @@
...
+
+```lua
+local function foo()
+    return true
+end
+
+if local b = foo() then
+   print(b, "truthy block")
+else
+   print(b, "falsy block")
```
alexmccord, line 82: Wanted to add that this design point is one we have to live with in perpetuity. If we decided to take this fallthrough semantics out of this RFC, then we're stuck with that behavior.

Using `boolean` is an ok example, but it doesn't strongly argue for it as opposed to some other example I posted earlier wrt `animal` and one that @vegorov-rbx posted with `pcall` example. If a design point is a "choose now or forever hold your peace," it especially needs an equally as compelling argument to support it.

So I'd rather block the RFC until we are able to agree on this design point, whether locals should fallthrough or not.

An example where the behavior is breaking change would also be good to include in the alternatives section.

```lua
if local x = f() in x % 2 == 0 then
  print(x) --> any even numbers
else
  print(x) --> nil
end
```

As is, this RFC proposes the `print(x)` in the `else` branch to print any odd numbers.

bradsharp:
I've very quickly skimmed the comments but could we drop the need for `local` inside of the assignment in favor of it being implicit to the outermost scope? In other words:
```lua
if x = f() and x == 5 then

end
```
Is just sugar for:
```lua
local x = f()
if x == 5 then

end
```

This works nicely with the fall-through `pcall` case too as:
```lua
if success, result = pcall(...) then
  -- success here
else
  -- failure here
  warn(result)
end
```
Is sugar for:
```lua
local success, result = pcall(...)
if success then
  -- success here
else
  -- failure here
  warn(result)
end
```

deviaze:
> I've very quickly skimmed the comments but could we drop the need for `local` inside of the assignment in favor of it being implicit to the outer scope? In other words:
> 
> ```lua
> if x = f() and x == 5 then
> 
> end
> ```
> 
> Is just sugar for:
> 
> ```lua
> local x = f()
> if x == 5 then
> 
> end
> ```

This defeats the point of the RFC, and, as we've stated earlier, causes confusion between assignments (=) and equality checks (==).  The new syntax needs to be obvious and explicit; `if a = expr()` needs to remain a syntax error since it's almost always a user error.

The purpose of this RFC is to provide syntax to 
- a) assign a local binding to an expression within a conditional scope, and
- b) test the value of that binding without evaluating the bound expression multiple times. 

The main usecase is any extremely common idiom like the following: 
```lua
-- with the implicit "in binding ~= nil" clause:
if local humanoid = hit:FindFirstChildOfClass("Humanoid") then
    -- binds local humanoid and executes branch if humanoid ~= nil
end
-- local humanoid is unbound here

-- or with an explicit "in" clause:
if local result = getSomeResult() in table.find(validOptions, result) then

end
```
At the moment, we kinda just need to agree upon fallthrough semantics to get this RFC passed.

bradsharp:
I am not convinced that requiring `local` is the right workaround to avoid the `if x = 5 then` footgun.

- If I write a for loop I don't include `local` as part of my initializer, and doing so is actually invalid syntax. This would introduce inconsistent syntax that our users will need to learn.
- Adding fallthrough behavior to a variable declared as `local` seems like something that may also lead to confusion for some users.

I understand the benefits this feature would bring but I think we need to be careful about the other "gotchas" we might introduce as part of it. I would like to see some more alternatives before we move forward with this.

deviaze:
> If I write a for loop I don't include local as part of my initializer, and doing so is actually invalid syntax. This would introduce inconsistent syntax that our users will need to learn.

This feature combines the existing and well-known `local variable = expr()` and `if condition() ...` and logically uses syntax from both features. Although I can see the similarities (injecting bindings into a block), `if locals` are sufficiently distinct from `for` loops for users to easily confuse the two. 

Explicit variable scoping is a primary motivator for this feature. The proposed "if local" syntax expresses this purpose and provides necessary visual distinction between standard if statements and if statement initializers. Use of the `in` keyword for the conditional clause might seem more confusing (imho) at first, but it's relatively easy to learn as new syntax.

Future expansions may introduce this syntax elsewhere in the language; we've discussed (on the Discord at least) expressive `in` matching, `while locals` and general bindings-in-expression syntax as followups to if statement initializers.

> Adding fallthrough behavior to a variable declared as local seems like something that may also lead to confusion for some users.

Counterargument: Due to Luau's informal left/right scope <-> indentation flow, variables bound to the top level `if` appear indented left relative to the branches and therefore scoped to entire control structure (not just its first conditional branch). As discussed earlier in this thread, fallthrough behavior allows for more expressive control flow (ergonomic switch/case-like semantics for example), and reduced verbosity for common control flow paradigms.

bradsharp:
If we were only to ever do this for `if` statements then I could be convinced but I think it is inevitable that we will also want this for `while` loops to. At that point I don't think there is such a strong distinction and it would be a shame to limit support for `while` loops due to the syntax we chose earlier on.

Unfortunately, I also don't have a great solution to this. If we had a `let` or `var` keyword then I can imagine us doing:
```
if var x = foo() and x == y then
while var x = foo() and x do
for var x = 1, 5 do
```
At least in this case the semantics seem a little more reasonable to me.

aatxe:
Agreed, I do not think that the fact that `local` is two characters longer than `let` somehow means that constructs like `if local` and `while local` shouldn't be done or are at all worse than `if let` and `while let`. Even though there are forms like `for` loops that introduce new local bindings as well, I don't think there's any particularly harsh burden associated with learning that `local` in conditional contexts (`if` and `while`) works the same as `local` outside of them combined with a test.

deviaze:
Definitely agreed; imho `if/while local` makes perfect sense in Luau anyway. It's `local` because it scopes a local variable into its associated control structure.

TheGreatSageEqualToHeaven:
@bradsharp 

Sorry I've been busy.

I'm fine with `if local` and `while local`, and additionally the RFC can be updated to include both `if` & `while` statement initializers. The current issue is how the initializers should be scoped.

```luau
if local a,b = foo() then
    print(a,b) --> "hello", "world"
else 
    print(a,b) --> nil, nil
end
print(a,b) --> nil, nil
```

vs

```luau
if local a,b = foo() then
    print(a,b) --> "hello", "world"
else 
    print(a,b) --> "hello", "world"
end
print(a,b) --> nil, nil
```

Since the RFC hasn't moved at all I am fine with the former syntax being the case rather than the latter, which is the current way.

Wunder-Wulfe:
I believe another format could also be considered in the form of a new assignment operator to function as `x = y; return x`

```luau
local temp
local root: = if (temp->player.Character)~=nil then temp:FindFirstChild("HumanoidRootPart") else nil

local health = if (temp->player.Character)~=nil and (temp->temp:FindFirstChildOfClass("Humanoid"))~=nil then temp.Health else 0
```

deviaze:
> `local health: number = if local temp=player.Character in temp~=nil then (if local temp = temp:FindFirstChildOfClass("Humanoid") in temp~=nil then temp.Health else 0) else 0;`

Note that `if local` *expressions* were removed from this RFC because they'd require major changes to the compiler. 

> `if temp->player.Character and temp->temp:FindFirstChildOfClass("Humanoid")`

This is similar to the `x = f() and x == c`-style syntax we agreed against due to operator precedence ambiguity. The new `->` operator would have to bind very tightly to its RHS and I don't really see the point.

PR #24, opened 2024-02-08
Key destructuring (`local {.a, .b} = t`)

Kampfkarren:
[Rendered](https://github.com/Kampfkarren/rfcs-1/blob/dict-destructuring/docs/syntax-key-destructuring.md)

### DIFF
--- a/docs/syntax-key-destructuring.md
+++ b/docs/syntax-key-destructuring.md
@@ -141,7 +141,7 @@ local { .real as aliased } = t
 local aliased = t.real
 ```
 
-This helps support mutliple assignments on the same name:
+This helps support multiple assignments on the same name:
 ```lua
 local { .name as nameA } = getObject(a)
 local { .name as nameB } = getObject(b)
@@ -176,7 +176,7 @@ The syntax could support `.["key with spaces"]`, but this may open up a can of w
 This also blocks nested destructuring, such as JavaScript's `const { a: { b } } = t` to mean `b = t.a.b`.
 
 ### Roblox - Property casing
-Today in Roblox, every index doubly works with camel case, such as `part.position` being equivalent to `part.Position`. This use is considered deprecated and frowned upon. However, even with variable renaming, this becomes significantly more appealing. For example, it is common you will only want a few pieces of informaiton from a `RaycastResult`, so you might be tempted to write:
+Today in Roblox, every index doubly works with camel case, such as `part.position` being equivalent to `part.Position`. This use is considered deprecated and frowned upon. However, even with variable renaming, this becomes significantly more appealing. For example, it is common you will only want a few pieces of information from a `RaycastResult`, so you might be tempted to write:
 
 ```lua
 local { .position } = Workspace:Raycast(etc)
@@ -188,7 +188,7 @@ local { .position } = Workspace:Raycast(etc)
 
 ### Syntax for destructuring
 
-Many syntaxes have been proposed for destructuring. The most significant problem with any proposal is that is must be unambiguous to the reader whether or not the destructor is for **dictionaries** or for **arrays**.
+Many syntaxes have been proposed for destructuring. The most significant problem with any proposal is that it must be unambiguous to the reader whether or not the destructor is for **dictionaries** or for **arrays**.
 
 An intuitive suggestion is `local { a, b } = t`, but this syntax fails this test--it is not obvious if this is `local a = t.a` or `local a = t[1]`, regardless of whatever syntax is chosen for array destructuring (should it exist).
 
--- a/docs/syntax-key-destructuring.md
+++ b/docs/syntax-key-destructuring.md
@@ -41,7 +41,7 @@ local style = props.style
 ```js
 const { position, style } = props
 
-// Or even...
+// Supported in JavaScript, but not this proposal
 function MyComponent({
 	position,
 	style,
@@ -63,9 +63,9 @@ get("/users", ({
 This RFC proposes expanding the grammar of `binding`:
 ```patch
 binding = NAME [':' Type]
-+         | `{` keydestructor [`,` keydestructor] [`,`] `}` [':' Type]
++         | `{` keydestructor { [`,` keydestructor] } [`,`] `}` [':' Type]
 
-+keydestructor = `.` NAME [':' Type]
++keydestructor = `.` NAME [`as` NAME] [':' Type]
 ```
 
 This would allow for the following:
@@ -74,9 +74,6 @@ This would allow for the following:
 local { .a, .b }, c = t
 
 for _, { .a, .b } in ts do
-end
-
-local function f({ .a, .b }, c)
 end
 ```
 
@@ -96,6 +93,14 @@ local { .x, .y }
 
 ...is valid to parse and will execute, but can be linted against. This might even be emergent, as this will be statically equivalent to `nil.x`, which Luau's type system can catch.
 
+#### Function arguments
+Functions use `parlist`, which eventually uses `binding`. However, attempting to use key destructuring in a function body is not supported.
+
+```lua
+-- NOT supported
+local function f({ .x, .y })
+```
+
 #### Types
 An optional type can be supplied, such as:
 ```lua
@@ -112,12 +117,6 @@ Without explicit types, local assignments and for loops will assume the type of 
 local { .x, .y } = position :: { number }
 ```
 
-As a function argument, it will act the same as if writing it out by hand, where anonymous types are created:
-```lua
--- Will be f({+ x: a, y: b +})
-local function f({ .x, .y })
-```
-
 Additionally, you can specify a type on the "table" as a whole.
 
 ```lua
@@ -131,14 +130,21 @@ Combining both is acceptable, in which case the type on the variable takes prior
 local { .x: number, .y }: T = p
 ```
 
-### Non-designs
-#### Renaming variables
-This proposal does not allow something like JavaScript's:
-```js
-const { a: b } = t
+#### Variable renaming
+
+This proposal allows for renaming the assignments using `as`.
+
+```lua
+local { .real as aliased } = t
+-- is equivalent to...
+local aliased = t.real
 ```
 
-...where `t.a` is bound to `b`. This is not undesired, but this RFC is not concerned with it for now. This also means that any field you want to destruct must be a valid Luau identifier.
+This helps support mutliple assignments on the same name:
+```lua
+local { .name as nameA } = getObject(a)
+local { .name as nameB } = getObject(b)
+```
 
 #### Arrays
 This RFC does not concern itself with array destructuring. This is in part because it is not obvious what a reasonable syntax for it would be, and because its purposes might be better suited by tuples.
@@ -161,22 +167,48 @@ While with only this RFC, this is unambiguous with 2 lookahead (`{` + `.`), it c
 
 ## Drawbacks
 
+### Only identifiers as keys
+This syntax only proposes `.NAME`, which means it must be a valid name, even if renamed.
+
+The syntax could support `.["key with spaces"]`, but this may open up a can of worms about custom expressions (`.[expr]`). This proposal does not add support for it, though it also does not block it from being added.
+
+This also blocks nested destructuring, such as JavaScript's `const { a: { b } } = t` to mean `b = t.a.b`.
+
 ### Roblox - Property casing
-Today in Roblox, every index doubly works with camel case, such as `part.position` being equivalent to `part.Position`. This use is considered deprecated and frowned upon. However, without (or even with) variable renaming, this becomes significantly more appealing. For example, it is common you will only want a few pieces of informaiton from a `RaycastResult`, so you might be tempted to write:
+Today in Roblox, every index doubly works with camel case, such as `part.position` being equivalent to `part.Position`. This use is considered deprecated and frowned upon. However, even with variable renaming, this becomes significantly more appealing. For example, it is common you will only want a few pieces of informaiton from a `RaycastResult`, so you might be tempted to write:
 
 ```lua
-local { position } = Workspace:Raycast(etc)
+local { .position } = Workspace:Raycast(etc)
 ```
 
 ...which would work as you expect, but rely on this deprecated style.
 
 ## Alternatives
 
-### Syntax
+### Syntax for destructuring
 
 Many syntaxes have been proposed for destructuring. The most significant problem with any proposal is that is must be unambiguous to the reader whether or not the destructor is for **dictionaries** or for **arrays**.
 
 An intuitive suggestion is `local { a, b } = t`, but this syntax fails this test--it is not obvious if this is `local a = t.a` or `local a = t[1]`, regardless of whatever syntax is chosen for array destructuring (should it exist).
 
-### Focus more precisely, rather than using binding
-We could only allow this more precisely, such as just on local assignments. This would satisfy the use case of library importing and React properties (partially, since it would be its own line), but this would limit the use case of tables as arguments to functions, such as in the motivating HTTP request example.
+### Syntax for renaming
+
+`as` does not currently exist in Luau, and thus will not be intuitive to write. We could use `=`, but this doesn't produce a delightful syntax either.
+
+If we simply replace `as` with `=`, though it would parse, it does not read like any other assignment in Luau.
+
+```lua
+local { .x = y } = t
+```
+
+In Luau, variable assignments follow "name = assignment", whereas this flips it around as "assignment = name".
+
+If we flip it on the left...
+
+```lua
+local { y = .x } = t
+```
+
+...then it is no longer obvious how we might implement defaulting, such as JavaScript's `const { x = 5 } = obj;`. Note that this RFC does not prevent defaulting, but we may want it at some point.
+
+Furthermore, it would not be as easily expandable to something like nested destruction in terms of readability.
--- a/docs/syntax-key-destructuring.md
+++ b/docs/syntax-key-destructuring.md
@@ -115,7 +115,7 @@ local b: number = t.b
 Without explicit types, local assignments and for loops will assume the type of `<rhs>.<field>`. For example...
 ```lua
 -- x and y will both be typed `number` here
-local { .x, .y } = position :: { number }
+local { .x, .y } = position :: { x: number, y: number }
 ```
 
 Additionally, you can specify a type on the "table" as a whole.
--- a/docs/syntax-key-destructuring.md
+++ b/docs/syntax-key-destructuring.md
@@ -8,6 +8,7 @@ Introduce a new syntax for unpacking key values into their own variables, such t
 local { .a, .b } = t
 -- a == t.a
 -- b == t.b
+-- ...where t is evaluated once.
 ```
 
 ## Motivation
--- a/docs/syntax-key-destructuring.md
+++ b/docs/syntax-key-destructuring.md
@@ -1,0 +1,182 @@
+# Key destructuring
+
+## Summary
+
+Introduce a new syntax for unpacking key values into their own variables, such that:
+
+```lua
+local { .a, .b } = t
+-- a == t.a
+-- b == t.b
+```
+
+## Motivation
+
+Simple indexes on tables are very common both in and outside of Luau. A common use case is large libraries. It is common in the web world to see something like:
+
+```js
+const { useState, useEffect } = require("react");
+```
+
+...which allows you to quickly use `useState` and `useEffect` without fully qualifying it in the form of `React.useState` and `React.useEffect`. In Luau, if you do not want to fully qualify common React functions, the top of your file will often look like:
+
+```lua
+local useEffect = React.useEffect
+local useMemo = React.useMemo
+local useState = React.useState
+-- etc
+```
+
+...which creates a lot of redundant cruft.
+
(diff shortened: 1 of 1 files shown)


docs/syntax-key-destructuring.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+### Non-designs
+#### Renaming variables
+This proposal does not allow something like JavaScript's:
+```js
+const { a: b } = t
+```
+
+...where `t.a` is bound to `b`. This is not undesired, but this RFC is not concerned with it for now. This also means that any field you want to destruct must be a valid Luau identifier.
```
vegorov-rbx, line 141: Strange that it's not a concern.
I constantly see myself and other people renaming the fields when they use destructuring in C++.

With current proposal, you can't even destruct two objects of the same type in a single scope.

docs/syntax-key-destructuring.md
```diff
@@ -0,0 +1,182 @@
...
+	nextPageCursor,
+}) => { /* code */ })
+```
+
+## Design
+
+This RFC proposes expanding the grammar of `binding`:
+```patch
+binding = NAME [':' Type]
++         | `{` keydestructor [`,` keydestructor] [`,`] `}` [':' Type]
```
Kampfkarren, line 66: This is supposed to be zero or more `, keydestructor` I just am blanking on how to write that at the moment

docs/syntax-key-destructuring.md
Kampfkarren: > With current proposal, you can't even destruct two objects of the same type in a single scope.

Can you explain what you mean?

Kampfkarren: Do you mean like `const { position: { x, y } } = info()`?

vegorov-rbx: Here's one example from Luau code:
```
auto [leftType, leftRefinement] = check(scope, binary->left, relaxedExpectedLhs);
auto [rightType, rightRefinement] = check(scope, binary->right, expectedType);
```

Kampfkarren: Why wouldn't that work in the current proposal?

vegorov-rbx: Because local names are forced to be field names.

Kampfkarren: Oh, I see. Yes it is a drag--it is excluded in this RFC out of concern about significantly increasing the scope, while not having it still solves a large amount of use cases.

alexmccord: I would not consider C++ structured bindings to be totally isomorphic to the proposed table destructuring, even though it's similar in some ways. C++ structured bindings are _rename by default_, but this means you are forced to have as many binding names as there are fields. With this, you can pick a number of fields (and separately, optionally rename those fields you picked out).

docs/syntax-key-destructuring.md
alexmccord: Reminder that braces exists. (`{some grammar}` means parse that in a loop)

docs/syntax-key-destructuring.md
alexmccord: fwiw, while field renames are not uncommon in JavaScript, they're pretty rare from what I've seen?

alexmccord: But come to think of it, it is important to be able to rename a field in case one shadows the other, which definitely happens all the time.

docs/syntax-key-destructuring.md
goldenstein64: I guess it would realistically get parsed like this but more statefully?

```
| '{' [keydestructors1] '}' [':' Type]

keydestructors1 = keydestructor [keydestructors2]
keydestructors2 = ',' [keydestructors1]
```

It's more human-readable like this though.

```
| '{' {keydestructor ','} [keydestructor] '}' [':' Type]
```

docs/syntax-key-destructuring.md
```diff
@@ -0,0 +1,182 @@
...
+This would allow for the following:
+
+```lua
+local { .a, .b }, c = t
+
+for _, { .a, .b } in ts do
+end
+
+local function f({ .a, .b }, c)
+end
```
vegorov-rbx, line 80: Can we please drop this?
Not sure why it's even proposed (argument list bindings are a whole different construct from local bindings)
It increases complexity of the implementation and I don't think this had strong motivation for it (in many discussions, discussion always came up in context of locals, not argument lists).
This introduces unnamed function arguments.
It also raises questions of how many arguments the function really has. `local {.a, .b}` creates two locals, but `f({.a, .b})` is one argument.

Since overall the value of the proposal is already low (as in, just do local t = smth; just do local a, b = t.a, t.b; it's not that hard and you can enjoy all the renaming you need), the complexity of this moves the proposal further into the area of negative points it has going for itself.

alexmccord: I think I agree on some fronts. This makes people tempted to author functions that requires the user to heap-allocate a table only to throw it away after the function call ends.

```lua
local function say({.name, .message})
  print(`{name}: {message}`)
end

-- heap allocate a table and then discarded.
-- garbage collector is not having a fun day!
say{
  name = "Alex",
  message = "Hello, world!"
}
```

That said, I don't resonate with some of the counterarguments...

> This introduces unnamed function arguments.

I'm not sure this is a strong argument against.

> It also raises questions of how many arguments the function really has. local {.a, .b} creates two locals, but f({.a, .b}) is one argument.

I think it's obvious: 1 argument. It just is not named.

alexmccord:
Also worth noting that implementing destructuring requires fixing type refinements to work with tagged unions correctly. This works in TypeScript:

```ts
type Ok<T> = {tag: "ok", data: {val: T}}
type Err<E> = {tag: "err", data: {err: E}}
type Result<T, E> = Ok<T> | Err<E>

function f<T, E>(res: Result<T, E>) {
    const {tag, data} = res;
    if (tag == "ok") {
        let {val} = data; // val : T
    } else {
        let {err} = data; // err : E
    }
}
```

docs/syntax-key-destructuring.md
vegorov-rbx: > I'm not sure this is a strong argument against. Nobody bats an eye in C++, JavaScript, and even Rust (with irrefutable patterns)

It's a new concept for Luau compiler and debugger.
And this feature is not that great to have this going for it.

vegorov-rbx: And since we mentioned it, C++ had good taste to not include destructuring in argument list.

alexmccord:
One possible footgun that a user might run into. Not sure how much of a concern this is.

```lua
local t = { p = 5 }
local {.p} = t
p = 7

print(t.p == p) -- a user may reason that this is true, but it's false
```

docs/syntax-key-destructuring.md
Kampfkarren: I'm fine with it

Kampfkarren:
> Also worth noting that implementing destructuring requires fixing type refinements to work with tagged unions correctly. This works in TypeScript:
> 
> ```ts
> type Ok<T> = {tag: "ok", data: {val: T}}
> type Err<E> = {tag: "err", data: {err: E}}
> // type Ok<T> = {tag: "ok", val: T}
> // type Err<E> = {tag: "err", err: E}
> type Result<T, E> = Ok<T> | Err<E>
> 
> function f<T, E>(res: Result<T, E>) {
>     const {tag, data} = res;
>     // const {tag} = res;
>     if (tag == "ok") {
>         let {val} = data; // val : T
>         // let {val} = res; // val : T
>     } else {
>         let {err} = data; // err : E
>         // let {err} = res; // err : E
>     }
> }
> ```

I'm not sure this is a **requirement** per se, this seems like a forwards compatible addition? Code that does not pass checks before can now pass?

nezuo:
> One possible footgun that a user might run into. Not sure how much of a concern this is.

I don't think this is a footgun. You could do the same without destructuring:
```lua
local t = { p = 5 }
local p = t.p
p = 7

print(t.p == p)
```
Users might not know which types are copied by value but I don't see what that has to do with destructuring.

Are you saying a user might not think that these two things are equivalent?
```lua
local {.p} = t
local p = t.p
```

alexmccord:
> Are you saying a user might not think that these two things are equivalent?

Yeah, that's what I'm thinking. I only mentioned it because it came across my mind.

docs/syntax-key-destructuring.md
alexmccord: Another possible argument against it: there is no place in the syntax to do key destructuring on
 `self` despite that it's still as useful too. I just wrote a function today that only uses `ZoomLevel` from `self`.

```lua
function Camera:GetCameraOffset() -- Where to say `{.ZoomLevel}` as `self`?
    local {.ZoomLevel} = self
    return 
end
```

docs/syntax-key-destructuring.md
Kampfkarren: I thought `{}` was one-or-more, sounds good.

docs/syntax-key-destructuring.md
vegorov-rbx: Just want to note that we can always propose additions in the future to extend to function arguments.

We can see how developers will start using this and if it's often used to destructure function argument at the start of the function, that will be a good signal to consider how to make it possible.

Even for loop can be added later, but I think we'll be ok with it now (even though that requires separate implementation work from local statement).

BradSharp:
I am not the biggest fan of the syntax. I like the idea, but given Luau has multiple returns I feel like we could come up with something nicer. Perhaps:
```
local x, y = x, y in foo
```

docs/syntax-key-destructuring.md
```diff
@@ -161,22 +167,48 @@ While with only this RFC, this is unambiguous with 2 lookahead (`{` + `.`), it c
 
 ## Drawbacks
 
+### Only identifiers as keys
+This syntax only proposes `.NAME`, which means it must be a valid name, even if renamed.
+
+The syntax could support `.["key with spaces"]`, but this may open up a can of worms about custom expressions (`.[expr]`). This proposal does not add support for it, though it also does not block it from being added.
```
alexmccord, line 173: I think if we do this, we don't need the `.` part. The `[` replaces `.`.

docs/syntax-key-destructuring.md
```diff
@@ -161,22 +167,48 @@ While with only this RFC, this is unambiguous with 2 lookahead (`{` + `.`), it c
...
 
 An intuitive suggestion is `local { a, b } = t`, but this syntax fails this test--it is not obvious if this is `local a = t.a` or `local a = t[1]`, regardless of whatever syntax is chosen for array destructuring (should it exist).
 
-### Focus more precisely, rather than using binding
-We could only allow this more precisely, such as just on local assignments. This would satisfy the use case of library importing and React properties (partially, since it would be its own line), but this would limit the use case of tables as arguments to functions, such as in the motivating HTTP request example.
+### Syntax for renaming
+
+`as` does not currently exist in Luau, and thus will not be intuitive to write. We could use `=`, but this doesn't produce a delightful syntax either.
+
+If we simply replace `as` with `=`, though it would parse, it does not read like any other assignment in Luau.
```
alexmccord, line 198: Note that in JavaScript the `=` has different semantics: if the key is missing from the object, then replace the key by the expression to the right of the `=`.

surfbryce:
> I am not the biggest fan of the syntax. I like the idea, but given Luau has multiple returns I feel like we could come up with something nicer. Perhaps:
> 
> ```
> local x, y = x, y in foo
> ```

From my perspective the initially proposed syntax is great and universal. If you had no concept of de-structuring, you would be able to grasp what this syntax roughly does due to its consistent/borrowed use of functional syntax from elsewhere in the language (with their functions representing the very same that make up the underlying concept itself). 

Your syntax proposal is primarily restricted to traditional variable initialization; how do you handle de-structuring a provided for loop value (which is one of the provided use-cases) with this syntax?

Additionally, I think there's significantly more potential for confusion from "x, y in foo" than from "{.x, .y} = foo": given that users may interpret this as using two independent variables - "x" and "y" - rather than indexing foo with "x" and "y".

docs/syntax-key-destructuring.md
aatxe: Honestly, I think it might be reasonable to say that this is the good bit of "increasing the scope" for this RFC, and e.g. the argument destructuring was a bad case. It's true that both can be added as a follow-up later, but I think renaming addresses a much more serious need.

AsynchronousAI:
> I am not the biggest fan of the syntax. I like the idea, but given Luau has multiple returns I feel like we could come up with something nicer. Perhaps:
> 
> ```
> local x, y = x, y in foo
> ```

I like this syntax more, but writing `x, y = x, y` can be repetitive maybe another shorthand:
```lua
local x, y in foo
-- same as:
local x2, y2 = x, y in foo
```

Cooldude2606:
_Taken from [previous discussion](https://github.com/luau-lang/luau/pull/629) on this topic._

An alternative to placing the braces on the lefthand side is to instead create a "tuple returning index" operation as follows:

```lua
local color = Color3.new()
local r, g, b = color.{ R, G, B }
--local r, g, b = color.R, color.G, color.B
```

This pattern reads well and allows the properties to be assigned to variable names which differ from the property name without adding new keywords or symbols that could be seen as setting a default.
It also allows for the destructing of a table to work outside of assignment, such as passing arguments to a function:

```lua
local camera = Instance.new("Camera")
local cameraData = { Position = Vector3.new(), Target = Vector3.new(1, 0, 0) } 
camera.CFrame = CFrame.new(cameraData.{ Position, Target })
--camera.CFrame = CFrame.new(cameraData.Position, cameraData.Target)
```

And as suggested by @RigidStudios it can support the same syntax used in table definations.

```lua
-- Where a table can be defined as
local indexStr = "foo"
local tab = {
    [1] = 0,
    [2] = 1,
    [indexStr] = 2
    bar = 3
}

-- And can be destructured as
local a, b, thing, bar = tab.{ [1], [2], [indexStr], bar }
-- local a, b, thing, bar = tab[1], tab[2], tab[indexStr], tab.bar
```

However, it does not work well on existing tuples such as those in for loops and would require an extra line of code. (@surfbryce)

```lua
for _, cameraData in pairs(allCameraData) do
    local position, target = cameraData.{ Position, Target }
end
```
Being required to always provide the lefthand side could be seen a repetitive. (@AsynchronousAI) Although the repetition would be required in most cases, regardless of the syntax, to avoid reliance on deprecated property access of classes in Roblox. The main issue is repetition of names when importing modules or deconstructing function returns.

```lua
-- for function returns
local position, target = getCameraData(a).{ Position, Target }
-- for long imports
local useContext, useRef, useState, useEffect = require("React")
    .{ userContext, useRef, useState, useEffect }
```

The repeition in long imports kills what is otherwise a great syntax alternative that supports more than the propossed while being simpler (effectively syntax sugar). Although, I recall seeing another [RFC](https://github.com/luau-lang/luau/pull/1013) discussing the inclusion of `import ... from ...` syntax for first-class importing which may return in the future, which would avoid this issue for all syntaxs.

Kampfkarren:
If you have to specify both sides, then not only can you no longer use it in function arguments, but you've also barely reduced the actual code.

Kampfkarren:
I have the same problem with `in`--we lock ourselves out of any binding that isn't a local assignment which is not good.

docs/syntax-key-destructuring.md
```diff
@@ -0,0 +1,214 @@
...
+This is to avoid ambiguity with potential table calls:
+
+```lua
+local a = b
+{ .x, .y } = c
+```
+
+While with only this RFC, this is unambiguous with 2 lookahead (`{` + `.`), it could potentially complicate our ability to expand the key destructuring syntax, such as to add renaming variables.
+
+## Drawbacks
```
potatoprank, line 168: IMO these drawbacks will come up more often than not and raise the cognitive load required to properly use and understand this syntax.

That seems counter to the intent of this proposal which is to make things more simple and convenient to use.

docs/syntax-key-destructuring.md
```diff
@@ -0,0 +1,214 @@
...
+
+...which would work as you expect, but rely on this deprecated style.
+
+## Alternatives
+
+### Syntax for destructuring
+
+Many syntaxes have been proposed for destructuring. The most significant problem with any proposal is that is must be unambiguous to the reader whether or not the destructor is for **dictionaries** or for **arrays**.
+
+An intuitive suggestion is `local { a, b } = t`, but this syntax fails this test--it is not obvious if this is `local a = t.a` or `local a = t[1]`, regardless of whatever syntax is chosen for array destructuring (should it exist).
```
potatoprank, line 192: thanks for this example! I was wondering why we chose the `.NAME` syntax. Are you also wanting to propose syntax for destructuring arrays or are you just saying that the syntax shouldn't also look like it could be reasonably used for destructuring arrays?

potatoprank (commented):
I'm a -1 on this. It's introducing new syntax with some odd edge cases that I think will ultimately make working with this syntax more trouble than it's worth.

IMO the benefit of this syntax is pretty marginal. It's mainly just:

```js
const { useState, useEffect } = require("react");
```

I would personally not want to encourage stuff like

```js
const { position, style } = props
```

since it raises the chance of name shadowing related issues.

potatoprank (commented):
BTW @potatoprank  = Peter Lu

docs/syntax-key-destructuring.md
aatxe: There's good reason to want to leave room open for destructuring arrays as well, and also good arguments for not bothering (e.g. that `table.unpack` is already essentially an array destructuring function and gives you the ability to just write a tuple of bindings on the left-hand side).

alexmccord (approved):

andyfriesen:
Is this also proposing array destructuring?

> Without explicit types, local assignments and for loops will assume the type of `<rhs>.<field>`. For example...
```lua
-- x and y will both be typed `number` here
local { .x, .y } = position :: { number }
```

How does this work?

docs/syntax-key-destructuring.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Only identifiers as keys
+This syntax only proposes `.NAME`, which means it must be a valid name, even if renamed.
+
+The syntax could support `.["key with spaces"]`, but this may open up a can of worms about custom expressions (`.[expr]`). This proposal does not add support for it, though it also does not block it from being added.
+
+This also blocks nested destructuring, such as JavaScript's `const { a: { b } } = t` to mean `b = t.a.b`.
+
+### Roblox - Property casing
+Today in Roblox, every index doubly works with camel case, such as `part.position` being equivalent to `part.Position`. This use is considered deprecated and frowned upon. However, even with variable renaming, this becomes significantly more appealing. For example, it is common you will only want a few pieces of informaiton from a `RaycastResult`, so you might be tempted to write:
```
vegorov-rbx, line 179:
```suggestion
Today in Roblox, every index doubly works with camel case, such as `part.position` being equivalent to `part.Position`. This use is considered deprecated and frowned upon. However, even with variable renaming, this becomes significantly more appealing. For example, it is common you will only want a few pieces of information from a `RaycastResult`, so you might be tempted to write:
```

docs/syntax-key-destructuring.md
```diff
@@ -0,0 +1,215 @@
...
+local { .position } = Workspace:Raycast(etc)
+```
+
+...which would work as you expect, but rely on this deprecated style.
+
+## Alternatives
+
+### Syntax for destructuring
+
+Many syntaxes have been proposed for destructuring. The most significant problem with any proposal is that is must be unambiguous to the reader whether or not the destructor is for **dictionaries** or for **arrays**.
```
vegorov-rbx, line 191:
```suggestion
Many syntaxes have been proposed for destructuring. The most significant problem with any proposal is that it must be unambiguous to the reader whether or not the destructor is for **dictionaries** or for **arrays**.
```

Kampfkarren:
> This section looks to be about array destructuring. Is this a mistake?
> 
> > Without explicit types, local assignments and for loops will assume the type of `<rhs>.<field>`. For example...
> 
> ```lua
> -- x and y will both be typed `number` here
> local { .x, .y } = position :: { number }
> ```

It's about object destructuring, just a typo in the type

docs/syntax-key-destructuring.md
```diff
@@ -0,0 +1,215 @@
...
+
+-- Equivalent to...
+local a: string = t.a
+local b: number = t.b
+```
+
+Without explicit types, local assignments and for loops will assume the type of `<rhs>.<field>`. For example...
+```lua
+-- x and y will both be typed `number` here
+local { .x, .y } = position :: { number }
```
Kampfkarren, line 118:
```suggestion
local { .x, .y } = position :: Position
```

andyfriesen:
Hey everyone.

This RFC triggered a whole bunch of last-minute internal discussion and we have decided not to include this feature at this time.

One of the difficult things about syntax extensions to programming languages is that reversing course after the fact is essentially impossible.  This RFC does an admirable job of navigating constraints passed down to us from Lua and presenting a least-worst solution to this specific problem, but it doesn't handle assignments or array destructuring and there doesn't seem to be any path at all to get there.

We need the bar to be higher than least-worst if we're going to successfully maintain Luau's identity as a small, predictable language.

PR #26, opened 2024-02-24
Shorthand field initialization syntax

4teapo:
[Rendered](https://github.com/4teapo/luau-rfcs/blob/master/docs/syntax-shorthand-field-init.md)

### DIFF
--- a/docs/syntax-shorthand-field-init.md
+++ b/docs/syntax-shorthand-field-init.md
@@ -1,0 +1,157 @@
+# Shorthand field initialization
+
+## Summary
+
+This RFC proposes a shorthand syntax for declaring fields in tables, which allows writing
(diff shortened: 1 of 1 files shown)


potatoprank (commented):
I appreciate the motivation and I think language level solution for it may appropriate. I'm mainly concerned with exactly the drawbacks you outlined. I'd avoid introducing permanent syntax complexity just to make certain common patterns slightly nicer.

alexmccord:
I'm not sold on this. This is probably pushing the line of syntactic conveniences.

alexmccord:
Since the other RFC on destructuring was ultimately rejected, I think this RFC is also bound by the same fate since if we were doing this, we'd need a consistent story on the syntax with destructuring.

PR #28, opened 2024-04-11
RFC: `buffer.random` for cryptographically secure randomness

3dsboy08:
[Rendered](https://github.com/3dsboy08/rfcs/blob/buffer-random/docs/function-buffer-random.md)

### DIFF
--- a/docs/function-buffer-random.md
+++ b/docs/function-buffer-random.md
@@ -1,0 +1,36 @@
+# buffer.random
+
+## Summary
+
+Add `buffer.random` as a function to generate cryptographically secure random bytes.
+
+## Motivation
+
+There is no mechanism in Luau for cryptographically secure random bytes, which is a requirement for cryptographic libraries implemented in Luau.
+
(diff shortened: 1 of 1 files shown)


joeengo:
when is v3 dropping

vegorov-rbx:
After discussion in the Luau team, we don't think this is an appropriate addition to the Luau standard library.

We want Luau to continue to be a small embeddable runtime library with support for many platforms and unfortunately, generating cryptographically secure randomness is not a feature of C++ language that we are based on and is hard to support on all platforms we target.

It is also strange to single out only this feature. While we understand that many other features of the JS 'crypto' library can be hand-coded on top and this is one of the functions that requires native support, it would be best to approach cryptography with a full API.

Having said that, full API will likely require even more dependencies and if you are using Luau in your own project, you always have the option to extend the environment with additional native libraries.

If you happen to develop on the Roblox platform, it might be best to request that as an engine API at https://devforum.roblox.com/c/feature-requests/engine-features/23
As Roblox already has dependencies for hashing and signing, it is easier to provide them as a service there.

As an extra note, while resource exhaustion is not considered for Luau open-source security model, it is certainly something that Roblox still considers for theirs.

PR #29, opened 2024-04-16
RFC: Negation types

alexmccord:
[Rendered](https://github.com/luau-lang/rfcs/blob/negation-types/docs/negation-types.md).

### DIFF
--- a/docs/negation-types.md
+++ b/docs/negation-types.md
@@ -55,7 +55,7 @@ As you can see, since through the series of rewrites `~~any` did not produce a t
 
 We currently only support negation of _testable_ types. Intuitively, a testable type is one that can be tested by type refinements at runtime. This includes singleton types, primitive types, and classes. Of course, unions and intersections can be negated as well if and only if all of its constituent types are also testable types, due to the distributivity property of unions and intersections.
 
-Since types like `{ p: P }` and `(T) -> U` are structural types and the runtime offers no way to test for them, they cannot be negated. An attempt to negate this will just turn the negation type into an error suppressed type. This means `~{ p: P }` and `~(T) -> U` are nonsensical. In the event that you negate some non-testable types, you get a type error.
+Since types like `{ p: P }` and `(T) -> U` are structural types and the runtime offers no way to test for them, they cannot be negated. An attempt to negate this will just turn the negation type into an error suppressed type. This means `~{ p: P }` and `~(T) -> U` are nonsensical. In the event that you negate some non-testable type, you get a type error.
 
 Another restriction is that negation types cannot negate generic types. So `<T>(T) -> ~T` is also not legible.
 
--- a/docs/negation-types.md
+++ b/docs/negation-types.md
@@ -8,7 +8,7 @@ A type that represents the complement of the type being negated, which is a unio
 
 Type refinements will produce negation types to get the complementation. For the most part, users will never need to write these negation types themselves, except for when they want to _maintain_ some invariants beyond the scope of the branch.
 
-A [cofinite set](https://en.wikipedia.org/wiki/Cofiniteness) arises when you have a negation type in conjunction with intersection types. For example, `string & ~"a"` is a cofinite string set that accepts any `string` excluding `"a"`. This happens often with type refinements where `typeof(x) == "string" and x ~= "a"` gives rise to the type `string & ~"a"`.
+A [cofinite set](https://en.wikipedia.org/wiki/Cofiniteness) arises when you have some negated `X` in conjunction with some supertype of `X`. For example, `string & ~"a"` is a cofinite string set that accepts any `string` excluding `"a"`. This happens often with type refinements where `typeof(x) == "string" and x ~= "a"` gives rise to the type `string & ~"a"`.
 
 ## Design
 
--- a/docs/negation-types.md
+++ b/docs/negation-types.md
@@ -20,7 +20,7 @@ We propose to allow users to write such a type. Formally:
 
 ```ebnf
 ty ::= `~` ty
-     | name [`<` [ty] {`,` ty} `>`]
+     | name [`<` [ty {`,` ty}] `>`]
      | ty {`|` ty}
      | ty {`&` ty}
      | `(` ty `)`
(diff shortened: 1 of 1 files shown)


docs/negation-types.md
```diff
@@ -0,0 +1,86 @@
+# Negation types
+
+## Summary
+
+A type that represents the complement of the type being negated, which is a union type of all inhabited types (`unknown`) excluding the type being negated.
+
+## Motivation
+
+Type refinements will produce negation types to get the complementation. For the most part, users will never need to write these negation types themselves, except for when they want to _maintain_ some invariances beyond the scope of the branch.
```
aatxe, line 9:
```suggestion
Type refinements will produce negation types to get the complementation. For the most part, users will never need to write these negation types themselves, except for when they want to _maintain_ some invariants beyond the scope of the branch.
```

docs/negation-types.md
```diff
@@ -0,0 +1,86 @@
...
+
+A [cofinite set](https://en.wikipedia.org/wiki/Cofiniteness) arises when you have a negation type in conjunction with intersection types. For example, `string & ~"a"` is a cofinite string set that accepts any `string` excluding `"a"`. This typically happens all the time with type refinements, where `typeof(x) == "string" and x ~= "a"` gives rise to the type `string & ~"a"`.
```
aatxe, line 11:
```suggestion
A [cofinite set](https://en.wikipedia.org/wiki/Cofiniteness) arises when you have a negation type in conjunction with intersection types. For example, `string & ~"a"` is a cofinite string set that accepts any `string` excluding `"a"`. This happens often with type refinements where `typeof(x) == "string" and x ~= "a"` gives rise to the type `string & ~"a"`.
```

docs/negation-types.md
```diff
@@ -0,0 +1,86 @@
...
+3. `~(~unknown & ~*error*)`
+4. `~(never & *error*)`
+5. `~never | ~*error*`
+6. `unknown | *error*` (`any` is an alias to `unknown | *error*`, so equivalence is achieved)
+
+### Restrictions
+
+We currently only support negation of _testable_ types. Intuitively, a testable type is one that can be tested by type refinements at runtime. This includes singleton types, primitive types, and classes. Of course, unions and intersections can be negated as well if and only if all of its constituent types are also testable types, due to the distributivity property of unions and intersections.
+
+Since types like `{ p: P }` and `(T) -> U` are structural types and the runtime offers no way to test for them, they cannot be negated. An attempt to negate this will just turn the negation type into an error suppressed type and also a type error. This means `~{ p: P }` and `~(T) -> U` are nonsensical.
```
aatxe, line 58: We should probably explicitly error if a user writes an annotation that is the negation of a structural type.

alexmccord: There was "and also a type error" but I can extract that into a separate sentence in this paragraph.

docs/negation-types.md
```diff
@@ -0,0 +1,86 @@
...
+
+1. the syntax,
+2. some guarantee that no negation types survive if it negates non-testables, and
+3. a type error if it negates a non-testable type in any way, shape, or form.
+
+The parser change for this is trivial, so this is of no concern.
+
+We can use type families to have this guarantee. Add a `negate<T>` type family which would be internal to the type inference engine, and have the syntax `~T` produce that type family, not a `NegationType`.
+
+As for the type error, we can just consider `negate<{ p: P }>` to be an uninhabited type family, and resolve that as an error type.
```
aatxe, line 74: I think this is very sensible, and aligned with where future proposals will go too (e.g. type indexing syntax like `SomeTableType.SomeProperty` will probably want to resolve to `index<SomeTableType, SomeProperty>`). Seems like the right call to me.

docs/negation-types.md
alexmccord: Word ordering is super important and it's rather annoying. Function type. Type function. Refinement type. Type refinement.

docs/negation-types.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Syntax
+
+The notation is `~T` where `T` is a type annotation. It binds tightly to the type it negates, so `~T | U` is `(~T) | U`, not `~(T | U)`. There's really not much to it, syntax-wise. It has the same precedence as the `not` expression, which is why `~T | U` is `(~T) | U` in the same way `not t or u` is `(not t) or u`.
+
+We propose to allow users to write such a type. Formally:
+
+```ebnf
+ty ::= `~` ty
+     | name [`<` [ty] {`,` ty} `>`]
```
alexmccord, line 23:
```suggestion
     | name [`<` [ty {`,` ty}] `>`]
```

Wunder-Wulfe:
I believe this has been partially implemented through `types.negationof()`, at least from what I see on the Roblox side of Luau Beta.
Previously, this resulted in a `Type Error`, as the resulting type class was not recognized by Luau, but now it works and the following code should function as intended:

```luau
--- Negate the current type
type function isnt(t)
	return types.negationof(t)
end

local x: isnt<number> = "test";
x = vector.create(1,2,3);
x = 2; -- Type Error: Type 'number' could not be converted into '~number'; type number (number) is not a subtype of ~number.negated() (number)

local y: isnt<vector | string> = 2
if typeof(y) == "string" then
	print(y) -- y: never
end
y = vector.create(3,2,1) -- Type Error: Type 'vector' could not be converted into '~(string | vector)'; type vector (vector) is not a subtype of ~(string | vector).negated() (string | vector)


local z: isnt<isnt<number>> = 2 -- z: ~~number
z = "hello world" -- Type Error: Type 'string' could not be converted into '~~number'; type string (string) is not a subtype of ~~number.negated() (~number)
```

All we are missing is the implementation of the `~` or `not` operator, or a dedicated `negate<T>` or `isnt<T>` type function, and the optimization to collapse `~~T` expressions (`(~(~(T)))`) back into `T` (`(T)`).

aatxe:
Negation types were already implemented when this RFC was written too. This RFC is exclusively a proposal for _syntax_ for writing them.

aatxe (approved):

PR #30, opened 2024-04-16
RFC: Function Attribute Parameters

aviralg:
[Rendered](https://github.com/aviralg/rfcs/blob/attribute-argument-syntax/docs/syntax-attributes-functions-parameters.md)


docs/syntax-attributes-functions-parameters.md
```diff
@@ -0,0 +1,32 @@
+# Function Attribute Parameters
+
+## Summary
+
+This RFC proposes a syntax for function attribute parameters. This is a follow up to the [Attributes](https://rfcs.luau-lang.org/syntax-attributes-functions.html) RFC, which proposed a design for function attributes. We only propose a syntactic extension in this RFC; the rest of the implementation details remain unchanged.
```
alexmccord, line 5: We could make this a relative path to generalize for GitHub and the RFC site.
```suggestion
This RFC proposes a syntax for function attribute parameters. This is a follow up to the [Attributes](./syntax-attributes-functions) RFC, which proposed a design for function attributes. We only propose a syntactic extension in this RFC; the rest of the implementation details remain unchanged.
```

aviralg: Fixed

docs/syntax-attributes-functions-parameters.md
```diff
@@ -0,0 +1,32 @@
...
+## Design
+
+The following syntax is proposed for attributes with parameters:
+
+```ebnf
+attribute = '@' NAME 
+          | '@' NAME '(' ATOMIC-LITERAL (',' ATOMIC-LITERAL)* ')'
+```
+
+The only extension proposed to the [Attributes](./syntax-attributes-functions.md) RFC is in the second line which allows one or more comma-separated literals to be supplied to the attribute. The `ATOMIC-LITERAL` category includes values of type `boolean`, `number`, `string`, and the `nil` value. There are three notable exclusions in this syntax.
```
Dekkonot, line 20: Is there a valid reason to support `nil` as a parameter value? It seems counterintuitive.

aviralg: Having `nil` gives us the ability to have attributes with "optional" parameters. If no valid value exists, the parameter will be `nil`. Take the case of `@deprecated(<new-function-name>, <warning-message>)` attribute. Assume a function is being marked deprecated because we are getting rid of the feature entirely and there is no other function replacing it. In that case, we can pass `nil` for the `<new-function-name>` parameter. An alternative would be to allow this attribute to have either one or two parameters. But that still won't help when an attribute can take up to three parameters and we want to ignore the second.

docs/syntax-attributes-functions-parameters.md
```diff
@@ -0,0 +1,34 @@
...
+
+There are three notable exclusions in this syntax. 
+
+1. Literal `table` values are excluded from being passed as parameters since they are composite structures; information supplied via table fields can be passed as separate parameters to the attribute. If they are still deemed useful, a follow-up RFC can introduce them with motivating use cases.
```
alexmccord, line 24: A motivating example is literally in the paragraph above, `@deprecated`. Sometimes you want to say:

* this symbol is deprecated: `@deprecated`
* this symbol is deprecated with a replacement: `@deprecated { use = "foo" }`
* this symbol is deprecated for a reason: `@deprecated { reason = "..." }`
* this symbol is deprecated with a replacement and a reason: `@deprecated { use = "foo", reason = "..." }`

I think in the case of `@deprecated`, you really don't know which argument `use` or `reason` will be most used, so if you get the parameter ordering wrong, you end up with a lot of redundant `nil` argument in all use sites, and potentially shoot yourself in the foot because you forgot to add `nil` arguments i.e. `@deprecated("my cool reason")`.

docs/syntax-attributes-functions-parameters.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Design
+
+The following syntax is proposed for attributes with parameters:
+
+```ebnf
+attribute = '@' NAME 
+          | '@' NAME '(' ATOMIC-LITERAL (',' ATOMIC-LITERAL)* ')'
```
alexmccord, line 17: What happens with this statement? We need some way to delimit the attribute from the `(`.

```lua
@my_cool_attribute
(f()).x = 7
```

alexmccord: And no, significant whitespace is _not_ the answer to this.

docs/syntax-attributes-functions-parameters.md
aviralg: I chose to keep the argument type simple because attributes are not user-extensible and we have only one use case for optional attribute arguments. I deliberately decided to be conservative and leave the door open for a future RFC to introduce table syntax if really needed. In this particular instance, at max, 2 `nil` arguments need to be supplied which feels fine to me. The table syntax is bulkier than I would like.

docs/syntax-attributes-functions-parameters.md
aviralg: This RFC is implementing attributes for function declarations only. `(f()).x = 7` is not a function declaration. 
If we want the freedom to attach attributes to arbitrary statements, we need another RFC.

docs/syntax-attributes-functions-parameters.md
aviralg: If we want table parameters, then there is no point in having multi-parameter attributes. Table can be used to supply all the parameters (with or without names). Attributes can be implemented to optionally accept a single table argument (without surrounding `(` and `)`).

docs/syntax-attributes-functions-parameters.md
alexmccord: That's true, but my whole charge is ensuring that the language is forward compatible for the rest of its evolution within the same edition. The syntax as is doesn't allow us this affordability.

aviralg: We know ahead of time if an attribute takes parameters. For instance `@native` does not but `@deprecated` does. The parser will interpret a `(f())` followed by `@deprecated` as an attribute parameter and throw an error since `f()` is not a valid parameter category. 
In general, the ambiguity you mention will only happen if we choose to implement an attribute to have both a parameter and a parameter-less form. In other words, we can avoid this ambiguity, always.

aviralg: I am wondering if I should rework the proposal as follows:

1. Attributes are allowed an optional table parameter (only one parameter which is always a table and nothing else).
2. Attributes without parameters start with a single `@`, like `@native`.
3. Attributes with a parameter start with two `@` symbols, like `@@deprecated <parameter>`.

The use of `@` and `@@` removes the ambiguity you mentioned. The use of a table parameter makes it as general as it could be. I like `@@deprecated {use = "...", reason = "..."}` a lot more over `[deprecated {use = "...", reason = "..."}]`.

aviralg: A simple way to avoid ambiguity, if we choose to support attributes on arbitrary expression, is to introduce empty parameter list for the attribute. For instance, the original example can be written like this:

```
@my_cool_attribute()
(f()).x = 7
```

aviralg:
Reworked the document to describe the new syntax for attributes.

docs/syntax-attributes-functions-parameters.md
```diff
@@ -6,48 +6,56 @@ This RFC proposes a syntax for function attribute parameters. This is a follow u
...
-The following syntax is proposed for attributes with parameters:
+The following syntax is proposed for attributes:
 
 ```ebnf
-attribute = '@' NAME 
-          | '@' NAME '(' ')'
-          | '@' NAME '(' ATOMIC-LITERAL (',' ATOMIC-LITERAL)* ')'
-```
+parameter-table = '{' '}'
+          | '{' parameter (sep parameter)* '}'
```
alexmccord, line 17: We will want an optional trailing `sep` too.

alexmccord, line 17: Usual zero-to-many repetition syntax in EBNF is `{}`
```suggestion
| '{' parameter {sep parameter} '}'
```

docs/syntax-attributes-functions-parameters.md
```diff
@@ -6,48 +6,56 @@ This RFC proposes a syntax for function attribute parameters. This is a follow u
...
+literal = BOOLEAN | NUMBER | STRING | NIL | parameter-table
 
-```lua
-@my_attribute()
-(f().x)
+list-attribute = NAME parameter-table
+               | NAME
+
+attribute = '@' NAME 
+          | '@[' list-attribute (',' list-attribute)* ']'
```
alexmccord, line 30: Ditto.
```suggestion
          | '@[' list-attribute {',' list-attribute} ']'
```

alexmccord, line 30: This also means `@[` is inseparable, which would mean a different token from `@`. Is that what you want?

```lua
@ [native]
function f() end
```

docs/syntax-attributes-functions-parameters.md
```diff
@@ -6,48 +6,56 @@ This RFC proposes a syntax for function attribute parameters. This is a follow u
...
+          | NAME '=' literal
 
-Since an opening parenthesis starts a parameter list for the attribute, the expression `(f().x)` will be treated as a parameter list of `@my_attribute`. This will lead to a syntax error since attribute parameter can only be `ATOMIC-LITERAL` and `f().x` is not one. This ambiguity can be resolved by adding `()` to `@my_attribute` as follows:
+literal = BOOLEAN | NUMBER | STRING | NIL | parameter-table
 
-```lua
-@my_attribute()
-(f().x)
+list-attribute = NAME parameter-table
+               | NAME
```
alexmccord, line 27: This doesn't offer a way to say `@[attr(5)]`. I assume you want this?
```suggestion
arguments-attribute = '(' {parameter sep} [parameter [sep]] `)`

list-attribute = NAME parameter-table
               | NAME arguments-attribute
               | NAME
```

alexmccord: ah. `parameter` is `literal` (which is fine) but _also_ allows `NAME = literal`! Grammar wise, that's strange to me. Can we rename `parameter` to `attr-arg-table-field` or something short?

alexmccord: I'll revisit the grammar next week. I was hoping we would model attributes to look like function calls, so e.g.

```lua
@[attr "something"]
@[deprecated { use = "something" }]
@[unroll(5)]
```

docs/syntax-attributes-functions-parameters.md
alexmccord: There's also an existing `fieldsep` grammar in Luau for table field separator.

docs/syntax-attributes-functions-parameters.md
```diff
@@ -0,0 +1,61 @@
...
+literal = BOOLEAN | NUMBER | STRING | NIL | parameter-table
+
+list-attribute = NAME parameter-table
+               | NAME
+
+attribute = '@' NAME 
+          | '@[' list-attribute (',' list-attribute)* ']'
+```
+
+In Luau scripts, attributes are specified before the `function` keyword in function definitions. In declaration files, attributes are specified in function type declarations, before the `function` keyword in `declare function (type, ...) : type` syntax, and before the `(` in the `name: (type, ...) : type` syntax. The primary extension proposed to the [Attributes](./syntax-attributes-functions.md) RFC is a new delimited syntax `@[]` for specifying multiple comma-separated attributes with parameters supplied through an optional Luau table. The important features are:
```
alexmccord, line 33: We discussed this internally, so this part of the RFC needs changing/removed.

```
declare @checked function f(): ()
local @native function f() end
```
is just weird. Let's put them before statements, not interpolated in them.

docs/syntax-attributes-functions-parameters.md
```diff
@@ -0,0 +1,61 @@
...
+
+1. Inside the attribute list, `@[]`, attribute names are not allowed to have a leading `@`.
+2. Attribute lists cannot be nested. Attributes cannot be specified on attribute parameters.
+3. Attributes inside `@[]` are separated by a `,`.
+4. Attributes inside `@[]` are allowed to take an optional table of parameters.
+5. The "parameter table" is a Luau table whose entries specify attribute parameters, with or without names, separated by commas or semicolons.
```
alexmccord, line 39: In table arguments, entries should always have a name. If you wanted no names, use `()`.

docs/syntax-attributes-functions-parameters.md
```diff
@@ -0,0 +1,61 @@
...
+7. Attributes can be specified before function declarations in multiple ways: `@attr1 @[attr2, attr3{2, "hi"}]`, `@attr1 @attr2 @[attr3{2, "hi"}]`, `@attr1 @[attr3{2, "hi"}] @attr2`, and  `@[attr1, attr2, attr3{2, "hi"}]` are all equivalent.
+8. An attribute with an empty parameter list can be equivalently specified without one, i.e., `@[attr{}]` is same as `@attr`.
+
+The parser is responsible for for enforcing the syntax specified by this RFC and ensuring that attributes are not repeated on a definition or a declaration. The number and type of parameters accepted by a particular attribute will be specified by its own RFC, and enforced by the relevant component of the implementation (typechecker, compiler, etc.), after parsing.
```
alexmccord, line 44: So if I had

```lua
@unroll("five")
```

The parser is not going to reject this (which is fair enough) but I won't see any errors about it, the compiler has two options: silently ignore `unroll` request, or throw a compilation error which currently does not get shown to the user.

This would mean the linter will have to analyze these attributes and whether they have the correct types so that users get warnings about incorrect arguments.

fwiw, parser rejecting duplicate attributes is not far-fetched. We do this already with table indexers. `{ [A]: B, [C]: D }` is a parse error, it doesn't use the last indexer.

docs/syntax-attributes-functions-parameters.md
```diff
@@ -0,0 +1,61 @@
...
+
+We currently have an ad-hoc implementation of `@checked` attribute, used in declaration files. It is specified in two ways:
+
+1. `declare function @checked abs(n: number): number`
+2. `writef64: @checked (b: buffer, offset: number, value: number) -> ()`
+
+To ensure uniformity with the specification of this RFC, declarations using the first style will be modified to use `@checked` attribute before the `function` keyword. Declarations using the second style will remain unchanged.
```
alexmccord, line 51: Ditto here: https://github.com/luau-lang/rfcs/pull/30/files#r1605633289.

docs/syntax-attributes-functions-parameters.md
```diff
@@ -0,0 +1,61 @@
...
+
+An alternative parameter syntax would be to implement function-style attribute syntax of the form `@attr(parameters, ...)`, but that would not allow specifying parameter names. Attribute syntax of the form `@attr {...}` will create ambiguity when attributes without parameters are used on tables, should we decide to allow attributes on them in the future. An advantage of the `@[]` syntax is that it provides the `]` delimiter which marks the end of attribute, leaving open the possibility of syntax evolution without introducing parsing ambiguity.
```
alexmccord, line 53: Looks like this belongs in `Alternatives` section.

docs/syntax-attributes-functions-parameters.md
alexmccord: It also violates the original attributes RFC which puts attributes before `local`.

> Multiple attributes are supported inline, so that a function with multiple attributes does not have several lines of boilerplate:
```lua
@attribute1 @attribute2 @attribute3 @attribute4 @attribute5
local function example()
end
```

docs/syntax-attributes-functions-parameters.md
Vighnesh-V: I think `@[` being inseparable is perfectly fine.

alexmccord: Yeah, just wanted to make this observation explicit and have it be included in the RFC too.

aatxe: My intention with the proposal in internal discussions was definitely that it'd be inseparable.

aviralg: Yes, `@[` is one token. Thinking of it as two separate tokens didn't even occur to me.

docs/syntax-attributes-functions-parameters.md
aviralg: Already supported. Will document.

docs/syntax-attributes-functions-parameters.md
aviralg: `()` for attribute parameters is not supported. You can have entries without names in the parameter table. Parameter table is just like a normal Luau table with restrictions on entry value types.

docs/syntax-attributes-functions-parameters.md
aviralg: Its not really an "Alternative" in the sense that its not an alternative to the attribute feature, but rather a discussion of why an obvious syntax was not adopted. The presentation has been changed to avoid this confusion.

docs/syntax-attributes-functions-parameters.md
aviralg: Fixed. Provided an ebnf syntax for declaration files.

docs/syntax-attributes-functions-parameters.md
aviralg: Addressed this and updated the `ebnf` syntax description accordingly.

docs/syntax-attributes-functions-parameters.md
aviralg: As discussed privately, `unroll(5)` is not legal. However `unroll 5` is. The other two syntax are legal. The `ebnf` has been updated accordingly.

docs/syntax-attributes-functions-parameters.md
aviralg: The `ebnf` syntax has been reworked entirely and this issue is fixed.

docs/syntax-attributes-functions-parameters.md
aviralg: The `ebnf` syntax has been reworked entirely and this issue is fixed. I adapted the table grammar from Lua's documentation.

aviralg: Documented now.

andyfriesen (approved):

alexmccord (approved):

PR #31, opened 2024-04-17
RFC: Native Attribute for Functions

aviralg:
[Rendered](https://github.com/aviralg/rfcs/blob/attribute-native/docs/syntax-attribute-functions-native.md)

### DIFF
--- a/docs/syntax-attribute-functions-native.md
+++ b/docs/syntax-attribute-functions-native.md
@@ -2,11 +2,11 @@
 
 ## Summary
 
-This RFC proposes a function-level `@native` attribute to request native compilation for individual functions, independent of the script-level `--!native` hotcomment.
+This RFC proposes a function-level `@native` attribute to request native compilation for individual functions, independent of the script-level `--!native` comment directive.
 
 ## Motivation
 
-Luau's native compiler currently compiles whole scripts annotated with `--!native` hotcomment. However, this provides very coarse-grained control. Since all functions in the script may not benefit from native compilation, developers might be forced to move unrelated functions together to natively compiled scripts. In this RFC, we propose a function-level `@native` attribute to facilitate developers to pick and choose individual functions for native compilation.
+Luau's native compiler currently compiles whole scripts annotated with `--!native` comment directive. However, this provides very coarse-grained control. Since all functions in the script may not benefit from native compilation, developers might be forced to move unrelated functions together to natively compiled scripts. In this RFC, we propose a function-level `@native` attribute to facilitate developers to pick and choose individual functions for native compilation.
 
 ## Design
 
@@ -40,7 +40,7 @@ The implementation _may_ choose to issue warning in the following cases where `@
 
 1. A function has more than one occurrence of `@native` attribute
 2. An inner function has one or more occurrences of `@native` attribute when an _ancestor_ function already has a `@native` attribute.
-3. A function has a `@native` attribute when the script is annotated with `--!native` hotcomment.
+3. A function has a `@native` attribute when the script is annotated with `--!native` comment directive.
 
 ## Drawbacks
 
@@ -51,4 +51,4 @@ Introducing this attribute will have two adverse consequences:
 
 ## Alternatives
 
-The alternative would be to not provide this attribute and rely on `--!native` hotcomment to make compilation decisions on a per-script basis. This might force developers to break their code organization and move unrelated functions together but it does not prevent them from getting performance benefits.
+The alternative would be to not provide this attribute and rely on `--!native` comment directive to make compilation decisions on a per-script basis. This might force developers to break their code organization and move unrelated functions together but it does not prevent them from getting performance benefits.
--- a/docs/syntax-attribute-functions-native.md
+++ b/docs/syntax-attribute-functions-native.md
@@ -6,13 +6,13 @@ This RFC proposes a function-level `@native` attribute to request native compila
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


docs/syntax-attribute-functions-native.md
```diff
@@ -0,0 +1,54 @@
+# Native Attribute for Functions
+
+## Summary
+
+This RFC proposes a function-level `@native` attribute to request native compilation for individual functions, independent of the script-level `--!native` hotcomment.
```
Dekkonot, line 5: I was under the impression that these were called directives, not hotcomments. They're referred to as such on Luau's website at least: https://luau-lang.org/lint#commentdirective-26

aviralg: Changed all occurrences to "comment directive".

aatxe: We do informally call them hot comments often, but yeah, directives is better as official terminology.

alexmccord: Pick any one of these at random: 1) hot comments, 2) directives, 3) pragmas.

docs/syntax-attribute-functions-native.md
```diff
@@ -0,0 +1,54 @@
+# Native Attribute for Functions
+
+## Summary
+
+This RFC proposes a function-level `@native` attribute to request native compilation for individual functions, independent of the script-level `--!native` comment directive.
+
+## Motivation
+
+Luau's native compiler currently compiles whole scripts annotated with `--!native` comment directive. However, this provides very coarse-grained control. Since all functions in the script may not benefit from native compilation, developers might be forced to move unrelated functions together to natively compiled scripts. In this RFC, we propose a function-level `@native` attribute to facilitate developers to pick and choose individual functions for native compilation.
```
alexmccord, line 9: I'm not actually seeing a direct benefit for higher granularity of native codegen. This just says "either the module is native or it isn't" as opposed to benefits for having certain set of functions running native code and others aren't.

aviralg: The part that is being left unsaid is that we have an upper limit to how much code we can compile (the details are not yet final).  Compiling the whole module, including functions that don't benefit from native compilation, unnecessarily exhausts this limit. Function-level granularity can help address this problem.

aviralg: Eventually,  our compiler should automatically decide which functions to compile natively. This might make the `--!native` directive redundant, but the `@native` attribute may still remain useful for creators to explicitly ask for native compilation of specific functions.

docs/syntax-attribute-functions-native.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Design
+
+Syntactically, the `@native` attribute takes no parameters. It can be used on both top-level and inner functions. It applies recursively to all the functions defined within the lexical scope of the attributed function because the "inner" functions logically constitute the implementation of the attributed function. Hence, attributing "inner" functions as `@native` will not have any effect if an _ancestor_ function already has a `@native` attribute. 
+
+In the following example, both `parent` and `child` will be natively compiled. 
+
+```lua
+@native
```
alexmccord, line 18: Maybe this is actually expected, but for me personally, this feels like an ick. I don't know if I like the idea that `@native` is applied recursively.

docs/syntax-attribute-functions-native.md
alexmccord: That part that's being left unsaid should probably be added into the motivation section. That's what it's for! You can leave _specific_ details out, but just give an idea.

docs/syntax-attribute-functions-native.md
aviralg: Reversed this behavior.

AsynchronousAI:
```lua
function myFunc() --!native
```
or 
```lua
function myFunc() --@native
```

***
an `@` might make more sense because it means 'native at/@ `myFunc`'

aviralg:
> ```lua
> function myFunc() --!native
> ```
> 
> or
> 
> ```lua
> function myFunc() --@native
> ```
> 
> an `@` might make more sense because it means 'native at/@ `myFunc`'

I don't follow. Neither `function myFunc() --!native`, nor `function myFunc() --@native` is valid. `--!native` is a comment directive added once to the top of the file, not next to function definitions.

alexmccord:
> ```lua
> function myFunc() --!native
> ```
> 
> or
> 
> ```lua
> function myFunc() --@native
> ```
> 
> an `@` might make more sense because it means 'native at/@ `myFunc`'

It wasn't in the RFC originally, but it's explained here now. https://github.com/luau-lang/rfcs/commit/1ca8399b0d1380f0aa0f7a97e2296dde7e0832e4

docs/syntax-attribute-functions-native.md
```diff
@@ -0,0 +1,54 @@
...
+       -- do something
+    end
+    -- do something
+end
+```
+
+The implementation _may_ choose to issue warning in the following cases where `@native` attribute is redundant:
+
+1. A function has more than one occurrence of `@native` attribute
+2. A function has a `@native` attribute when the script is annotated with `--!native` comment directive.
```
vegorov-rbx, line 43: Let's remove this section.

'1' is wrong as Luau parser errors on duplicate attributes.
'2' is kind of unnecessary, as implementations already can issue any lint warnings they want (speaking about existing open-source projects and for built-in linter this is too vague)

aviralg: Removed.

vegorov-rbx (approved):

PR #32, opened 2024-04-17
RFC: 24-bit integer support for buffers

Dekkonot:
[Rendered document](https://github.com/Dekkonot/luau-rfcs/blob/buffer-24-bit-rfc/docs/function-buffer-24-bit.md)

I've seen a bunch of people on Roblox's developer forum and in the Roblox Open Source discord server desire 24-bit support for buffers. It seems reasonable to add, so here's an RFC for it.

### DIFF
--- a/docs/function-buffer-24-bit.md
+++ b/docs/function-buffer-24-bit.md
@@ -9,13 +9,13 @@ The built-in `buffer` library will gain 4 new functions:
 - `buffer.writei24`
 - `buffer.readi24`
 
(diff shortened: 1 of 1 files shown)


vegorov-rbx:
Buffer tries to match what is available directly on the CPU and us such, we are not interested in adding additional bit widths for every use case.
We also don't want to unnecessarily expend on the budget of individual functions we have to support multiple ways (lib, fastcall, ir).

PR #34, opened 2024-05-13
RFC: Vector library

ffrostfall:
[Rendered](https://github.com/ffrostflame/rfcs/blob/master/docs/vector-library.md)

### DIFF
--- a/docs/vector-library.md
+++ b/docs/vector-library.md
@@ -1,0 +1,62 @@
+# Vector library
+
+## Summary
+
+Implement a standard library which provides functionality for the vector type.
+
+## Motivation
+
+Currently, vectors are a primitive type implemented internally. All of the heavy work to implement vectors is done, but there is no runtime-agnostic way to use vectors in Luau code. This creates a struggle to use native vectors; currently the most popular Luau runtime [Lune](https://github.com/lune-org/lune) doesn't use native vectors. This results in performance drawbacks & difficulty utilizing vectors in cross-runtime code.
+
+## Design
+
+Implement a standard library for creating & using the existing vector type.
+
+### Library functions
+
+It's important to keep in mind that this list of implementable functions isn't intended to be exhaustive, but rather to serve as a starting point.
+
+`vector.create(x: number?, y: number?, z: number?)`
+Creates a vector with 3 components: x, y, z. If the feature flag for wide vectors is enabled, a fourth argument `w: number?` will be introduced. As per standard, vectors wouldn't have a metatable by default. A vector's metatable would need to be set by the C API `lua_setmetatable`.
+
+`vector.magnitude(vecA: vector): vector`
+Calculates the magnitude of a given vector.
+
+`vector.unit(vec: vector): vector`
+Returns the unit vector (aka normalized vector) of a given vector.
+
+`vector.cross(vecA: vector, vecB: vector): vector`
+Returns the cross product of two vectors.
+
+`vector.dot(vecA: vector): vector`
+Returns the dot product of a vector.
+
+### Arithmetic operations
+
+Primitive operators for vectors are already implemented, so this RFC doesn't concern vector arithmetic.
+
+### Native codegen
+
+In the future, vectors will have special treatment in native codegen. This creates an expectation that the vector library will also have special treatment in codegen, but it isn't clear what that will look like, or if the performance benefits make sense.
+
+### Compiler options
+
+Currently, there are 2 compiler options relating to vectors, `vectorLib` & `vectorCtor`. This poses an interesting problem: a builtin library would remove the _requirement_ for such compiler options, however these options still need to be supported. The intuitive solution to this is to leave both compiler options working and maintained, but provide the built-in vector library by default, allowing two vector libraries and two vector constructors.
+
+## Drawbacks
+
+As per all additional globals, this creates a new global `vector`. This is a commonly used variable name, and many style guides will advise to avoid using `vector` as a variable name due to the global being added.
+
+Introducing a vector library means introducing more globals. Due to the prescence of vectors in performance-intensive code, it can be assumed that the built-in functions will have fastcalls. This means more fastcall slots will be used.
+
+Existing code won't get any improvements from a built-in library. Developers will have to switch over to the built-in library in order to gain any benefit.
+
+## Alternatives
+
+Do nothing; vectors have an internal constructor, which lets the runtime implement vectors, and subsequently a library for vector math.
+
+### Alternative implementations
+
+Due to the extremely common and heavy usage of vectors, an alternative to `vector.create` could be something more concise, like `vector.new`. This does break the precedent set by `buffer.create`, `table.create`, and `coroutine.create` though.
+
+Instead of `vector.magnitude`, the magnitude could be derived from the vector's coordinates themselves, and be accessed by a property instead of a function. There is a downside to this: all current properties of vectors are hard-coded to the VM, so any new property to vectors requires a lot of additional complexity & changes to the VM to allow for this. A library function, however, would be trivial. An easy and quick workaround to the verbosity would be at the runtime/C API level. It's trivial to set the metatable of vectors to be the vector library: this allows for `vec:magnitude()` without much issue.
--- a/docs/vector-library.md
+++ b/docs/vector-library.md
@@ -30,7 +30,7 @@ Calculates the magnitude of a given vector.
 
 `vector.normalized(vec: vector): vector`
 
-Returns the normalized version (aka unit vector) of a given vector. If a zero vector is passed, return `0`. `vector(0, 0, 0):normalized()` should return zero.
+Returns the normalized version (aka unit vector) of a given vector. If a zero vector is passed, return zero. For further clarification, `vector.normalized(vector(0, 0, 0))` should return zero.
 
 `vector.cross(vecA: vector, vecB: vector): vector`
 
--- a/docs/vector-library.md
+++ b/docs/vector-library.md
@@ -18,7 +18,7 @@ This RFC proposes the following basic functions & constants as a starting point,
 
 ---
 
-`vector(x: number?, y: number?, z: number?)`
+`vector(x: number, y: number, z: number)`
 
 Creates a vector with 3 components: x, y, z. If the feature flag for wide vectors is enabled, a fourth argument `w: number?` will be introduced.
 
@@ -36,7 +36,7 @@ Returns the normalized version (aka unit vector) of a given vector. If a zero ve
 
 Returns the cross product of two vectors. If 4-wide vectors are enabled, this function will ignore the fourth component, and return the 3-dimensional cross product.
 
-`vector.dot(vecA: vector, vecB: vector): vector`
+`vector.dot(vecA: vector, vecB: vector): number`
 
 Returns the dot product of two vectors.
 
@@ -48,9 +48,9 @@ Equivalent of `math.floor` for vectors.
 
 Equivalent of `math.ceil` for vectors.
 
-`vector.angle(vecA: vector, vecB: vector): vector`
+`vector.angle(vecA: vector, vecB: vector): number`
 
-Returns the angle between two vectors.
+Returns the angle between two vectors in radians.
 
 `vector.max(input: vector, maximum: vector): vector`
 
--- a/docs/vector-library.md
+++ b/docs/vector-library.md
@@ -92,6 +92,13 @@ Vector where `x=1, y=1, z=1, w?=1`.
 
 Primitive operators for vectors are already implemented, so this RFC doesn't concern vector arithmetic.
 
+### Component access
+
+Component access is already supported in the VM using the fields 'x', 'y', 'z', 'X', 'Y' and 'Z' (plus 'w' and 'W' in 4-wide mode).
+Note that writes to a single component are not supported as the vector value is immutable.
+This RFC doesn't define any changes to this behavior.
+
+
 ### Compiler options
 
 Currently, there are 2 compiler options relating to vectors, `vectorLib` & `vectorCtor`. This poses an interesting problem: a builtin library would remove the _requirement_ for such compiler options, however these options still need to be supported. The intuitive solution to this is to leave both compiler options working and maintained, but provide the built-in vector library by default, allowing two vector libraries and two vector constructors.
--- a/docs/vector-library.md
+++ b/docs/vector-library.md
@@ -54,33 +54,33 @@ Returns the angle between two vectors in radians.
 
 `vector.abs(vec: vector): vector`
 
-Applies math.abs component-wise for each vector. Equivalent of: `vector(math.abs(vec.x), ect)`
+Applies math.abs component-wise for each vector. Equivalent of: `vector(math.abs(vec.x), ect)`.
 
 `vector.sign(vec: vector): vector`
 
-Applies math.sign component-wise for each vector. Equivalent of: `vector(math.sign(vec.x), ect)`
+Applies `math.sign` component-wise for each vector. Equivalent of: `vector(math.sign(vec.x), ect)`.
 
 `vector.clamp(vec: vector, min: vector, max: vector): vector`
 
-Applies math.clamp component-wise for each vector. Equivalent of: `vector(math.clamp(vec.x, min.x, max.x), ect)`
+Applies `math.clamp` component-wise for each vector. Equivalent of: `vector(math.clamp(vec.x, min.x, max.x), ect)`.
 
-`vector.max(list: ...vector): vector`
+`vector.max(...: vector): vector`
 
-Finds the highest vector xyz in the list of vectors provided and returns a new vector with its xyz being those highest values
+Applies `math.max` component-wise for each vector. Equivalent of: `vector(math.max((...).x), ect)`.
 
-`vector.min(list: ...vector): vector`
+`vector.min(...: vector): vector`
 
-Inverse of `vector.min` where instead of looking for the highest it looks for the lowest
+Applies `math.min` component-wise for each vector. Equivalent of: `vector(math.max((...).x), ect)`.
 
 ---
 
 `vector.zero`
 
-Vector where `x=0, y=0, z=0, w?=0`
+Vector where `x=0, y=0, z=0, w?=0`.
 
 `vector.one`
 
-Vector where `x=1, y=1, z=1, w?=1`
+Vector where `x=1, y=1, z=1, w?=1`.
 
 ---
 
--- a/docs/vector-library.md
+++ b/docs/vector-library.md
@@ -24,7 +24,7 @@ Creates a vector with 3 components: x, y, z. If the feature flag for wide vector
 
 Due to the common usage of vectors, vector creation should be ergonomic. Therefore, it is probably worth breaking the `create()` naming standard.
 
-`vector.magnitude(vecA: vector): vector`
+`vector.magnitude(vecA: vector): number`
 
 Calculates the magnitude of a given vector.
 
@@ -36,9 +36,9 @@ Returns the unit vector (aka normalized vector) of a given vector.
 
 Returns the cross product of two vectors.
 
-`vector.dot(vecA: vector): vector`
+`vector.dot(vecA: vector, vecB: vector): vector`
 
-Returns the dot product of a vector.
+Returns the dot product of two vectors.
 
 ---
 
@@ -69,5 +69,7 @@ Do nothing; vectors have an internal constructor, which lets the runtime impleme
 ### Alternative implementations
 
 A more standard alternative to `vector(...)` could be something like `vector.create` or `vector.new`. This follows the standard set by `buffer.create`, `coroutine.create`, and `table.create`, and doesn't involve calling a table.
+
+Another alternative to vector creation is special syntax for creating buffers, such as `<x, y, z>`. This would require a lot more complexity & discussion, however it would fall more in line with the other primitive types.
 
 Instead of `vector.magnitude`, the magnitude could be derived from the vector's coordinates themselves, and be accessed by a property instead of a function. There is a downside to this: all current properties of vectors are hard-coded to the VM, so any new property to vectors requires a lot of additional complexity & changes to the VM to allow for this. A library function, however, would be trivial. An easy and quick workaround to the verbosity would be at the runtime/C API level. It's trivial to set the metatable of vectors to be the vector library: this allows for `vec:magnitude()` without much issue.
--- a/docs/vector-library.md
+++ b/docs/vector-library.md
@@ -111,7 +111,7 @@ A new primitive `vector` type should be added to the type system.
 
 A better vector constructor with new syntax, such as `<x, y, z>` or `(x, y, z)` or `|x, y, z|` or `vector(x, y, z)` or `[x, y, z]` could be implemented. This would make code that does stuff with vectors much less verbose.
 
-The buffer library could include functions to work with vectors, such as `buffer.writevector2`, `buffer.writevector3`, and in 4-wide mode, `buffer.writevector4`. The corrosponding read functions would of course be included.
+The buffer library could include functions to work with vectors, such as `buffer.writevector2`, `buffer.writevector3`, and in 4-wide mode, `buffer.writevector4`. The corresponding read functions would of course be included.
 
 ## Drawbacks
 
--- a/docs/vector-library.md
+++ b/docs/vector-library.md
@@ -54,11 +54,11 @@ Returns the angle between two vectors.
 
 `vector.max(input: vector, maximum: vector): vector`
 
-Equivalent of `math.max` for vectors.
+Equivalent of: `vector(math.max(a.x, b.x), etc)`
 
 `vector.min(input: vector, minimum: vector): vector`
 
-Equivalent of `math.min` for vectors.
+Equivalent of: `vector(math.min(a.x, b.x), etc)`
 
 ---
 
--- a/docs/vector-library.md
+++ b/docs/vector-library.md
@@ -66,7 +66,7 @@ Applies `math.sign` component-wise for each vector. Equivalent of: `vector.creat
 
 `vector.clamp(vec: vector, min: vector, max: vector): vector`
 
-Applies `math.clamp` component-wise for each vector. Equivalent of: `vector(math.clamp(vec.x, min.x, max.x), etc)`.
+Applies `math.clamp` component-wise for each vector. Equivalent of: `vector.create(math.clamp(vec.x, min.x, max.x), etc)`.
 
 `vector.max(...: vector): vector`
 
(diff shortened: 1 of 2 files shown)


kalrnlo:
I think the min, max functions should accept multiple vectors as the current math.max and min allow multiple numbers, plus roblox's vector3s min and max methods also accept multiple vectors.

There should also be abs and clamp functions, as those would be useful for vectors. Plus id assume it would be intended to keep the vector library in feature parity somewhat with the math lib.

docs/vector-library.md
```diff
@@ -0,0 +1,111 @@
...
+
+The default metatable for vectors should now be the `vector` library. While buffers and coroutines do not have this, vectors likely should for ergonomics.
+
+### Library functions & constants
+
+This RFC proposes the following basic functions & constants as a starting point, but as with all builtin libraries, future RFCs can propose additional functions.
+
+---
+
+`vector(x: number?, y: number?, z: number?)`
```
vegorov-rbx, line 21:
```suggestion
`vector(x: number, y: number, z: number)`
```

It's best to encourage practice of hitting the fastest constructor we have available.

It will also avoid confusion of what vector(1) is.
While `Vector3.new(1)` is [1, 0, 0] on Roblox, for people coming from GLSL where `vec3(1)` is [1, 1, 1], it might be unintuitive.
Even if supported by runtime, the typing can be more strictly defined.

docs/vector-library.md
```diff
@@ -0,0 +1,111 @@
...
+
+Equivalent of `math.floor` for vectors.
+
+`vector.ceil(vec: vector): vector`
+
+Equivalent of `math.ceil` for vectors.
+
+`vector.angle(vecA: vector, vecB: vector): vector`
+
+Returns the angle between two vectors.
```
vegorov-rbx, line 53: In what units?

docs/vector-library.md
```diff
@@ -0,0 +1,111 @@
...
+
+`vector.normalized(vec: vector): vector`
+
+Returns the normalized version (aka unit vector) of a given vector. If a zero vector is passed, return zero. For further clarification, `vector.normalized(vector(0, 0, 0))` should return zero.
+
+`vector.cross(vecA: vector, vecB: vector): vector`
+
+Returns the cross product of two vectors. If 4-wide vectors are enabled, this function will ignore the fourth component, and return the 3-dimensional cross product.
+
+`vector.dot(vecA: vector, vecB: vector): vector`
```
vegorov-rbx, line 39: dot product results in a number.

bradsharp:
Would we include this in the Roblox engine?

alexmccord:
> Would we include this in the Roblox engine?

In the RFC there's the wording "but there is no runtime-agnostic way to use vectors in Luau code," so the answer is yes.

vegorov-rbx:
> Would we include this in the Roblox engine?

Yes and that's the main problem to solve for the RFC.

jackdotink:
I think that this should be included in the engine, I don't think there is a blocker and it does allow for more runtime agnostic code. Vectors are a handy way to do many things. As an example when I writing a parser recently I wanted to use vectors for spans, but I could not because I had no way to create them.

ffrostfall:
> > Would we include this in the Roblox engine?
> 
> Yes and that's the main problem to solve for the RFC.

What about that is a blocker for this RFC?

docs/vector-library.md
```diff
@@ -0,0 +1,123 @@
...
+
+---
+
+`vector(x: number, y: number, z: number)`
+
+Creates a vector with 3 components: x, y, z. If the feature flag for wide vectors is enabled, a fourth argument `w: number?` will be introduced.
+
+Due to the common usage of vectors, vector creation should be ergonomic. Therefore, it is probably worth breaking the `create()` naming standard.
+
+`vector.magnitude(vecA: vector): number`
```
vegorov-rbx, line 27:
```suggestion
`vector.magnitude(vec: vector): number`
```

docs/vector-library.md
```diff
@@ -0,0 +1,123 @@
...
+
+Calculates the magnitude of a given vector.
+
+`vector.normalized(vec: vector): vector`
+
+Returns the normalized version (aka unit vector) of a given vector. If a zero vector is passed, return zero. For further clarification, `vector.normalized(vector(0, 0, 0))` should return zero.
```
vegorov-rbx, line 33: For compatibility with Roblox runtime and for code size and performance (including branchless native codegen lowering) I suggest removing requirements for zero length vector detection.

docs/vector-library.md
```diff
@@ -0,0 +1,123 @@
...
+
+`vector.cross(vecA: vector, vecB: vector): vector`
+
+Returns the cross product of two vectors. If 4-wide vectors are enabled, this function will ignore the fourth component, and return the 3-dimensional cross product.
+
+`vector.dot(vecA: vector, vecB: vector): number`
+
+Returns the dot product of two vectors.
```
vegorov-rbx, line 41: Can we note how it behaves for 4-wide vectors?

docs/vector-library.md
```diff
@@ -0,0 +1,123 @@
...
+
+---
+
+### Arithmetic operations
+
+Primitive operators for vectors are already implemented, so this RFC doesn't concern vector arithmetic.
+
+### Native codegen
+
+In the future, vectors will have special treatment in native codegen. This creates an expectation that the vector library will also have special treatment in codegen, but it isn't clear what that will look like, or if the performance benefits make sense.
```
vegorov-rbx, line 93: This can be removed as vectors already have special treatment in codegen.

Generally, library proposals shouldn't have to think about fastcalls/lowering hooks as those are implementation details which are implemented without a proposal being required.
Whether or not existing custom lowering hooks from Roblox will be included (or written from scratch) doesn't have to be specified.

docs/vector-library.md
```diff
@@ -0,0 +1,123 @@
...
+
+Existing code won't get any improvements from a built-in library. Developers will have to switch over to the built-in library in order to gain any benefit.
+
+## Alternatives
+
+Do nothing; vectors have an internal constructor, which lets the runtime implement vectors, and subsequently a library for vector math.
+
+### Alternative implementations
+
+A more standard alternative to `vector(...)` could be something like `vector.create` or `vector.new`. This follows the standard set by `buffer.create`, `coroutine.create`, and `table.create`, and doesn't involve calling a table, which requires a metatable with `__call` to be set on the library. This breaks the standard previously set by Lua, as there is currently no library with a metatable set by default.
```
vegorov-rbx, line 117: string library sets the metatable of string values by default, so there is precedent (`luaopen_string`).

vegorov-rbx:
Any considerations around interactions with the `buffer` library?
It's not even mentioned as an avenue for future work (might resolve this discussion https://github.com/luau-lang/luau/discussions/1296)

docs/vector-library.md
jackdotink: I'm unsure of the behavior it should have - I would simply use only the first three components even when compiled in four component mode. The other options are to error or compute the dot product of vectors with four components, which a quick google returns little results for.

jackdotink:
I think adding `buffer.writevector` and `buffer.readvector` could prove to be valuable, even though they are already possible currently. I'll add it the RFC.

docs/vector-library.md
ffrostfall: That makes sense.

docs/vector-library.md
ffrostfall: To clarify, I mean there is no library which has an implicit `__call` metamethod on it. There's no luau library that lets you do, for example, `string()`.

docs/vector-library.md
ffrostfall: Sounds good

docs/vector-library.md
```diff
@@ -0,0 +1,123 @@
...
+
+`vector.floor(vec: vector): vector`
+
+Equivalent of `math.floor` for vectors.
+
+`vector.ceil(vec: vector): vector`
+
+Equivalent of `math.ceil` for vectors.
+
+`vector.angle(vecA: vector, vecB: vector): number`
```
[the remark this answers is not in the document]
ffrostfall: No.

docs/vector-library.md
alicesaidhi: I don't think having it only work on the first 3 components is correct.
If we were to follow the definition of the dot product between 2 vectors
$$a \cdot b = \sum_{i=1}^n a_i b_i$$
we can say that the dot product for four component vectors would be:
$$\vec{a} \cdot \vec{b} = a_1 b_1 + a_2 b_2 + a_3 b_3 + a_4 b_4 $$

MagmaBurnsV:
Coming back to this, it would be very useful to have.

<blockquote>
Due to the common usage of vectors, vector creation should be ergonomic. Therefore, it is probably worth breaking the create() naming standard.
</blockquote>

Why? This would only serve to make the library inconsistent and confusing for new users. Considering most will be introduced to it through Roblox's editor autocomplete, it won't be obvious there how it should be used.

***

How would you be able to read individual components from the vector? Indexing might be inconsistent, so there should be some `vector.readx` functions and etc.

Of course, `vector.readw` would return an optional number. There could also be some `vector.unpack` function for deconstructing a vector into a tuple.

***

How should constructing with 4 arguments work with relation to `LUA_VECTOR_SIZE` being implementation-specific? In my mind it should always allow the fourth argument, even when it isn't stored. That way there won't be any need for messy error-handling or nil-checking on the programmer's end.

Then it can provide a `vector.size` constant so users can determine how they can use `w`.

kalrnlo:
> How should constructing with 4 arguments work with relation to `LUA_VECTOR_SIZE` being implementation-specific? In my mind it should always allow the fourth argument, even when it isn't stored. That way there won't be any need for messy error-handling or nil-checking on the programmer's end.
> 
> Then it can provide a `vector.size` constant so users can determine how they can use `w`.

Theres no way to change the size of a vector during runtime so it cant always allow the forth argument.

vegorov-rbx:
> Theres no way to change the size of a vector within luau, and I'd assume it would come with a degradation of performance to vectors if it were added (@vegorov-rbx correct me if I'm wrong on this).

I believe that was being suggested is a method to detect the kind of current environment, not change it.
Similar to conformance tests for vectors using:
`local vector_size = if pcall(function() return vector(0, 0, 0).w end) then 4 else 3`


> How would you be able to read individual components from the vector? Indexing might be inconsistent, so there should be some `vector.readx` functions and etc.
> 
> Of course, `vector.readw` would return an optional number.

Property access is already implemented in the VM, so there will be no need for additional methods and that's why proposal doesn't include it.
FYI, components are accessed as X/Y/Z/W/x/y/z/w.
When `LUA_VECTOR_SIZE` is 3, there is no W/w and it's an error to access them.
Your suggestions will not be accepted by the Luau team.

MagmaBurnsV:
<blockquote>
How? it'd be documented plus vector() is much easier to type compared to vector.create especially when with the case of vectors you're usually creating a lot of them at a time. Also what do you mean by autocomplete posing an issue?
</blockquote>

I see no reason why we should break a reasonable convention kept with every other native library (`table.create`, `coroutine.create` and `buffer.create`) just to save 7 characters. Having it as a `__call` metamethod might also impact performance, but what do I know.

The studio LSP concern was based on learning and exposure, but I see now that it is out of the scope here.

<blockquote>
Property access is already implemented in the VM, so there will be no need for additional methods and that's why proposal doesn't include it.
</blockquote>

<blockquote>
Whats the usecase for `vector.unpack`?
</blockquote>

I did not know vectors had property access already, so this suggestion was intended to avoid having to do multiple read function calls.

<blockquote>
When LUA_VECTOR_SIZE is 3, there is no W/w and it's an error to access them.
Your suggestions will not be accepted by the Luau team.
</blockquote>

Apologies if I came off as abrasive, but my concern which I should've stated was how this would fare with the typechecker. If a script uses 4-width vectors when possible and then otherwise 3, how would it be determined if this was allowed?

```lua
local Vector = vector.zero
local W = nil

if vector.size == 4 then
	-- (x: number, y: number, z: number, w: number) -> vector
	Vector = vector.create(1, 2, 3, 4)
	W = Vector.w
else
	-- (x: number, y: number, z: number) -> vector
	Vector = vector.create(1, 2, 3)
end
```

vegorov-rbx:
> If a script uses 4-width vectors when possible and then otherwise 3, how would it be determined if this was allowed?

I don't think this will be determined by the current typechecker.
And I would say there is no goal to solve the problem. You configure your Luau build use 3 or 4, typechecker might load up a different definition based on the setting.
Any casts can be used as usual in cases where typechecker is not powerful enough.
I would actually not even include `size` property or function for this reason as it implies support beyond what we want to include for our 4 wide support, which is a non-standard configuration we are treating in a 'best effort' manner (native codegen component support is completely missing for example).

docs/vector-library.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Buffer Library
+
+`buffer.writevector(b: buffer, offset: number, vec: vector)`
+
+Writes the vector into the buffer at the offset. Each component will be written as a f32. In four component mode, it will write all four components and write 16 bytes. In three component mode, it will write 12 bytes.
+
+`buffer.readvector(b: buffer, offset: number): vector`
+
+Reads a vector from the buffer at the offset. In four component mode it will read 16 bytes and in three component mode it will read 12 bytes.
```
vegorov-rbx, line 95: This might not be the path for vector storage in buffers.
I would suggest taking a look at the discussion we had on this topic: https://github.com/luau-lang/luau/discussions/1296

I think a good path forward would be `readvector2`, `readvector3` and `readvector4` with similar methods to write.
I don't think f32xN is a good choice because it implies multiple value read/write unrelated to the vector type.

I do like the idea of having 2-component access as an option.

docs/vector-library.md
vegorov-rbx: That's not what the RFC text says. It still says "there is currently no library with a metatable set by default" and that's not true. If it said that there's no library with a `__call` metamethod set on a type, that would be true.

bradsharp:
The motivation for supporting `vector()` over `vector.create()` is highly dependent on the type of code being written and I do not find it a compelling enough reason to break convention. I don't think we can move forward with this RFC without changing that.

As an aside, I'm open to us introducing these shorthands but I think that should be a separate RFC for all types rather than as a one-off for this case.

aatxe:
There is a problem in that the libraries are already inconsistent about what they name constructors though. I don't think it's particularly bad to pick a more ergonomic one for `vector`, in light of that fact.

alexmccord:
How about `[1, 2, 3]`. If we want a convenient syntax for vectors, we might as well add syntax to the language. I hate `vector.create()` suggestion with a burning passion and would like to never see that suggestion again. For 4 component wide vectors, the syntax would be `[1, 2, 3, 4]`.

It does conflict with `t[[1, 2, 3]]` but I don't think I care.

alexmccord:
Changing the suggestion to `(1, 2, 3)`. It doesn't conflict with `t[[1, 2, 3]]` which is nice, and still follows the standard vector notation.

memorycode:
I think `vector(x, y, z)` and `<x, y, z>` are aligned to the goals and that the other versions are just those but with more downsides:

- If we assume new syntax like `<x, y, z>` would be confusing to new people and readability, then `(x, y, z)` would be at least that complex or more
- `(x, y, z)` is common tuple syntax, but more importantly I think it's undesirable that it's so close to `f(x, y, z)`
- `[x, y, z]` conflicts with `t[[1, 2, 3]]`
- `vector.create(x, y, z)` and `vector.new(x, y, z)` are more verbose than `vector(x, y, z)` and ergonomics is a big motivator here

To me, `<x, y, z>` also more strongly implies that vectors are special in Luau.

docs/vector-library.md
jackdotink: resolved

docs/vector-library.md
jackdotink: resolved

docs/vector-library.md
jackdotink: functions removed and your idea has been placed into future work

green-real:
I think we should leave any shorthand vector constructors to a different RFC and add a default constructor in vector library which can always be used for vector creation. In that case we'd have to discuss if we want to use `vector.create(x, y, z)` or `vector.new(x, y, z)` for now.

docs/vector-library.md
vegorov-rbx: I can create a buffer update later when I go over pending buffer RFCs as well.

docs/vector-library.md
```diff
@@ -0,0 +1,119 @@
...
+
+Implement a standard library that provides functionality for the vector type.
+
+## Motivation
+
+Currently, vectors are a primitive type implemented internally. All of the heavy work to implement vectors is done, but there is no runtime-agnostic way to use vectors in Luau code. Individual Luau runtimes, such as [Lune](https://github.com/lune-org/lune) or Roblox, must provide their own distinct libraries to work with native vectors which may be inconsistent. In cross-runtime code, this results in performance drawbacks & difficulty utilizing native vectors.
+
+## Design
+
+The default metatable for vectors should now be the `vector` library. While buffers and coroutines do not have this, vectors likely should for ergonomics.
```
vegorov-rbx, line 13:
```suggestion
The default metatable for vectors should now be the `vector` library. While buffers and coroutines do not have this, vectors likely should for ergonomics.

Luau VM can be configured in two different vector value modes.
Default configuration uses vectors with 3 components (`xyz`) and if the `LUA_VECTOR_SIZE` configuration option is set to 4, vector values get an additional forth `w` component.
This mode will be referred to as '4-wide mode' in this proposal.
```

docs/vector-library.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Library functions & constants
+
+This RFC proposes the following basic functions and constants.
+
+`vector.create(x: number, y: number, z: number): vector`
+
+Creates a new vector with the given components. When in 4-wide mode takes a fourth `w` argument. None of the arguments are optional.
```
vegorov-rbx, line 21: Constructor is something that already has a defined behavior in both compiler and runtime.
This change will not be backwards compatible for 4-wide vector mode.

For 4-wide vectors I propose a different description:

```suggestion
`vector.create(x: number, y: number, z: number): vector`

Creates a new vector with the given components. In 4-wide mode, the forth `w` component value is set to 0.0.

`vector.create(x: number, y: number, z: number, w: number): vector`

Additional constructor available in 4-wide mode. Creates a new vector with the given components.
```

I think this matches the compiler behavior where `vector.create(1, 2, 3, nil)` will not be recognised as a constant and the fast-call behavior where again, `vector.create(1, 2, 3, nil)` will exit the fast-call into a library call fallback. 
`vector.create(1, 2, 3)` is allowed in constant folding and fast-calls in 4-wide mode today.

While we do define the fallback library function in Conformance.test.cpp to treat explicit `nil` as 0.0, we can change the method to use the new library one, since it's just a unit test.

docs/vector-library.md
```diff
@@ -0,0 +1,119 @@
...
+
+Returns the angle between two vectors in radians. The axis, if specified, is used to determine the sign of the angle. If 4-wide vectors are enabled, this function will ignore the fourth component and return the 3-dimensional angle.
+
+`vector.floor(vec: vector): vector`
+
+Applies `math.floor` component-wise for each vector. Equivalent of: `vector(math.floor(vec.x), etc)`.
+
+`vector.ceil(vec: vector): vector`
+
+Applies `math.ceil` component-wise for each vector. Equivalent of: `vector(math.ceil(vec.x), etc)`.
```
vegorov-rbx, line 49:
```suggestion
Applies `math.ceil` component-wise for each vector. Equivalent of: `vector.create(math.ceil(vec.x), etc)`.
```

docs/vector-library.md
```diff
@@ -0,0 +1,119 @@
...
+
+`vector.abs(vec: vector): vector`
+
+Applies `math.abs` component-wise for each vector. Equivalent of: `vector(math.abs(vec.x), etc)`.
```
vegorov-rbx, line 53:
```suggestion
Applies `math.abs` component-wise for each vector. Equivalent of: `vector.create(math.abs(vec.x), etc)`.
```

docs/vector-library.md
```diff
@@ -0,0 +1,119 @@
...
+
+`vector.sign(vec: vector): vector`
+
+Applies `math.sign` component-wise for each vector. Equivalent of: `vector(math.sign(vec.x), etc)`.
```
vegorov-rbx, line 57:
```suggestion
Applies `math.sign` component-wise for each vector. Equivalent of: `vector.create(math.sign(vec.x), etc)`.
```

docs/vector-library.md
```diff
@@ -0,0 +1,119 @@
...
+
+`vector.clamp(vec: vector, min: vector, max: vector): vector`
+
+Applies `math.clamp` component-wise for each vector. Equivalent of: `vector(math.clamp(vec.x, min.x, max.x), etc)`.
```
vegorov-rbx, line 61:
```suggestion
Applies `math.clamp` component-wise for each vector. Equivalent of: `vector.create(math.clamp(vec.x, min.x, max.x), etc)`.
```

docs/vector-library.md
```diff
@@ -0,0 +1,119 @@
...
+
+`vector.max(...: vector): vector`
+
+Applies `math.max` component-wise for each vector. Equivalent of: `vector(math.max((...).x), etc)`.
```
vegorov-rbx, line 65:
```suggestion
Applies `math.max` component-wise for each vector. Equivalent of: `vector.create(math.max((...).x), etc)`.
```

docs/vector-library.md
```diff
@@ -0,0 +1,119 @@
...
+
+`vector.min(...: vector): vector`
+
+Applies `math.min` component-wise for each vector. Equivalent of: `vector(math.min((...).x), etc)`.
```
vegorov-rbx, line 69:
```suggestion
Applies `math.min` component-wise for each vector. Equivalent of: `vector.create(math.min((...).x), etc)`.
```

docs/vector-library.md
```diff
@@ -0,0 +1,127 @@
...
+
+`vector.one`
+
+Vector where `x=1, y=1, z=1, w?=1`.
+
+---
+
+### Arithmetic operations
+
+Primitive operators for vectors are already implemented, so this RFC doesn't concern vector arithmetic.
```
vegorov-rbx, line 93:
```suggestion
Primitive operators for vectors are already implemented, so this RFC doesn't concern vector arithmetic.

### Component access

Component access is already supported in the VM using the fields 'x', 'y', 'z', 'X', 'Y' and 'Z' (plus 'w' and 'W' in 4-wide mode).
Note that writes to a single component are not supported as the vector value is immutable.
This RFC doesn't define any changes to this behavior.

```

vegorov-rbx (approved):
I understand that not everyone is happy with the constructor we have chosen for consistency with libraries of other types, but we can look at the shorthand or special constructor syntax separately.

Note that `local vector = vector.create` or other name can always be used without an impact on constant-folding or fastcalls as part of the 'constant locals' optimization in the compiler.

docs/vector-library.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Typechecking
+
+A new primitive `vector` type should be added to the type system.
+
+## Future Work
+
+A better vector constructor with new syntax, such as `<x, y, z>` or `(x, y, z)` or `|x, y, z|` or `vector(x, y, z)` or `[x, y, z]` could be implemented. This would make code that does stuff with vectors much less verbose.
+
+The buffer library could include functions to work with vectors, such as `buffer.writevector2`, `buffer.writevector3`, and in 4-wide mode, `buffer.writevector4`. The corrosponding read functions would of course be included.
```
dcope-rbx, line 114:
```suggestion
The buffer library could include functions to work with vectors, such as `buffer.writevector2`, `buffer.writevector3`, and in 4-wide mode, `buffer.writevector4`. The corresponding read functions would of course be included.
```

docs/vector-library.md
```diff
@@ -0,0 +1,134 @@
...
+## Motivation
+
+Currently, vectors are a primitive type implemented internally. All of the heavy work to implement vectors is done, but there is no runtime-agnostic way to use vectors in Luau code. Individual Luau runtimes, such as [Lune](https://github.com/lune-org/lune) or Roblox, must provide their own distinct libraries to work with native vectors which may be inconsistent. In cross-runtime code, this results in performance drawbacks & difficulty utilizing native vectors.
+
+## Design
+
+The default metatable for vectors should now be the `vector` library. While buffers and coroutines do not have this, vectors likely should for ergonomics.
+
+Luau VM can be configured in two different vector value modes.
+Default configuration uses vectors with 3 components (`xyz`) and if the `LUA_VECTOR_SIZE` configuration option is set to 4, vector values get an additional forth `w` component.
```
dcope-rbx, line 16: Do we support other values for `LUA_VECTOR_SIZE` other than 3 and 4? If so, how will these other values be supported (or not)?

vegorov-rbx: We do not support any other configurations.

vrn-sn, line 16:
```suggestion
Default configuration uses vectors with 3 components (`xyz`) and if the `LUA_VECTOR_SIZE` configuration option is set to 4, vector values get an additional fourth `w` component.
```

docs/vector-library.md
```diff
@@ -0,0 +1,134 @@
...
+This mode will be referred to as '4-wide mode' in this proposal.
+
+### Library functions & constants
+
+This RFC proposes the following basic functions and constants.
+
+`vector.create(x: number, y: number, z: number): vector`
+
+Creates a new vector with the given components. In 4-wide mode, the forth `w` component value is set to 0.0.
```
vrn-sn, line 25:
```suggestion
Creates a new vector with the given components. In 4-wide mode, the fourth `w` component value is set to 0.0.
```

vrn-sn (commented):
Small typos

docs/vector-library.md
```diff
@@ -0,0 +1,134 @@
...
+
+A better vector constructor with new syntax, such as `<x, y, z>` or `(x, y, z)` or `|x, y, z|` or `vector(x, y, z)` or `[x, y, z]` could be implemented. This would make code that does stuff with vectors much less verbose.
+
+The buffer library could include functions to work with vectors, such as `buffer.writevector2`, `buffer.writevector3`, and in 4-wide mode, `buffer.writevector4`. The corresponding read functions would of course be included.
+
+## Drawbacks
+
+As per all additional globals, this creates a new global `vector`. This is a commonly used variable name, and many style guides will advise to avoid using `vector` as a variable name due to the global being added.
+
+Introducing a vector library means introducing more globals. Due to the prescence of vectors in performance-intensive code, it can be assumed that the built-in functions will have fastcalls. This means more fastcall slots will be used.
```
vegorov-rbx, line 120:
```suggestion
Introducing a vector library means introducing more globals. Due to the presence of vectors in performance-intensive code, it can be assumed that the built-in functions will have fastcalls. This means more fastcall slots will be used.
```

PR #35, opened 2024-05-18
RFC: Lambda Function Syntax

goldenstein64:
[Rendered](https://github.com/goldenstein64/luau-rfcs/blob/syntax-lambda-function/docs/syntax-lambda-function.md)

### DIFF
--- a/docs/syntax-lambda-function.md
+++ b/docs/syntax-lambda-function.md
@@ -153,7 +153,7 @@ This adds a special symbol `\` to the language and increases the language's comp
 
 ## Alternatives
 
-This syntax does not have to be implemented at all. It encourages users to write less error-prone code at best via functional programming, and it is another thing for a newcomer to learn how to read at worst. Some users may prefer writing their code in an inherently less performant style.
+This syntax does not have to be implemented at all. It encourages users to write less error-prone code at best via functional programming, and it is another thing for a newcomer to learn how to read at worst.
 
 There were a lot of other syntaxes considered before this, some examples and explanations why they were not chosen are listed below:
 
--- a/docs/syntax-lambda-function.md
+++ b/docs/syntax-lambda-function.md
@@ -149,7 +149,7 @@ print(add(10, 20)) --> 30
 This adds a special symbol `\` to the language and increases the language's complexity.
 - It is not the most accessible on all keyboards, e.g. QWERTZ and AZERTY, so it may be annoying for non-US users to write.
 - It may conflict with a potential set difference operator in type space, which is usually denoted as `T \ U` in mathematics. e.g. `\() nil :: T \ U \`
-- This syntax might appear alien to users. The only language I know that uses `\` to refer to a function is Haskell.
+- This syntax might appear alien to users. Haskell may be one of the only languages that use `\` to refer to a function.
 
 ## Alternatives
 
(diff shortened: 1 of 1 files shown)


andyfriesen:
The main reason we haven't pushed harder for a short lambda form in Luau is that we haven't been able to come up with a syntax that's succinct yet harmonious with the rest of the language.

The proposals in this RFC are certainly nice and succinct and they could be made to work in the parser, but they look way too different both from everything else in Luau and from lambdas in other programming languages.

goldenstein64:
I have a weird idea: what about flipping the order of the return expressions and parameters?

```lua
local add = do a + b for(a, b)

-- typed
local add = do a + b for(a: number, b: number): number

-- curried -- hard to read :(
-- function(a) return function(b) return a - b end end
local currySub = do do a - b for(b) for(a)

-- it may be better to allow for more param "groups" after the first?
-- this decision can be delayed!!!
local currySub = do a - b for(a)(b)

-- generics
local first = do ls[1] for<T>(ls: {T}): T

-- higher-order
local sum = foldLeft(numbers, do a + b for(a, b), 0)
```

Here's the grammar:

```
-- without curry params
lambdafunc = 'do' explist 'for' ['<' GenericTypeList '>'] '(' [parlist] ')' [':' ReturnType]

-- with curry params
lambdafunc = 'do' explist 'for' ['<' GenericTypeList '>'] '(' [parlist] ')' { '(' [parlist] ')' } [':' ReturnType]

simpleexp' = simpleexp | lambdafunc
```

I understand completely if it's too weird to read the expression first, but this seems to have a lot more things going for it in spite of that.

- Restrictions 1 and 2 are justified because the keywords make those problem cases look unnatural
- No new symbols
- Harmonizes better with other syntax
- Looks like Luau *I think*

It now has another issue though. Since the return expression is first, it may get confusing to know what is a parameter and what is an upvalue without reading the parameters first.

```lua
local addElem = do manager:appendElement(elem) for(manager, elem)

-- vs

local manager = ...
-- ...
local addElem = do manager:appendElement(elem) for(elem)
```

For long lambdas with many variables, this may be important. Then again, this could be solved by using an anonymous function (which should be the case for long lambdas), or by sticking to a different naming convention like short names.

The RFC's current syntax doesn't look function-y, but I'm not sure if this is any better. I am open to feedback!

bradsharp:
I'm very interested in any kind of sugar we could come up with using lambdas. There have been a few cases where this would be nice. For example, I could imagine it being used in something like Fusion.
```lua
-- In a table
X = Computed value => value * 2
-- Equivalent to
X = Computed(value => value * 2)
```
Note, I used JS lambda like syntax to avoid any bias.

bradsharp:
I like the `do` based syntax. I've tried to come up with a few versions of it myself but never found something I was satisfied with. If it wasn't ambiguous my preference would be:
```lua
(x, y) do x + y
```
Then you could use it in the previous example I gave like:
```lua
Computed (value) do value * 2
```
I wonder if we could do it with some variation for lambdas with variables versus those without:
```lua
local reset = do x = nil
local set = take value do x = value
```

vegorov-rbx:
My vote is to not have this in Luau, there's nothing wrong with using `function() end`.
Like mentioned in the previous RFCs, in the end it's such a non-issue to solve with such syntax tricks.

goldenstein64:
> My vote is to not have this in Luau, there's nothing wrong with using `function() end`.
> Like mentioned in the previous RFCs, in the end it's such a non-issue to solve with such syntax tricks.

The last RFC was not machine-parsable. The RFC before that and its accompanying proposal in a comment *was* machine-parsable, but didn't have all the features a function does. This one does both, but I forgot the most important thing: it should look like a Luau function. Evidently, it does not fulfill this requirement by a large margin.

PR #36, opened 2024-05-20
RFC: `table.pretty`, changing print to pretty print tables

LordOfSpelunky:
[Rendered](https://github.com/LordOfSpelunky/rfcs/blob/table-pretty/docs/pretty-tables.md)

### DIFF
--- a/docs/pretty-tables.md
+++ b/docs/pretty-tables.md
@@ -1,0 +1,142 @@
+# `table.pretty` and changes to printing tables
+
+## Summary
+
+We will add a function to display a pretty version of a table for debugging purposes. We will use this function as the backbone to changing the default implementation of `print(table)` in CLI contexts.
+
+## Motivation
+
+In Roblox Studio, printing out a table with "Log mode" off will display an interactive table. In roblox-cli, Lune, and luau.exe, it will print the memory address, which is faintly useful, and makes debugging tables a massive chore. Making tables pretty print is a clear win in this regard.
+
+Adding a function to create these pretty forms and using that as the underlying implementation behind `print` is useful for custom loggers that do not use print and go to file, or for something like debug commands.
+
+## Design
+
+### `table.pretty`
+
+The following function will be added to the standard library:
+
+`table.pretty(input: { [any]: any }, indentation: number?): string`
+
+`indentation` defaults to 2.
+
+For a non-empty table, this will perform the following behavior:
+
+1. The initial opening and ending brace will be on their own line
+2. Form the string `[$key]: $value,` for every key and value, based on the same behavior as generalized iteration, even using `__iter` if it exists.
+	- `$key` and `$value` will be pretty printed if it is both a table and if it would be acceptable to do so (no `__metatable`). This will be indented `indentation` number of spaces.
+	- If `$value` is a string, it will be wrapped in quotes.
+	- Otherwise, `$key` and `$value` will be the result of `tostring(value)`.
+	- The comma is always included.
+	- `$key` is not included if it is a number part of a sequence of keys starting with `1`.
+3. In the event of a cyclic table with no tostring implementation, the literal string `"<cyclic: memory address>"` will be put in its place.
+
+If indentation is 0, then it will all be put onto a single line, with spaces being used in place of new lines. In this case, trailing commas will not be present.
+
+#### Examples
+
+```lua
+print(table.pretty({
+	width = 12,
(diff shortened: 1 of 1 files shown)


docs/pretty-tables.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+#### Caveats
+An empty table will always produce "{}".
+
+Calling `table.pretty` on an input table with a custom `__metatable` will throw an error.
+
+### Changes to printing tables
+
+The default behavior of printing will change from calling `__tostring` to instead using `__tostring` if it exists, and `table.pretty(t)` if it does not. The existing behavior of printing a memory address will still be available through `print(tostring(t))`, which is already useful in a context like Roblox when working with Log mode.
```
LordOfSpelunky, line 126: Should this care about __metatable? I assume no since it just uses __iter which is already available anyway

docs/pretty-tables.md
```diff
@@ -0,0 +1,142 @@
...
+{
+  x = 1,
+  y = <cyclic: 0xDEADBEEF>,
+}
+```
+
+#### Caveats
+An empty table will always produce "{}".
+
+Calling `table.pretty` on an input table with a custom `__metatable` will throw an error.
```
LordOfSpelunky, line 122: Same comment here WRT __metatable, wonder if I'm caring too much

docs/pretty-tables.md
```diff
@@ -0,0 +1,142 @@
...
+
+-->
+
+```
+{
+  width = 12,
+  toppings = {
+    "pepperoni",
+    "sausage",
+  }
```
LordOfSpelunky, line 56:
```suggestion
  },
```

docs/pretty-tables.md
```diff
@@ -0,0 +1,142 @@
+# `table.pretty` and changes to printing tables
+
+## Summary
+
+We will add a function to display a pretty version of a table for debugging purposes. We will use this function as the backbone to changing the default implementation of `print(table)` in CLI contexts.
```
LordOfSpelunky, line 5: Wording of "CLI contexts" is misleading, I should remove that. This behavior would be just as useful in something like an in-game console, which is not really CLI.

docs/pretty-tables.md
```diff
@@ -0,0 +1,142 @@
...
+	- `$key` is not included if it is a number part of a sequence of keys starting with `1`.
+3. In the event of a cyclic table with no tostring implementation, the literal string `"<cyclic: memory address>"` will be put in its place.
+
+If indentation is 0, then it will all be put onto a single line, with spaces being used in place of new lines. In this case, trailing commas will not be present.
+
+#### Examples
+
+```lua
+print(table.pretty({
+	width = 12,
```
LordOfSpelunky, line 40: Nit: Need to change the indentation here or in the other Lua things. GitHub web editor just stuck to two spaces once I started using it elsewhere =P

andyfriesen:
I like function names to be verbs.  How about `table.pretty_print`?

deviaze:
> In roblox-cli, Lune, and luau.exe, it will print the memory address
Small nitpick in **Motivation**, but Lune already pretty-prints tables and does so in color. Would this be a breaking change for Lune's pretty-print implementation?

LordOfSpelunky:
Didn't know Lune already did that. It would not be a breaking change since that means Lune is already overriding print to begin with.

@andyfriesen Sure, but `table.prettyprint` right? That's Luau style with multiple words

kalrnlo:
I think table.pretty or table.prettify is more accurate, as table.prettyprint suggests that it actually prints to the console. Rather than just give a pretty string version of the table.

andyfriesen:
aha.  Well spotted.

Python calls it `pprint.pformat`, which I don't think I like very much.

vegorov-rbx:
Sounds like a very strange idea for a built-in functionality.

Should be just implemented in Luau as a library that people can use and customize.

Especially strange to see Roblox behaviors not only defined in open-source RFC, but request a different execution mode, so basically suggests that Roblox has to implement the method twice. I'm really against it.

CompeyDev:
> Would this be a breaking change for Lune's pretty-print implementation

No, lune implements its own, this would not change lune's ability to do so.

bradsharp:
Agree with @vegorov-rbx. Additionally, if we were to add this should it even support non-tables? I'd rather have a version of `concat` that worked nicely with recursion.

PR #38, opened 2024-05-30
RFC: `index` Type Operator

joonyoo181:
This RFC proposes the addition of one type operator, `index`, which can be used to look up a specific property of another type (like TypeScript's Indexed Access Type).

[Rendered](https://github.com/joonyoo181/luau-lang-rfcs/blob/jeremyyoo-index-type-operator/docs/index-type-operator.md)

### DIFF
--- a/docs/index-type-operator.md
+++ b/docs/index-type-operator.md
@@ -20,7 +20,7 @@ local bob: Person = {
   alive = true
 }
 
-local function doSmt(param: typeof(bob["age"])) -- param = string
+local function doSmt(param: typeof(bob["age"])) -- param = number
   -- rest of code
 end
 
--- a/docs/index-type-operator.md
+++ b/docs/index-type-operator.md
@@ -29,7 +29,7 @@ type unionType = typeof(bob["age"]) | typeof(bob["name"]) | typeof(bob["alive"])
 
 This is a valid Luau program; however, in order to define the type of `Person["age"]` we had to first declare a variable `bob` and utilize the `typeof` type operator. This is quite cumbersome when developers want to typecheck using the type of `Person["age"]` without having to declare a variable first. Additionally, in order to define the union type of all the properties of `Person`, current Luau requires an explicit list of each property using `typeof`.
 
-The expected outcome of the index type operator is that it will enhance the developing experience and allow the Roblox community to develop better type-checked programs.
+The expected outcome of the index type operator is that it will enhance developer experience and allow Luau developers to more easily develop well-typed programs.
 
 ## Design
 
--- a/docs/index-type-operator.md
+++ b/docs/index-type-operator.md
@@ -76,6 +76,8 @@ type idxType3 = index<Person | Person2, "age"> -- idxType3 = number | string
 type idxType4 = index<Person | Person2, "alive" | "age"> -- Error message: Property 'alive' does not exist on type 'Person2'.
 ```
 
+In the circumstance that the indexee is a type class or table with an `__index` metamethod, `index` will *only* invoke `__index` if indexer is not found within the current scope.
+
 Implementation is straight forward: the type of the indexee will be determined (table, class, etc) -> search through the properties of the indexee and reduce to the corresponding type of the indexer if it exists; otherwise, reduce to an error. 
 
 ## Drawbacks
--- a/docs/index-type-operator.md
+++ b/docs/index-type-operator.md
@@ -41,7 +41,7 @@ type Person = {
   alive: boolean
 }
 
-local function doSmt(param: index<Person, "age">) -- param = string
+local function doSmt(param: index<Person, "age">) -- param = number
   -- rest of code
 end
 
--- a/docs/index-type-operator.md
+++ b/docs/index-type-operator.md
@@ -76,7 +76,7 @@ type idxType3 = index<Person | Person2, "age"> -- idxType3 = number | string
 type idxType4 = index<Person | Person2, "alive" | "age"> -- Error message: Property '"age" | "alive"' does not exist on type 'Person | Person2'
 ```
 
-In the circumstance that the indexee is a type class or table with an `__index` metamethod, the `__index` metamethod will *only* be invoked if the indexer is not found within the current scope:
+In the circumstance that the indexee's type is a class or table with an `__index` metamethod, the `__index` metamethod will *only* be invoked if the indexer is not found within the current scope:
 ```luau
 local exampleClass = { Foo = "eight" }
 local exampleClass2 = setmetatable({ Foo = 8 }, { __index = exampleClass })
--- a/docs/index-type-operator.md
+++ b/docs/index-type-operator.md
@@ -76,7 +76,15 @@ type idxType3 = index<Person | Person2, "age"> -- idxType3 = number | string
 type idxType4 = index<Person | Person2, "alive" | "age"> -- Error message: Property '"age" | "alive"' does not exist on type 'Person | Person2'
 ```
 
-In the circumstance that the indexee is a type class or table with an `__index` metamethod, `index` will *only* invoke `__index` if indexer is not found within the current scope.
+In the circumstance that the indexee is a type class or table with an `__index` metamethod, the `__index` metamethod will *only* be invoked if the indexer is not found within the current scope:
+```lua
+local exampleClass = { Foo = "eight" }
+local exampleClass2 = setmetatable({ Foo = 8 }, { __index = exampleClass })
+local exampleClass3 = setmetatable({ Bar = "nine" }, { __index = exampleClass })
+
+type exampleTy2 = index<typeof(exampleClass2), "Foo"> -- exampleTy2 = number
+type exampleTy3 = index<typeof(exampleClass3), "Foo"> -- exampleTy3 = string
+```
 
 Implementation is straight forward: the type of the indexee will be determined (table, class, etc) -> search through the properties of the indexee and reduce to the corresponding type of the indexer if it exists; otherwise, reduce to an error. 
 
--- a/docs/index-type-operator.md
+++ b/docs/index-type-operator.md
@@ -61,27 +61,16 @@ If the given value used to access a property is not a type,
 local key = "age"
 type age = index<Person, key> -- Error message: Type 'key' cannot be used as an index type.
 ```
-
-One edge case for this feature is in the case of:
-```lua
-type Person2 = {
-  age: number,
-  name: string,
-  alive: boolean,
-  job: string
-}
-
-local function edgeCase(p: Person)
-  type unknownType = index<typeof(p), "job"> -- unknownType = unknown
-end
-```
-Although the literal type `job` is not a property in `Person`, the program should not error here, but instead return the type `unknown`. This is due to the fact that at run time, parameter `p` can be of type `Person` or `Person2` (since `Person2` is a subtype of `Person`). As a result, the index type of `job` on `Person` throws an error, but on `Person2` returns a type of `string`. As a result, at compile time, the type of `unknownType` should be `unknown`.
+Note: these errors will be part of the general type family reduction errors since `index` will be built into the type family system.
 
 Implementation is straight forward: the type of the indexee will be determined (table, array, etc) -> search through the properties of the indexee and return the corresponding type of the indexer if it exists; otherwise, return an error or unknown type depending on the scope. 
 
 ## Drawbacks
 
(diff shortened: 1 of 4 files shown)


docs/index-type-operator.md
```diff
@@ -0,0 +1,88 @@
...
+local function doSmt(param: typeof(bob["age"])) -- param = string
+  -- rest of code
+end
+
+type unionType = typeof(bob["age"]) | typeof(bob["name"]) | typeof(bob["alive"]) -- unionType = number | string | boolean
+```
+
+This is a valid Luau program; however, in order to define the type of `Person["age"]` we had to first declare a variable `bob` and utilize the `typeof` type operator. This is quite cumbersome when developers want to typecheck using the type of `Person["age"]` without having to declare a variable first. Additionally, in order to define the union type of all the properties of `Person`, current Luau requires an explicit list of each property using `typeof`.
+
+The expected outcome of the index type operator is that it will enhance the developing experience and allow the Roblox community to develop better type-checked programs.
```
aatxe, line 32: Luau is not specific to Roblox, so we'd prefer not to bring it up unnecessarily.
```suggestion
The expected outcome of the index type operator is that it will enhance developer experience and allow Luau developers to more easily develop well-typed programs.
```

docs/index-type-operator.md
```diff
@@ -0,0 +1,88 @@
...
+
+Error messages will be displayed for incorrect syntax. If the given type used to access a property is invalid, 
+```lua
+type age = index<Person, "ager"> -- Error message: Property 'ager' does not exist on type 'Person'.
+```
+If the given value used to access a property is not a type,
+```lua
+local key = "age"
+type age = index<Person, key> -- Error message: Type 'key' cannot be used as an index type.
+```
```
aatxe, line 63: It's worth probably stating explicitly since this type operator will be built into the type family system, these errors will be part of the existing errors for type family reduction errors. This sort of specific message is fine though, as the stringification for type family reduction errors already specializes the errors for the existing type operators for the sake of better DX.

docs/index-type-operator.md
```diff
@@ -0,0 +1,88 @@
...
+type Person2 = {
+  age: number,
+  name: string,
+  alive: boolean,
+  job: string
+}
+
+local function edgeCase(p: Person)
+  type unknownType = index<typeof(p), "job"> -- unknownType = unknown
+end
```
aatxe, line 76: This point can probably be made clearer: `index<Person, "job">` could reasonably be made to either fail to reduce (signaling an error) or can reduce to `unknown` (since all table types support width typing currently) because `edgeCase({ age: 4, name: "Foo", alive: true, job: "Chef" })` is a valid call. Ultimately, I think it probably makes sense to propose the erroring behavior though, since I suspect people are most likely going to use `index` in type aliases, rather than e.g. annotations inside of a function with `typeof` on a parameter.

docs/index-type-operator.md
```diff
@@ -0,0 +1,88 @@
...
+
+Implementation is straight forward: the type of the indexee will be determined (table, array, etc) -> search through the properties of the indexee and return the corresponding type of the indexer if it exists; otherwise, return an error or unknown type depending on the scope. 
+
+## Drawbacks
+
+The only drawback to implementing this feature is that supporting type operators will require a lot of work in the underlying infrastructure. However, the new type inference engine for Luau already contains the infrastructure needed to support type operators. In fact, the implementation of this new feature can be mirrored by the implementation of the `keyof` type operator as they share the same syntax and a small difference in semantics. As such, the author of this RFC hopes that this effort is a win for OOP in Luau, supported by the technical credit built up from the new type inference engine.
```
aatxe, line 84: This is way too copypasta'd from the `keyof` RFC. 😔 The way I think we probably ought to frame the drawback here is that the behavior being proposed for missing properties is potentially surprising, and discuss why we feel that isn't an issue (or change our minds about the behavior itself). Otherwise, the drawback is sort of just a general: adding more built-in type operators increases the cost of maintaining the type system over time, and the reason we feel `index` is worthwhile to include in spite of this is because it corresponds to a built-in operation in the language which is kind of the overarching principle right now for determining what things _should_ be a built-in type operator.

docs/index-type-operator.md
alexmccord: I kind of want synonymous behavior between `t.x` and `index<t, "x">`. Different behavior would be very surprising to me!

aatxe: Yeah, this is also the thing that convinced me too, reading it. I was like "Hmm, well, if you wrote `p.job` in this function, we'd produce a type error, even though, indeed, `p.job` at runtime might actually be fine."

Kampfkarren:
An alternative that ought to be listed is the TypeScript `T["key"]` format, and with Luau I'd expect T.key. I'd really like to know why we're not choosing either of those syntaxes, they are way more intuitive than thjs

aatxe:
> An alternative that ought to be listed is the TypeScript `T["key"]` format, and with Luau I'd expect T.key. I'd really like to know why we're not choosing either of those syntaxes, they are way more intuitive than this

This RFC introduces no new syntax at all. Type operators (like the kind you can define writing `type Array<T> = {T}`) already have the syntax `name<args>` in general, and this is just using that existing syntax. We can propose a follow-up RFC to add syntactic sugar for `T.key`.

docs/index-type-operator.md
```diff
@@ -0,0 +1,88 @@
...
+```lua
+type age = index<Person, "ager"> -- Error message: Property 'ager' does not exist on type 'Person'.
+```
+If the given value used to access a property is not a type,
+```lua
+local key = "age"
+type age = index<Person, key> -- Error message: Type 'key' cannot be used as an index type.
+```
+
+One edge case for this feature is in the case of:
```
Vighnesh-V, line 65: For a table with non string properties on it, what should `index` do? E.g. if we have a table `type Dict = { [number] : string }`, could index take a number?  I saw above that you could pass `"age" | "name"` as an argument to `index` so I was just curious.

docs/index-type-operator.md
```diff
@@ -0,0 +1,88 @@
...
+
+The solution to this problem is a type operator, `index`, that can compute the type based on the static properties of `Person`. Formally, the `index` type operator will take in two arguments: the type to index and the type to index it with. This would allow us to instead write the following code:
+```lua
+type Person = {
+  age: number,
+  name: string,
+  alive: boolean
+}
+
+local function doSmt(param: index<Person, "age">) -- param = string
```
deviaze, line 44: age param should be number, not string. 
```suggestion
local function doSmt(param: index<Person, "age">) -- param = number
```

docs/index-type-operator.md
```diff
@@ -0,0 +1,88 @@
...
+  alive: boolean
+}
+
+local bob: Person = {
+  age = 22,
+  name = "Bob",
+  alive = true
+}
+
+local function doSmt(param: typeof(bob["age"])) -- param = string
```
deviaze, line 23: found a smol typo: age should be a number here, not a string (as defined by the Person type on line 11)
```suggestion
local function doSmt(param: typeof(bob["age"])) -- param = number
```

deviaze (changes requested):
Small typo fixes on Person's "age" type

docs/index-type-operator.md
joonyoo181: yes,`index<Dict, number>` would be a valid expression. In this case, it will reduce to a type `string`.

docs/index-type-operator.md
joonyoo181: oops, thank you for catching this!

docs/index-type-operator.md
```diff
@@ -0,0 +1,77 @@
...
+end
+
+type idxType = index<Person, keyof<Person>> -- idxType = number | string | boolean
+
+type idxType2 = index<Person, "age" | "name"> -- idxType2 = number | string
+```
+
+Now, the type of `doSmt()`'s parameter can be defined without declaring a variable `bob`. Additionally, regardless of how the type `Person` grows, `idxType` will always be defined as the union of all the properties.
+
+Error messages will be displayed for incorrect syntax. If the given type used to access a property is invalid,
```
alexmccord, line 55:
```suggestion
Error messages will be displayed for incorrect type arguments. If the given type used to access a property is invalid, 
```

docs/index-type-operator.md
alexmccord, line 1: One observation is that this RFC doesn't mention what you'd do if the first argument is a union type, e.g. `index<T | U, "p">`.

joonyoo181: We will handle this case by distributing the arguments to be `index<T, "p"> | index<U, "p">`. I will mention this in my next commit, thank you for mentioning this!

alexmccord:
Speaking of, there probably should also be `rawget` which is the equivalent of `index` but without `__index` invocation. I'm using `rawget` instead of `rawindex` because there's a function called `rawget`.

alexmccord:
Here's a fun one: `{ p: number?, metatable { __index: { p: string } } }`. If at runtime `p: number?` is `nil`, then `__index` is invoked to find `p: string`. So the actual type of `index<T, "p">` is `number | string`, not `number?` or `(number | string)?`.

docs/index-type-operator.md
```diff
@@ -0,0 +1,121 @@
+# `index` type operator
+
+## Summary
+
+This RFC proposes the addition of one type operator, `index`, which can be used to look up a specific property of another type (like TypeScript's Indexed Access Type).
+
+## Motivation
+
+The primary motivation of this proposal is to allow Luau to define a type by accessing a property of another existing type. For instance, consider the following example code:
+```lua
```
alexmccord, line 10:
```suggestion
```luau
```

docs/index-type-operator.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+This is a valid Luau program; however, in order to define the type of `Person["age"]` we had to first declare a variable `bob` and utilize the `typeof` type operator. This is quite cumbersome when developers want to typecheck using the type of `Person["age"]` without having to declare a variable first. Additionally, in order to define the union type of all the properties of `Person`, current Luau requires an explicit list of each property using `typeof`.
+
+The expected outcome of the index type operator is that it will enhance developer experience and allow Luau developers to more easily develop well-typed programs.
+
+## Design
+
+The solution to this problem is a type operator, `index`, that can compute the type based on the static properties of `Person`. Formally, the `index` type operator will take in two arguments: the type to index (indexee) and the type to index it with (indexer). This would allow us to instead write the following code:
+```lua
```
alexmccord, line 37: Ditto.

docs/index-type-operator.md
```diff
@@ -0,0 +1,121 @@
...
+}
+
+-- equivalent of `index<Person, "age"> | index<Person2, "age">`
+type idxType3 = index<Person | Person2, "age"> -- idxType3 = number | string
+
+-- equivalent of `index<Person, "alive" | "age"> | index<Person2, "alive" | "age">`
+type idxType4 = index<Person | Person2, "alive" | "age"> -- Error message: Property '"age" | "alive"' does not exist on type 'Person | Person2'
+```
+
+In the circumstance that the indexee is a type class or table with an `__index` metamethod, the `__index` metamethod will *only* be invoked if the indexer is not found within the current scope:
```
aatxe, line 79:
```suggestion
In the circumstance that the indexee's type is a class or table with an `__index` metamethod, the `__index` metamethod will *only* be invoked if the indexer is not found within the current scope:
```

docs/index-type-operator.md
```diff
@@ -0,0 +1,121 @@
...
+end
+```
+In our current design, the program simply fails to reduce (and throws an error). However, it is worth noting that `index<typeof(p), "job">` can also reduce to type `unknown` because the parameter `p` can be of type `Person` or `Person3` (since tables support width subtyping; hence, `Person3` is a subtype of `Person`):
+- If `p` is of type `Person`, `index<typeof(p), "job">` should reduce to an error. 
+- If `p` is of type `Person3`, `index<typeof(p), "job">` should reduce to type `string`.
+
+Because there are conflicting types for `p` depending on the run time, it is safest for the program to reduce to a type `unknown`. In this design, we would need a way to determine if the indexee can be different at runtime. We could determine this through the implementation of more table types, specifically exact and inexact table types. Then, the program will have two cases when an indexee does not contain the indexer type:
+1. If the indexee is an inexact table type, reduce to an `unknown` type.
+2. If the indexee is an exact table type, fail to reduce and throw an error.
+
```
aatxe, line 120: I would add a brief bit here mentioning that we can consider adding syntactic sugar for the `index` type operator in the future, with suggestions like `type["index"]` and `type.index`.

aatxe, line 120: I would add a brief bit here mentioning that we can consider adding syntactic sugar for the `index` type operator in the future, with suggestions like `type["index"]` and `type.index`.

alexmccord:
One last edge case I just remembered: `__index` only supports 100 nested `__index` until it gives up.

aatxe (approved):

PR #39, opened 2024-06-08
RFC: `rawget` Type Operator

joonyoo181:
This RFC proposes the addition of another type operator, `rawget`, which can be used to look up a specific property of another type *without* invoking the `__index` metamethod. Other than not invoking the `__index` metamethod, this new type operator is equivalent to the `index` type operator in terms of behavior.

[Rendered](https://github.com/joonyoo181/luau-lang-rfcs/blob/jeremyyoo-rawget-type-operator/docs/rawget-type-operator.md)


docs/rawget-type-operator
alexmccord, line 1: Forgot `.md`

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Summary
+
+This RFC proposes the addition on one type operator, `rawget`, which can be used to look up a specific property of another type *without* invoking the `__index` metamethod.
+
+## Motivation
+
+There exists `index` type operator that allow developers to obtain a type of a property from classes / tables. If the a type is not found in the given class / table, the operator recursively indexes into the `__index` metamethod to continue searching for the property. Sometimes, this is could be an unwanted behavior. For example, given this code: 
+
+```lua
```
alexmccord, line 11:
```suggestion
```luau
```

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,58 @@
...
+
+This can be undesirable for developers who expect the `index` operator to only check the properties directly on the variable without considering the metamethod chain. In the given example, the expectation might be to receive an error when querying `var2` for `property`, indicating that the property does not exist on `var2`.
+
+To address this issue, we want to reconsider the behavior of the `index` type operator in the context of the `__index` metamethod. Developers may require a more predictable and controlled mechanism for type checking that does not involve recursively following the `__index` chain.
+
+## Design
+
+The proposed solution is the implementation of `rawget` type operator. Like the existing [rawget Lua Globals](https://create.roblox.com/docs/reference/engine/globals/LuaGlobals#rawget), this type operator aims to provide a way to look up a specific property of a type without invoking the `__index` metamethod. 
+ 
+```lua
```
alexmccord, line 30: Ditto. Also everything else too.

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,58 @@
+# `rawget` type operator
+
+## Summary
+
+This RFC proposes the addition on one type operator, `rawget`, which can be used to look up a specific property of another type *without* invoking the `__index` metamethod.
+
+## Motivation
+
+There exists `index` type operator that allow developers to obtain a type of a property from classes / tables. If the a type is not found in the given class / table, the operator recursively indexes into the `__index` metamethod to continue searching for the property. Sometimes, this is could be an unwanted behavior. For example, given this code:
```
vegorov-rbx, line 9:
```suggestion
There exists an `index` type operator that allows developers to obtain a type of a property from classes / tables. If a type is not found in the given class / table, the operator recursively indexes into the `__index` metamethod to continue searching for the property. Sometimes, this could be an unwanted behavior. For example, given this code: 
```

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,58 @@
...
+```luau
+local var1 = {
+  property = "hello"
+}
+
+local var2 = setmetatable({ }, { __index = seq })
+
+type doesExist = index<typeof(var2), "property"> -- reduces to `string`
+```
+The `index` type operator is used to obtain the type of `property` from a class `var1`. The operator continues to search recursively through the `__index` metamethod if the property is not initially found. This behavior can lead to unexpected results, as shown in the code snippet where the `__index` metamethod of `var2` points to `var1`, causing the `index` operator to return `string` for the `property` key even though `var2` itself does not have a property.
```
vegorov-rbx, line 20: Don't really understand why it's unexpected given that it exactly matches the indexing at runtime.

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,58 @@
...
+
+This can be undesirable for developers who expect the `index` operator to only check the properties directly on the variable without considering the metamethod chain. In the given example, the expectation might be to receive an error when querying `var2` for `property`, indicating that the property does not exist on `var2`.
+
+To address this issue, we want to reconsider the behavior of the `index` type operator in the context of the `__index` metamethod. Developers may require a more predictable and controlled mechanism for type checking that does not involve recursively following the `__index` chain.
```
vegorov-rbx, line 24: Again, don't understand what's unpredictable about how indexing in the language works if the type system matches that.

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,58 @@
...
+
+local key = "age"
+type age = rawget<Person, key> -- Error message: Second argument to rawget<Person,_> is not a valid index type; Unknown type 'key'
+```
+
+The implementation effort for this type operator is relatively small. Since the `rawget` type operator functions similarly to the `index` type operator, we can reuse the functions already used to implement the `index` type operator.
+
+## Drawbacks
+
+There aren't appreciable drawbacks. One possible drawback is the increase in the size of the codebase, which may complicate maintenance. However, since code for `index` type operators already exists, the additional code needed to implement this operator will be minimal.
```
vegorov-rbx, line 52: Drawbacks is not only about code that is written, I would even say that code is easy to write.
One example of a drawback for the additions to the type language is the amount of details the users of the language have to learn.

docs/rawget-type-operator.md
vegorov-rbx, line 20: `var1` is not a class, it's a table.

vegorov-rbx (changes requested):
This RFC talks about classes and tables as if it's the same thing.

Considering that what we call classes right now are `ClassType` types which are used to represent 'userdata' values, this RFC would not apply to them correctly.

Main problem is that `userdata` doesn't have direct fields, the indexing is always resolved using a metamethod lookup.
`rawget` runtime method only works on tables, which would mean that `rawget` should never reduce for classes.
If `rawget` would reduce for classes, it would not be match runtime `rawget` and that will confuse developers who expect a type system behavior that somehow mirrors values at runtime.

So this RFC fails to include classes in the design correctly.

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,58 @@
...
+  property = "hello"
+}
+
+local var2 = setmetatable({ }, { __index = seq })
+
+type doesExist = index<typeof(var2), "property"> -- reduces to `string`
+```
+The `index` type operator is used to obtain the type of `property` from a class `var1`. The operator continues to search recursively through the `__index` metamethod if the property is not initially found. This behavior can lead to unexpected results, as shown in the code snippet where the `__index` metamethod of `var2` points to `var1`, causing the `index` operator to return `string` for the `property` key even though `var2` itself does not have a property.
+
+This can be undesirable for developers who expect the `index` operator to only check the properties directly on the variable without considering the metamethod chain. In the given example, the expectation might be to receive an error when querying `var2` for `property`, indicating that the property does not exist on `var2`.
```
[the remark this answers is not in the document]
aatxe: The real motivation here should pretty much be "the ability to type generic code using `rawget`, much like the `index` type operator enables the typing of generic code using indexing."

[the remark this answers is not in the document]
vegorov-rbx: It's just that `rawget` is rarely used (e.g. 84 calls in one specific multi-Mb Luau codebase) compared to indexing (immeasurable in that codebase).
It's not even clear how many of those 84 places need the type operator and cannot be just inferred.

Rarely used doesn't mean that it's not worth to have it for a full type system, but makes me wonder what motivated (heh) the inclusion so early in the dev cycle.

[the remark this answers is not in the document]
vegorov-rbx: It does help that the implementation is easy.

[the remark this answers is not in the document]
vegorov-rbx: Rarity of `rawget` also reinforces my comments about how 'unpredictable' and 'unexpected' the most common way to index tables is claimed to be in this RFC.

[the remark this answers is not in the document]
vegorov-rbx: Closing holes in type language is a good motivation.

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,58 @@
...
+## Motivation
+
+There exists an `index` type operator that allows developers to obtain a type of a property from classes / tables. If a type is not found in the given class / table, the operator recursively indexes into the `__index` metamethod to continue searching for the property. Sometimes, this could be an unwanted behavior. For example, given this code: 
+
+```luau
+local var1 = {
+  property = "hello"
+}
+
+local var2 = setmetatable({ }, { __index = seq })
```
Vighnesh-V, line 16: Is this meant to be `{ __index = var1}`?

joonyoo181: yes! thank you for pointing this out. my most recent commit has these mistakes patched

joonyoo181:
> This RFC talks about classes and tables as if it's the same thing.
> 
> Considering that what we call classes right now are `ClassType` types which are used to represent 'userdata' values, this RFC would not apply to them correctly.
> 
> Main problem is that `userdata` doesn't have direct fields, the indexing is always resolved using a metamethod lookup. `rawget` runtime method only works on tables, which would mean that `rawget` should never reduce for classes. If `rawget` would reduce for classes, it would not be match runtime `rawget` and that will confuse developers who expect a type system behavior that somehow mirrors values at runtime.
> 
> So this RFC fails to include classes in the design correctly.

Thank you for all the suggestions! On my most recent commit, I have addressed the mistakes you've pointed out and have made changes to the motivation and design section to better reflect the behavior of this type operator. Please let me know if there are anything else that look off in this RFC :)

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,51 @@
+# `rawget` type operator
+
+## Summary
+
+This RFC proposes the addition on one type operator, `rawget`, which can be used to look up a specific property of another type *without* invoking the `__index` metamethod.
```
vegorov-rbx, line 5:
```suggestion
This RFC proposes the addition of a new type operator, `rawget`, which can be used to look up a specific property of a table type *without* invoking the `__index` metamethod.
```

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Motivation
+
+Given that `rawget` is a built-in runtime operator in the language ([rawget Lua Globals](https://create.roblox.com/docs/reference/engine/globals/LuaGlobals#rawget)), it feels natural for there to be a version of this as a type operator. As such, the main motivation behind this feature is to close out holes in Luau's type families and allow Luau developers to be more expressive writing typed code:
```
vegorov-rbx, line 9:
```suggestion
Given that `rawget` is a built-in runtime operator in the language ([rawget in Luau Global functions](https://luau-lang.org/library#global-functions)), it feels natural for there to be a version of this as a type operator. As such, the main motivation behind this feature is to close out holes in Luau's type families and allow Luau developers to be more expressive writing typed code:
```
While Luau documentation doesn't have a specific anchor for the method (PRs welcome), it's best to refer to open-source Luau and the right language.

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,51 @@
...
+
+```luau
+local prop: rawget<someTy, someProp> = rawget(someTy, someProp)
+```
+
+## Design
+
+The functionality of `rawget` type operator behaves the same as its [runtime pair](https://create.roblox.com/docs/reference/engine/globals/LuaGlobals#rawget): provide a way to look up a specific property of a type without invoking the `__index` metamethod.
```
vegorov-rbx, line 17:
```suggestion
The functionality of `rawget` type operator behaves the same as its [runtime counterpart](https://luau-lang.org/library#global-functions): provide a way to look up a specific property of a table type without invoking the `__index` metamethod. 
```

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,51 @@
...
+
+The implementation effort for this type operator is very minimal. Since the `rawget` type operator functions similarly to the `index` type operator, we can reuse the functions already used to implement the `index` type operator.
+
+## Drawbacks
+
+There aren't appreciable drawbacks. One possible drawback is the increase in the size of the codebase, which may complicate maintenance. However, this is not expected to be a major drawback since the code for `index` type operators already exists. By reusing the code, it is even fair to say that there will be less than 50 lines of code needed to implement this type operator.
+
+Another drawback can come from the extra knowledge a user may need in order to use this type operator. For example, users will need to know the inner workings of types and how different operations interact with them (for instance, how `index` interacts with metatables that have the `__index` metamethod). However, there are documentations that outline each interactions, and thus this drawback does not seem to pose an issue.
```
vegorov-rbx, line 47:
```suggestion
Another drawback can come from the extra knowledge a user may need in order to use this type operator. For example, users will need to know the inner workings of types and how different operations interact with them (for instance, how `index` interacts with metatables that have the `__index` metamethod). However, there are documentations that outline each interaction, and thus this drawback does not seem to pose an issue.
```

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,51 @@
+# `rawget` type operator
+
+## Summary
+
+This RFC proposes the addition of a new type operator, `rawget`, which can be used to look up a specific property of a table type *without* invoking the `__index` metamethod.
+
+## Motivation
+
+Given that `rawget` is a built-in runtime operator in the language ([rawget in Luau Global functions](https://luau-lang.org/library#global-functions)), it feels natural for there to be a version of this as a type operator. As such, the main motivation behind this feature is to close out holes in Luau's type families and allow Luau developers to be more expressive writing typed code:
```
aatxe, line 9:
```suggestion
Currently, `rawget` is a built-in runtime operator in the language ([rawget in Luau Global functions](https://luau-lang.org/library#global-functions)) that is missing a corresponding type operator that captures its behavior. This RFC seeks to address this hole in the type system by providing a builtin type operator for `rawget` that will allow Luau developers to express more accurate types in their code:
```

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,51 @@
...
+
+```luau
+local prop: rawget<someTy, someProp> = rawget(someTy, someProp)
+```
+
+## Design
+
+The functionality of `rawget` type operator behaves the same as its [runtime counterpart](https://luau-lang.org/library#global-functions): provide a way to look up a specific property of a table type without invoking the `__index` metamethod.
```
aatxe, line 17:
```suggestion
The functionality of the `rawget` type operator behaves the same as its [runtime counterpart](https://luau-lang.org/library#global-functions). Namely, it provide a way to look up a specific property of a table type without invoking the `__index` metamethod. 
```

docs/rawget-type-operator.md
```diff
@@ -0,0 +1,51 @@
...
+
+The implementation effort for this type operator is very minimal. Since the `rawget` type operator functions similarly to the `index` type operator, we can reuse the functions already used to implement the `index` type operator.
+
+## Drawbacks
+
+There aren't appreciable drawbacks. One possible drawback is the increase in the size of the codebase, which may complicate maintenance. However, this is not expected to be a major drawback since the code for `index` type operators already exists. By reusing the code, it is even fair to say that there will be less than 50 lines of code needed to implement this type operator.
+
+Another drawback can come from the extra knowledge a user may need in order to use this type operator. For example, users will need to know the inner workings of types and how different operations interact with them (for instance, how `index` interacts with metatables that have the `__index` metamethod). However, there are documentations that outline each interaction, and thus this drawback does not seem to pose an issue.
```
aatxe, line 47: I think I'd say more that it's not a _new_ issue because the semantics of the language already features this distinction, and so programmers already need to be aware of the differing semantics of indexing and `rawget`.

aatxe (approved):
Nit, but otherwise, LGTM.

PR #42, opened 2024-06-11
RFC: Export Keyword for Values

bradsharp:
Allow users to export methods and values from their libraries directly through the use of the export keyword.

[Rendered View](https://github.com/luau-lang/rfcs/blob/rfc-export-keyword/docs/export-keyword.md)

### DIFF
--- a/docs/export-keyword.md
+++ b/docs/export-keyword.md
@@ -2,125 +2,162 @@
 
 ## Summary
 
-Allow users to export methods and values from their libraries through the use of the export keyword.
+Extend the `export` keyword to support values and functions.
 
 ## Motivation
 
-Users can export types from their libraries through the use of the export keyword.
+Today, it is possible to export a type from a module with the `export` keyword:
 
 ```luau
 export type Point = { x: number, y: number }
 ```
 
-Consumers of these libraries are able to access exported types by indexing the libraries identifier.
-
-```luau
-local Library = require("Library")
-local point : Library.Point = { x = 0, y = 0 }
-```
-
-Right now, this is the only use of the export keyword but it would be great if we let users export more with it to offer more utility and make exporting library methods and values easier.
+However, this mechanism is currently limited to types. Extending `export` to values and functions would provide a consistent way to expose a stable API from a module, while avoiding the extra boilerplate of returning a table.
 
 ## Design
 
 ### Syntax
-You can prefix any non-local declaration with the export keyword in the top level scope of a module. For values this looks like:
+
+Allow `export` in declarations anywhere `local` is allowed, but only at the top-level of a module. It may prefix a value, function, or type:
 
 ```luau
 export version = "5.1"
+
+export function init()
+	-- do a thing
+end
+
+export type Point = { x: number, y: number }
 ```
 
-While methods use:
+`export` for values is not permitted inside function bodies or other non–top-level scopes. Doing so results in a parse error.
+
+### Semantics
+
+#### Const By-Default
+
+Exported values and functions are implicitly [`const`](https://github.com/luau-lang/rfcs/pull/166). This means that exports must be initialized at the point of declaration and cannot be reassigned after initialization.
 
 ```luau
-export function init()
-  -- do a thing
+export foo = 5
+foo = 6 -- error: exported bindings are const
+```
+
+Without this, it would be possible to mutate the variable without changing to the exported table, leading to the following foot-gun:
+
+```luau
+export foo = 5
+
+export function increment()
+	foo += 1 -- 'foo' won't change in the exported table
 end
 ```
 
-### Behavior
+#### Export Table Construction
 
-When a value or method is prefixed with export, it is automatically added to a hidden export table which is frozen as soon as the module returns.
+Rather than populating an export table incrementally, the set of exported bindings is collected during parsing. When module execution completes, the module implicitly returns a frozen table containing the exported bindings.
 
-Take the following module:
+For example, the following module:
 
 ```luau
 export type Point = { x: number, y: number }
 
 export function distance(a: Point, b: Point)
-  local x, y = a.X - b.X, a.Y - b.Y
-  return math.sqrt(x * x + y * y)
+	local x, y = a.X - b.X, a.Y - b.Y
+	return math.sqrt(x * x + y * y)
 end
 ```
 
-This essentially becomes sugar for:
+Can be considered sugar for:
 
 ```luau
-local _EXT = {}
+export type Point = { x: number, y: number }
 
-type Point = { x: number, y: number }
-type _EXT.Point = Point -- note: this doesn't actually work today but one day it should
-
-local function distance(a: Point, b: Point)
-  local x, y = a.X - b.X, a.Y - b.Y
-  return math.sqrt(x * x + y * y)
+const function distance(a: Point, b: Point)
+	local x, y = a.X - b.X, a.Y - b.Y
+	return math.sqrt(x * x + y * y)
 end
-_EXT.distance = distance
 
-return table.freeze(_EXT)
+return table.freeze({
+	distance = distance,
+})
 ```
 
-If the user attempts to assign to an exported identifier then we would throw an error explaining that the interface cannot be changed once it has been exported.
+#### Export Shorthand
 
-### Nuances
-Due to the implementation, most things should "just-work". Here are some examples to consider:
+For convenience, we provide a shorthand form of `export` which will export an existing binding with the same name:
 
-#### Calling an Exported Function
-You can call an exported function as it's registered as a local before being added to the export table.
+```luau
+const foo = computeFoo()
+export foo
+```
+
+#### Order of Declarations and Mutual Dependencies
+
+As with `local` and `const` exports are evaluated in source order. Using the shorthand from above, mutually recursive or dependent functions can be declared before they are exported.
+
+```luau
+function a()
+	return b()
+end
+
+function b()
+	return 42
+end
+
+export a
+export b
+```
+
+Exporting before a binding is initialized is an error. Supporting hoisted exports or forward declarations is out of scope for this proposal and may be explored in a future RFC.
+
+#### Calling Exported Functions Internally
+
+Exported functions are still available within the module and can be called normally:
 
 ```luau
 export function distance(a: Point, b: Point)
-  local x, y = a.X - b.X, a.Y - b.Y
-  return math.sqrt(x * x + y * y)
+	return math.sqrt((a.X - b.X)^2 + (a.Y - b.Y)^2)
 end
 
 distance({0, 0}, {1, 1})
 ```
 
 #### Nested Tables
-You can export tables with additional values inside of them.
+
+Exporting a table exports the binding, not the contents of the table:
 
 ```luau
 export triangle = {}
 
 function triangle.draw()
-
+	-- ...
 end
 ```
 
-Something important to note here is that the nested table, `triangle` is not frozen.
+The `triangle` binding is immutable, but the table itself is mutable unless explicitly frozen.
 
 #### Returns
-Today, you can use the export keyword along with a return statement at the end of your module. If you use the export keyword with a value however we will throw an error if you also attempt to return.
+
+A module that uses `export` for values or functions may not also return a custom value. Attempting to do so is an error:
 
 ```luau
 export function distance(a: Point, b: Point)
-  local x, y = a.X - b.X, a.Y - b.Y
-  return math.sqrt(x * x + y * y)
+	return math.sqrt((a.X - b.X)^2)
 end
 
-return table.freeze {
-  distance = distance
-}
+return { distance = distance } -- error
 ```
 
+Type-only exports may continue to coexist with an explicit return value, as they do today.
+
 ## Drawbacks
-The keyword already exists for types so there's not much cost in us adding support to values. The main drawback is that it's another way to do module exports. We already have a way to do that, do we really need another?
+
+* Introduces another way to define module exports alongside explicit return tables.
+* Requires exported bindings to be immutable, which may require restructuring some existing patterns.
+* Extends the surface area of the `export` keyword beyond types.
 
 ## Alternatives
 
-### Do Nothing
-It's already possible to export values from modules using a return statement. We don't actually need to do this, it's more of a nice-to-have.
-
-### Automatically Export Globals
-Luau has a separate global scope for each module rather than a shared global scope across the entire program. We could automatically export any values that are stored in the global scope. This isn't backwards compatible though, we'd likely need a new import mechanism to resolve this.
+- Modules could implicitly export all top-level bindings. This is not backwards compatible, reduces explicitness, and complicates reasoning about module APIs.
+- Modules can already export values by returning tables explicitly. This proposal is a quality-of-life and consistency improvement but is not strictly necessary.
--- a/docs/export-keyword.md
+++ b/docs/export-keyword.md
@@ -18,7 +18,7 @@ However, this mechanism is currently limited to types. Extending `export` to val
 
 ### Syntax
 
-Allow `export` in declarations anywhere `local` is allowed, but only at the top-level of a module. It may prefix a value, function, or type:
+Allow `export` in declarations anywhere that defines a name (`local`s, `function`s, and `type`s), but only at the top-level of a module:
 
 ```luau
 export version = "5.1"
@@ -92,7 +92,7 @@ const foo = computeFoo()
 export foo
 ```
 
-If `foo` was declared as `local` instead of `const` then subequent uses of it here will be treates as if it were declared as `const`.
+If `foo` was declared as `local` instead of `const` then subsequent uses of it here will be treated as if it were declared as `const`.
 
 #### Order of Declarations and Mutual Dependencies
 
--- a/docs/export-keyword.md
+++ b/docs/export-keyword.md
@@ -100,7 +100,7 @@ end
 
 Something important to note here is that the nested table, `triangle` is not frozen.
 
-#### Multiple Returns
+#### Returns
 Today, you can use the export keyword along with a return statement at the end of your module. If you use the export keyword with a value however we will throw an error if you also attempt to return.
 
 ```luau
--- a/docs/export-keyword.md
+++ b/docs/export-keyword.md
@@ -92,6 +92,8 @@ const foo = computeFoo()
 export foo
 ```
 
+If `foo` was declared as `local` instead of `const` then subequent uses of it here will be treates as if it were declared as `const`.
+
 #### Order of Declarations and Mutual Dependencies
 
 As with `local` and `const` exports are evaluated in source order. Using the shorthand from above, mutually recursive or dependent functions can be declared before they are exported.
--- a/docs/export-keyword.md
+++ b/docs/export-keyword.md
@@ -1,0 +1,126 @@
+# Export Keyword
+
+## Summary
+
+Allow users to export methods and values from their libraries through the use of the export keyword.
+
+## Motivation
+
+Users can export types from their libraries through the use of the export keyword.
+
+```luau
+export type Point = { x: number, y: number }
+```
+
+Consumers of these libraries are able to access exported types by indexing the libraries identifier.
+
+```luau
+local Library = require("Library")
+local point : Library.Point = { x = 0, y = 0 }
+```
+
+Right now, this is the only use of the export keyword but it would be great if we let users export more with it to offer more utility and make exporting library methods and values easier.
+
+## Design
+
+### Syntax
+You can prefix any non-local declaration with the export keyword in the top level scope of a module. For values this looks like:
+
+```luau
+export version = "5.1"
+```
+
+While methods use:
+
+```luau
+export function init()
+  -- do a thing
+end
+```
+
+### Behavior
+
+When a value or method is prefixed with export, it is automatically added to a hidden export table which is frozen as soon as the module returns.
+
+Take the following module:
+
+```luau
+export type Point = { x: number, y: number }
+
+export function distance(a: Point, b: Point)
+  local x, y = a.X - b.X, a.Y - b.Y
+  return math.sqrt(x * x + y * y)
+end
+```
+
+This essentially becomes sugar for:
+
+```luau
+local _EXT = {}
+
+type Point = { x: number, y: number }
+type _EXT.Point = Point -- note: this doesn't actually work today but one day it should
+
+local function distance(a: Point, b: Point)
+  local x, y = a.X - b.X, a.Y - b.Y
+  return math.sqrt(x * x + y * y)
+end
+_EXT.distance = distance
+
+return table.freeze(_EXT)
+```
+
+If the user attempts to assign to an exported identifier then we would throw an error explaining that the interface cannot be changed once it has been exported.
+
+### Nuances
+Due to the implementation, most things should "just-work". Here are some examples to consider:
+
+#### Calling an Exported Function
+You can call an exported function as it's registered as a local before being added to the export table.
+
+```luau
+export function distance(a: Point, b: Point)
+  local x, y = a.X - b.X, a.Y - b.Y
+  return math.sqrt(x * x + y * y)
+end
+
+distance({0, 0}, {1, 1})
+```
+
+#### Nested Tables
+You can export tables with additional values inside of them.
+
+```luau
+export triangle = {}
+
+function triangle.draw()
+
+end
+```
+
+Something important to note here is that the nested table, `triangle` is not frozen.
+
+#### Multiple Returns
+Today, you can use the export keyword along with a return statement at the end of your module. If you use the export keyword with a value however we will throw an error if you also attempt to return.
+
+```luau
+export function distance(a: Point, b: Point)
+  local x, y = a.X - b.X, a.Y - b.Y
+  return math.sqrt(x * x + y * y)
+end
+
+return table.freeze {
+  distance = distance
+}
+```
+
+## Drawbacks
+The keyword already exists for types so there's not much cost in us adding support to values. The main drawback is that it's another way to do module exports. We already have a way to do that, do we really need another?
+
+## Alternatives
+
+### Do Nothing
+It's already possible to export values from modules using a return statement. We don't actually need to do this, it's more of a nice-to-have.
+
+### Automatically Export Globals
+Luau has a separate global scope for each module rather than a shared global scope across the entire program. We could automatically export any values that are stored in the global scope. This isn't backwards compatible though, we'd likely need a new import mechanism to resolve this.


docs/export-keyword.md
```diff
@@ -0,0 +1,126 @@
...
+```luau
+export function distance(a: Point, b: Point)
+  local x, y = a.X - b.X, a.Y - b.Y
+  return math.sqrt(x * x + y * y)
+end
+
+return table.freeze {
+  distance = distance
+}
+```
```
aatxe, line 115: Perhaps worth mentioning that the future change that would enable
`type _EXT.Point = Point -- note: this doesn't actually work today but one day it should` would also likely necessitate this restriction on explicit returning as well, since the way we'd likely be able to make that possible is by giving tables the ability to have (phantom, i.e. non-existent at runtime) types associated with them as a value.

bradsharp: To me that implies we need to make this functional (even if it's bad). We can't silently drop the second return, and we can't let it override the returned table. If modules supported multiple returns we might be able to get around this by making sure the export table is always the last value returned but I suspect @vegorov-rbx might have opinions on that.

aatxe: Sorry, maybe I wasn't clear. I think the correct behavior here is that we should throw an error on any explicit return in a module with `export function` or `export id`, and I also think that even though this may seem a bit odd right now since the restriction is specific to exporting values, it's also a restriction we will have if we do the work to enable assigning a type into a "field" (not one that exists at runtime) as you had written in the earlier example.

docs/export-keyword.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Drawbacks
+The keyword already exists for types so there's not much cost in us adding support to values. The main drawback is that it's another way to do module exports. We already have a way to do that, do we really need another?
+
+## Alternatives
+
+### Do Nothing
+It's already possible to export values from modules using a return statement. We don't actually need to do this, it's more of a nice-to-have.
```
alexmccord, line 123: Note that it's not entirely syntactic sugar.

```lua
local mod = {}

function mod.distance(p1: Point, p2: Point)
  local x, y = a.X - b.X, a.Y - b.Y
  return math.sqrt(x * x + y * y)
end
```

If you wanted to call `mod.distance` somewhere else in this module, it will not get inlined because in theory, `mod.distance` could have been transmuted to a different function with a different implementation.

docs/export-keyword.md
alexmccord, line 1: One thing I noticed not mentioned here is how this would work with two mutually dependent exported functions. If you replace `export` with `local` in this snippet, it doesn't do what you want it to do.

```lua
export function f()
  g()
end

export function g()
  f()
end
```

alexmccord, line 1: Another thing: what happens if you use `export` in a scope that isn't the module scope?

```luau
function what()
  export function f() end -- ????
end
```

docs/export-keyword.md
```diff
@@ -0,0 +1,126 @@
...
+
+function triangle.draw()
+
+end
+```
+
+Something important to note here is that the nested table, `triangle` is not frozen.
+
+#### Multiple Returns
+Today, you can use the export keyword along with a return statement at the end of your module. If you use the export keyword with a value however we will throw an error if you also attempt to return.
```
alexmccord, line 104: Throw an error as in syntactic? Or compile error?

docs/export-keyword.md
```diff
@@ -0,0 +1,126 @@
...
+export triangle = {}
+
+function triangle.draw()
+
+end
+```
+
+Something important to note here is that the nested table, `triangle` is not frozen.
+
+#### Multiple Returns
```
alexmccord, line 103:
```suggestion
#### Returns
```

docs/export-keyword.md
aatxe: I think my personal expectation here is that this is going to behave the same way as locals, and we can perhaps provide a plain `export f` `export g` (though I suspect syntactic ambiguity will be an issue here) to let you export a global.

I would be open to hearing reasons why it should go the other way instead and behave like globals, but I think regardless of the choice, we don't want to introduce a new third semantics here, `export` should behave like one of them.

docs/export-keyword.md
aatxe: I'd imagine compile-time.

alexmccord: I thought so too, but wanted this to be specified in the RFC.

Also, note that compile-time errors are not shown to the user at edit time currently, and has a problem of only throwing one compiler error. We'll need a linter to warn on mixed use of `export`/`return`.

docs/export-keyword.md
bradsharp: Agree with @aatxe on this one.
If we want to support the described behavior I think we should look into hoisting but that would be a separate RFC.

bradsharp: I've actually been wondering how this should work. My feeling is that we either don't allow it or we make it work exactly how it does in the modules function scope. In other words, using `export` in any function scope creates a module table and returns it as that functions return value. It would be a weird language quirk so I'm leaning towards "it's not allowed" but maybe that's done through linting and not something we actually enforce.

docs/export-keyword.md
bradsharp: Seems like it should be a parse error similar to
```
return "Foo"
return "Bar"
```

docs/export-keyword.md
bradsharp: This is true, and also worth calling out. I think it's unlikely we'd expose `_EXT` to users unless we think there's a valid use-case (maybe metatable stuff?)

alexmccord: Yeah I don't want to expose `_EXT`. It's trivial for the VM to not expose it too, it already does this with loops which introduces 3 locals you can't even use.

alexmccord: I also question the value of adding metatable to the export table. I've never seen anyone want to do this, and if they do they can use `return` with no `export`s.

docs/export-keyword.md
aatxe: Yeah, I suppose it's not hard to make it a parsing error, you just keep a little bit of state to record if you've already seen an `export`/`return` and then whichever you see second can produce an error about it.

nezuo:
I think this is a great UX improvement! I do have one issue with it:

When you require a module with no return in Roblox, it gives the error: "Module code did not return exactly one value."

Having to specify a return by default and then remove it once you add something you want to export would be a bit annoying. I'm not sure what the right solution would be. Is it possible to change the behavior of modules that don't return anything?

aatxe:
> Having to specify a return by default and then remove it once you add something you want to export would be a bit annoying. I'm not sure what the right solution would be. Is it possible to change the behavior of modules that don't return anything?

I think it would be reasonable to have an implicit `return {}` if no return statement is present since I think that's more-or-less what this RFC is proposing as the meaning of a module (i.e. it returns a table, always).

docs/export-keyword.md
bradsharp: Agree with this so let's keep it locked down and if legitimate use-cases come up we can discuss as part of another RFC.

bradsharp:
> > Having to specify a return by default and then remove it once you add something you want to export would be a bit annoying. I'm not sure what the right solution would be. Is it possible to change the behavior of modules that don't return anything?
> 
> I think it would be reasonable to have an implicit `return {}` if no return statement is present since I think that's more-or-less what this RFC is proposing as the meaning of a module (i.e. it returns a table almost always).

Not sure I agree with `return {}` being the default. I think it should be `return;` since you didn't return or export anything.

ccuser44 (commented):
I very much would like this feature in Luau for many reasons (expecially easier AST analysis and standardization), but because it's a syntax change I don't think it meets the bar to be added as a syntax feature.

ccuser44 (commented):
With further introspection I'm leaning fully on the side of not accepting this RFC. While nice in theory the practical gains of the feature are marginal and (in my opinion) don't justifu the increased lexer/parser complexity.

aatxe:
> > > Having to specify a return by default and then remove it once you add something you want to export would be a bit annoying. I'm not sure what the right solution would be. Is it possible to change the behavior of modules that don't return anything?
> > 
> > 
> > I think it would be reasonable to have an implicit `return {}` if no return statement is present since I think that's more-or-less what this RFC is proposing as the meaning of a module (i.e. it returns a table almost always).
> 
> Not sure I agree with `return {}` being the default. I think it should be `return;` since you didn't return or export anything.

This is already what the behavior is (leaving nothing on the stack) though, and what's why you get an error saying it has to return exactly one value. Returning nothing is returning zero values! There's a question of whether `ModuleScript` could potentially just be okay with returning zero values as well, but that is technically out of scope for the RFC itself.

docs/export-keyword.md
alexmccord: Given that `export` probably only makes sense at module level, it stands to reason they should work like globals.

alexmccord:
There's an opportunity to talk about the possibility of adding re-exports here that I think this misses:

```luau
export local M = require("./foo")`
```

Which can re-export both types and values.

ccuser44:
> As someone who has written a luau parser in luau, and contributed to other luau parsers, the added parsing complexity is just about 0.

Yeah I don't know why I just said that. Adding the export keyword would only require adding a new lex token type and having it generate a `SETTABLE` instruction that adds the var to the export table.
My point was is that I don't think I don't think this RFC meets the high bar of usefulness that's required for new syntax features. I was originally for approving this RFC but now due to further introsoection I'm againts the idea.

dphfox:
Here's a question; right now all imports are namespaced, e.g. you access them via `ModuleName.foo`. However for files that export single functions etc. this doesn't make a lot of sense.

Would we be open to an anonymous version of the syntax for exporting a single function or type for symmetry, or would we remain using return? If so, would we ever be able to return a type that is not namespaced alongside a function that is not namespaced?

alexmccord:
For those, you can still use `return the_thing`, but it does mean an XOR: either you use `return` or you use `export`, not both. It's not too hard to change between them, though. Also seems a bit of a semver hazard if you have to export more than one thing.

docs/export-keyword.md
bradsharp: While we can make this well defined we're going to treat it as a parsing error to avoid the confusion that may come with it. Added this to the RFC.

bradsharp: I've adopted the `local` semantics since they seem like the best place to start, with the option to forward-declare using `const` or `local`. In the future we may consider hoisting functions on the form `<keyword> function <ident>()`

docs/export-keyword.md
bradsharp: Since we're going to build the exports table at the very end this will end up being a parsing error

docs/export-keyword.md
bradsharp: I've updated the RFC to remove the `_EXT` table entirely, in favor of the language building the frozen table at the end and using `const` declarations instead.

docs/export-keyword.md
```diff
@@ -0,0 +1,165 @@
...
+#### Export Shorthand
+
+For convenience, we provide a shorthand form of `export` which will export an existing binding with the same name:
+
+```luau
+const foo = computeFoo()
+export foo
+```
+
+If `foo` was declared as `local` instead of `const` then subequent uses of it here will be treates as if it were declared as `const`.
```
alexmccord:
```suggestion
If `foo` was declared as `local` instead of `const` then subsequent uses of it here will be treated as if it were declared as `const`.
```

docs/export-keyword.md
```diff
@@ -0,0 +1,165 @@
...
+export type Point = { x: number, y: number }
+```
+
+However, this mechanism is currently limited to types. Extending `export` to values and functions would provide a consistent way to expose a stable API from a module, while avoiding the extra boilerplate of returning a table.
+
+## Design
+
+### Syntax
+
+Allow `export` in declarations anywhere `local` is allowed, but only at the top-level of a module. It may prefix a value, function, or type:
```
alexmccord:
```suggestion
Allow `export` in declarations anywhere that defines a name (`local`s, `function`s, and `type`s), but only at the top-level of a module:
```

docs/export-keyword.md
```diff
@@ -0,0 +1,165 @@
...
+end
+
+export type Point = { x: number, y: number }
+```
+
+`export` for values is not permitted inside function bodies or other non–top-level scopes. Doing so results in a parse error.
+
+### Semantics
+
+#### Const By-Default
```
alexmccord, line 37: Alternative framing: this is not necessarily a `const` binding.
```luau
export foo = 7
print(foo)
foo = 8
```
After elaboration, all occurrences of `foo` are a field projection of the export table that you cannot utter in the surface syntax:
```luau
local mod = {}
mod.foo = 7
table.freeze(mod)

print(mod.foo)
mod.foo = 8
```

And since a field projection `foo` inherits the permissions of the projected-from lvalue `mod` which is frozen, `foo` cannot be reassigned. This avoids talking about `const` bindings and everything is still framed around the pre-existing mechanism of frozen tables.

docs/export-keyword.md
```diff
@@ -0,0 +1,165 @@
...
+```luau
+export type Point = { x: number, y: number }
+
+export function distance(a: Point, b: Point)
+	local x, y = a.X - b.X, a.Y - b.Y
+	return math.sqrt(x * x + y * y)
+end
+```
+
+Can be considered sugar for:
```
alexmccord, line 71: Not sure the RFC should show the desugared representation with any constructs that doesn't exist.

docs/export-keyword.md
```diff
@@ -0,0 +1,165 @@
...
+
+function b()
+	return 42
+end
+
+export a
+export b
+```
+
+Exporting before a binding is initialized is an error. Supporting hoisted exports or forward declarations is out of scope for this proposal and may be explored in a future RFC.
```
alexmccord, line 114: > Exporting before a binding is initialized is an error.

Agreed.

> Supporting hoisted exports or forward declarations is out of scope for this proposal and may be explored in a future RFC.

Disagree. This is not backward compatible to retrofit and as such it is not a decision we can defer until later, and mutually dependent functions are fundamental in nontrivial programs. Pre-existing idiom allows mutual dependencies already, so this RFC is more or less a downgrade in its current form. In short, this is a design choice we have to make _now_ or _never_, and deferring a backward-incompatible design choice is the same as choosing to never do it.

If we could resolve names in function bodies _after_ parsing the body itself rather than immediately, then we already support mutual dependencies. This is sort of easy from a conceptual point of view: after parsing the module, all occurrences of `AstExprGlobal` implies a free variable that are yet to be bound. Walk each occurrence and check if an `export` binder name exists and then rebind that occurrence to the `export` binder.

```luau
export type t<a> =
  | { type: "leaf", value: a }
  | { type: "node", children: {t<a>} }

export function walk_node<a, b>(
  node: t<a>,
  leaf_fn: (a) -> b,
  node_fn: ({b}) -> b
): b
  return if node.type == "leaf"
    then leaf_fn(node.value)
    else node_fn(walk_children(node.children, leaf_fn, node_fn))
end

export function walk_children<a, b>(
  children: {t<a>},
  leaf_fn: (a) -> b,
  node_fn: ({b}) -> b
): {b}
  local result = table.create(#children)
  for _, node in children do
    table.insert(result, walk_node(node, leaf_fn, node_fn))
  end
  return result
end
```

As you can see, it is not possible to reorder either `walk_node` or `walk_children`. Initially, when the AST is produced, the `walk_children` at line 12 is an `AstExprGlobal` which is analogous to a free variable in lambda calculus, and during name resolution we find out that we could bind that `walk_children` to the binder `export function walk_children` turning it into a bound variable. Voila, you have mutual dependencies.

This approach will require changes in everything else though, since if `export function walk_children` defines a `local` named `walk_children`, then the reference at line 12 must be an `AstExprLocal`, which lexically occurs _before_ its declaration! There's a bunch of ICEs that will happen and needs to be taken care of. Presumably the bytecode compiler requires more fixes as well if this was added (@vegorov-rbx?).

With that said, I will be contrarian for a minute because there are problems on the other side of this beyond the existing implementation assumptions that needs to be solved if we decided to support mutual dependencies out of the box, because there's a _critical_ caveat to note: we don't want to rebind any `AstExprGlobal` to any `export`s before it was declared except those _inside_ of a function body, and the above approach must continue to maintain this invariant, otherwise you'll end up with nasty bugs in the program like this:

```luau
print(nasty()) -- nil

local x = 5

export function nasty()
  return x
end
```

Which is precisely equivalent to JavaScript's hoisting bug:

```js
console.log(nasty()); // undefined

var x = 5; // change to `let` or `const` for TDZ semantics

function nasty() {
  return x;
}
```

ES6 added `let`/`const` as well as TDZ (short for "temporal dead zone") semantics. Note that we can't make this a compile error because of the below snippet which is "technically well-defined" as per the rule I gave earlier, but is still ill-formed because `local x = 5` has yet to be executed:

```luau
export function hella_nasty()
  return nasty()
end

print(hella_nasty())

local x = 5

export function nasty()
  return x
end
```

Unfortunately it is not possible to turn this class of bugs into a compile error because the analysis is nontrivial, and, per Rice's theorem, is undecidable in the general case (hence why TDZ was punted onto the runtime), so if we did support mutual dependencies, then we have these choices:

1. TDZ semantics (please, for the love of god, no)
2. Don't allow module-level code to call exported functions
   - Honestly, module-level code should seldom have side-effects anyway, sans entry points. They're a bit like `impl` blocks in Rust.
   - That would mean removing this: https://github.com/luau-lang/rfcs/pull/42/changes#diff-444355a7a572290ed4e1533ed7a9de5b36d25acde77f1c7e4d3893558bf2801bR116-R126
3. OCaml-style `and` for explicit grouping of mutual dependencies.
   ```luau
   export function f()
     ... -- g is available here
   end
   and g()
     ... -- f is available here
   end
   ```

docs/export-keyword.md
SPY: It doesn't work if we try to desugar several declarations this way:
```
export a = 2 -- 1
a = 3        -- 2
export b = 4 -- 3
```
If `table.freeze` happens before 2, then 3 should rise an error.
If `table.freeze` happens at the end 2 shouldn't be an error.
It is easer to use `const` semantic here

docs/export-keyword.md
SPY: I don't think we need to introduce hoisting for this purpose.
It brings more harm than convenience.
I would prefer the code snippet above can be expressed with usage of temporary locals.
```luau
export type t<a> =
  | { type: "leaf", value: a }
  | { type: "node", children: {t<a>} }

local walk_children: <a, b> (
  children: {t<a>},
  leaf_fn: (a) -> b,
  node_fn: ({b}) -> b
) -> {b}

export function walk_node<a, b>(
  node: t<a>,
  leaf_fn: (a) -> b,
  node_fn: ({b}) -> b
): b
  return if node.type == "leaf"
    then leaf_fn(node.value)
    else node_fn(walk_children(node.children, leaf_fn, node_fn))
end

walk_children = function<a, b>(
  children: {t<a>},
  leaf_fn: (a) -> b,
  node_fn: ({b}) -> b
): {b}
  local result = table.create(#children)
  for _, node in children do
    table.insert(result, walk_node(node, leaf_fn, node_fn))
  end
  return result
end

export walk_children
```
I also like OCaml style approach. In this case my snippet will be the desugaring of it.

docs/export-keyword.md
SPY: I think it is acceptable as soon as dependency on other RFC explicitly stated and `const` RFC is not rejected yet.
Of course `export` proposal cannot be land in this form until `const` proposal is.

docs/export-keyword.md
alexmccord: Elaboration is not desugaring and is allowed to fail. Consider:

```luau
export a = 1
print(a)
export a = 2
print(a)
```

If we desugar to its `const` form, the above snippet is valid because `const` is allowed to shadow. This is effectively the same as your example. You'd need a rule in this case that rejects any `export` whose name have already been `export`ed. This makes `export`s inexpressible in terms of `const` bindings.

docs/export-keyword.md
```diff
@@ -0,0 +1,165 @@
...
+
+Without this, it would be possible to mutate the variable without changing to the exported table, leading to the following foot-gun:
+
+```luau
+export foo = 5
+
+export function increment()
+	foo += 1 -- 'foo' won't change in the exported table
+end
+```
```
alexmccord, line 54: This is quite awkward. What is the value of `mod.foo` after you call `mod.increment()`?

```luau
local foo = 5

export function increment()
	foo += 1
end

export foo -- ???
```

Currently, `mod.foo` after calling `mod.increment()` would always be `5` because it copies the value of `foo` into the field `foo` of the exported table. Instinctively, I _want_ it to be `6` because, to me, `export foo` looks like it exports an alias to the inner `foo` which is distinct from `export foo = foo` which copies the value of `foo` into the read-only symbol `foo`.

I know that this RFC says `export foo` is sugar for `export foo = foo`, but I'm not so sure about that now, it enables a footgun:

```luau
local foo = 5

export function increment()
  foo += 1
end

export foo

export function get()
  return foo -- this will always be 5
end
```

And we can't even have a linter fire a warning on this (doesn't make sense).

A fix is to be isomorphic to something akin to this, using `__index` to provide aliasing.

```luau
local foo = 5

local function increment()
  foo += 1
end

-- this is an empty table because of `for k, v in mod do` which conflicts with aliasing
return table.freeze(setmetatable({}, {
  __index = function(_, k)
    return if k == "foo" then foo
      else if k == "increment" then increment
      else nil
  end,
}))
```

docs/export-keyword.md
SPY: I would consider that code snipped valid. It outputs 1 and 2 and exports `a` as 2.
I understand it can be confusing for corner cases, but I think this approach makes the language more consistent in design choices.
We could prevent this kind of behavior to disallow constant shadowing on the same level(as JS does), allowing it only in nested scopes. I think it is a good idea in general, but it contradicts Lua(u)'s behavior of shadowing rules for locals. So I would like not to do it to stay consistent.
Also, I consider idiomatic to use "local" intuition for explaining Lua(u)'s syntactic sugar.
For example, `local function f` is traditionally explained as `local f + f = function()` formula. Desugaring `export` with `const` we keep it 'local' in reasoning without bringing global file-level semantic.

docs/export-keyword.md
SPY: I agree this approach would be more flexible and beneficial for some code patterns, but in this case we will loose ability to statically reason about exported values during loading. It will impede some potential optimizations for cross module inlining and constant folding. As soon as Luau doesn't allow to control readonlyness with property level granularity we will need to make some compromises here.
User intention can be expressed using getter function or intermediate object.

alexmccord: I mean, an optimization you can do is literally just move some of the exported symbols into the table part for all exported bindings that have not been mutated (or rather, not mutated through functions). From that you have a guarantee that `increment` always points to the same function object, whereas `foo` still goes through `__index`.

The reason why I didn't do that was because of `for k, v in mod do` would not iterate over `foo`. But here, let's just do that. The downside is that now `pairs(mod)`/`rawget(mod, "increment")` is weird.

```luau
local order = {"foo", "increment"} -- sorted by hash, just like hashmaps
return table.freeze(setmetatable({ increment = increment }, {
  __index = function(_, k)
    return if k == "foo" then foo
      else nil
  end,
  __iter = function(self)
    local i = 0
    return function()
      i += 1
      local k = order[i]
      return k, self[k]
    end
  end,
}))
```

Problem solved, the compiler or VM can still see that `increment` field exists in the table and did not require `__index`.

Module systems are complicated for a reason. You can't hand-wave all the issues away in an RFC with 165 lines that barely even talks about edge cases.

MagmaBurnsV: I say if you plan on reintroducing assignability through shorthand exports, don't even bother making exported values unassignable in the first place. I'm actually leaning more in the camp of not even supporting shorthand exports because of this issue.

In my personal opinion, exporting should just be pure syntax sugar that maps exactly to returning a plain frozen table. There should be no special metatable mechanics like the proposed aliasing rules in the desugared form.

docs/export-keyword.md
```diff
@@ -0,0 +1,165 @@
...
+export version = "5.1"
+
+export function init()
+	-- do a thing
+end
+
+export type Point = { x: number, y: number }
+```
+
+`export` for values is not permitted inside function bodies or other non–top-level scopes. Doing so results in a parse error.
```
MagmaBurnsV, line 33: I believe you should still be able to export values from within `do end` blocks, as this could allow for neat namespacing with lexical scoping:

```luau
do
	local count = 0
	
	export function increment(): number
		count += 1
		return count
	end
end

-- count and increment not visible here
```

MagmaBurnsV:
I've been thinking about possible alternative semantics, and this is what I've come up with: what if we treat exporting values as like assigning keys to the export table and allow exported values to be reassigned until the end of the module when the export table is frozen and returned.

This mirrors the way people currently write module exports, and it solves all the problems with mutually dependent functions, aliased shorthand exports (you don't need shorthand syntax anymore), and still allows for lexical scoping rules:

```luau
export a = 1
a = 2

-- becomes

local _EXP = {}
_EXP.a = 1
_EXP.a = 2

return table.freeze(_EXP)
```

```luau
-- like export f, g = nil, nil
export f, g

function f()
	-- calling g() before it is assigned
	-- throws an "attempt to call a nil value" error
	g()
end

function g()
	f()
end

-- becomes

local _EXP = {}
_EXP.f = nil
_EXP.g = nil

function _EXP.f()
	_EXP.g()
end

function _EXP.g()
	_EXP.f()
end

return table.freeze(_EXP)
```

```luau
export foo = 5

export function increment()
	-- this works until the module ends and freezes the export table
	-- which subsequently throws an "attempt to modify a readonly table" error
	foo += 1
end

export function get()
  return foo
end

-- becomes

local _EXP = {}
_EXP.foo = 5

function _EXP.increment()
	_EXP.foo += 1
end

function _EXP.get()
	return _EXP.foo
end

return table.freeze(_EXP)
```

```luau
do
	local count = 0
	
	export function increment(): number
		count += 1
		return count
	end
	
	-- count and increment only visible in here
end

-- becomes

local _EXP = {}

do
	local count = 0
	
	function _EXP.increment(): number
		count += 1
		return count
	end
end

return table.freeze(_EXP)
```

There are no extra features like `const` required to make this work, and I would imagine it would also still be possible to unlock cross-module optimization opportunities this way by inlining and folding exports that are never reassigned.

docs/export-keyword.md
alexmccord: Yeah, I think just remove the shorthand `export`. This RFC is proposing CommonJS-esque modules, not anything fancy like ES modules which has live bindings.

docs/export-keyword.md
alexmccord: This is what local modules does, if we had some sort of `module` keyword.

```luau
module M is
  local count = 0

  export function increment(): number
    count += 1
    return count
  end
end

export increment = M.increment
```

Which is equivalent to a closure, modulo (heh) namespacing/qualified paths/etc.

```luau
local function makeM(): () -> number
  local count = 0

  return function() -- ignoring the _actual_ exporting machinery as red herring
    count += 1
    return count
  end
end

export increment = makeM()
```

You could even instantiate it multiple times. Module functors.

```luau
module Make(n: number)
  local count = 0
  
  export function next(): number
    count += n
    return count
  end
end

module Counter = Make(1)

export increment = Counter.next
```

```luau
export incrementX = makeM() -- has its own independent `count` state
export incrementY = makeM() -- and its own independent `count` state
```

alexmccord:
One optimization that I noticed this enables as a side effect is lambda dropping. You can replace a function whose body branches on a const with an if expression that branches on the same condition and maps to its original body, e.g.

```luau
local platform = require("@std/system/platform")

local win32 = platform.win32

export function f()
  if win32 then
    -- Win32-specific logic
  else
    -- POSIX-specific logic
  end
end
```

Instead of paying the cost to branch per function call, the compiler performs a lambda drop on the `win32` branch by factoring it outside the function and pay the cost to branch when the module runs for the first time: <sup><sup>(obviously you'd share the debugname and source locations)</sup></sup>

```luau
local platform = require("@std/system/platform")

local win32 = platform.win32

export f = if win32
  then function() --[[ Win32-specific logic ]] end
  else function() --[[ POSIX-specific logic ]] end
```

This can be done today on its own, but not across module boundaries, and that's what is actually interesting here.

jackhexed:
@alexmccord This optimization can be done today with any function binding that isn't reassigned to. Nothing about export enables it.

alexmccord:
> @alexmccord This optimization can be done today with any function binding that isn't reassigned to. Nothing about export enables it.

Look at the last line in my comment at the bottom. I already said that (badly, I guess).

jackhexed:
I seem to be illiterate lol

jackhexed:
Although this isn't an optimization done across module boundaries, unless you mean the folding of constants from outside the module.

docs/export-keyword.md
```diff
@@ -0,0 +1,165 @@
...
+
+A module that uses `export` for values or functions may not also return a custom value. Attempting to do so is an error:
+
+```luau
+export function distance(a: Point, b: Point)
+	return math.sqrt((a.X - b.X)^2)
+end
+
+return { distance = distance } -- error
+```
```
alexmccord, line 152: What about this?

```luau
export function distance(a: Point, b: Point)
  return math.sqrt((a.X - b.X)^2)
end

do -- or `for i = 1, 1 do`, or `repeat ... until false`, or `while true do ... end`, or any other variation thereof
  return { distance = distance } -- ???
end
```

It's obviously nonsense, but `do` starts a new block. Ditto this:

```luau
do -- or `for i = 1, 1 do`, or `repeat ... until false`, or `while true do ... end`, or any other variation thereof
  local m = {}
    
  function m.distance(a: Point, b: Point)
    return math.sqrt((a.X - b.X)^2)
  end

  return m
end

export function distance(a: Point, b: Point)
  return math.sqrt((a.X - b.X)^2)
end
```

The first one can be rejected at parse time, the second one can also be rejected at parse time, but it means you need to track if `return` was encountered at top-level _and_ in any number of nested _non-function_ blocks, and `export` was encountered at top-level, and then error if both flags are true. This is a nonlocal syntactic restriction.

hgoldstein:
Thanks for all of the discussion here folks, closing this out given that https://github.com/luau-lang/rfcs/pull/179 has been approved and merged in.

PR #48, opened 2024-07-10
RFC: Deprecated Attribute for Functions

aviralg:
[Rendered](https://github.com/aviralg/rfcs/blob/syntax-attribute-functions-deprecated/docs/syntax-attribute-functions-deprecated.md)

### DIFF
--- a/docs/syntax-attribute-functions-deprecated.md
+++ b/docs/syntax-attribute-functions-deprecated.md
@@ -24,7 +24,7 @@ The `@deprecated` attribute can take upto two named string parameters. The param
 | `@[deprecated {use = string}]`                  | `"Function 'foo' is deprecated. Use '<use>' instead."`                 |
 | `@[deprecated {reason = string, use = string}]` | `"Function 'foo' is deprecated. Use '<use>' instead. <reason>`         |
 
-The `@deprecated` attribute can only be used on named functions. It does not apply recursively to the functions defined within the lexical scope of the attributed function.
+The `@deprecated` attribute can only be used on named functions and methods. It does not apply recursively to the functions defined within the lexical scope of the attributed function.
 
 ## Related Work
 
--- a/docs/syntax-attribute-functions-deprecated.md
+++ b/docs/syntax-attribute-functions-deprecated.md
@@ -29,8 +29,10 @@ If `foo` is a member of class `bar`, the warning messages above will start as `"
 Consider the following example. Function `printTable` is marked deprecated.
 
 ```lua
-@[deprecated {use = "table.foreach(table, print)", 
-              reason = "`printTable` forces tables to be printed in one style; `table.foreach` allows configurable printing."}]
+@[deprecated {
+    use = "table.foreach(table, print)", 
+    reason = "`printTable` forces tables to be printed in one style; `table.foreach` allows configurable printing.",
+}]
 local function printTable(t)
     for k,v in pairs(t) do
         print(k, v)
--- a/docs/syntax-attribute-functions-deprecated.md
+++ b/docs/syntax-attribute-functions-deprecated.md
@@ -1,0 +1,66 @@
+# Deprecated Attribute for Functions
+
+## Summary
+
+This RFC proposes a `@deprecated` attribute to annotate Luau functions for deprecation. Using these functions should result in a deprecation warning from the linter. The Luau LSP should also indicate the deprecated status of these functions in its completion list. 
+
+Using these functions should result in deprecation warnings from the linter. In Studio, 
+
+## Motivation
+
+Luau should allow a way for developers to evolve APIs and warn clients if some API will be removed in the future. Code using these deprecated APIs should be rewritten according to the solution proposed in the deprecation message.
+
+Roblox already has numerous deprecated Luau APIs. LSP strikes out the names of these APIs in the autocompletion popup in Roblox Studio, their use issues deprecation warnings, and their deprecation status is reflected in Roblox's Creator Hub documentation. Unfortunately, Luau lacks support to leverage this tooling for deprecating custom creator-defined functions. Even for core Luau deprecated functions, `getfenv` and `setfenv`, the Linter has hardcoded name comparisons to identify and report deprecation messages. Deprecating new APIs requires ad-hoc modifications to the tooling which can only be performed by the language designers.
+
+This RFC proposes the introduction of a `@deprecated` attribute to mark deprecated Luau APIs. This has multiple benefits:
+1. It provides an explicit syntactic marker to identify deprecated functions in Luau code.
+2. The attribute can be read by LSP and the corresponding API can be displayed in a visually distinct style in the autocompletion popup.
+3. Linter can issue warning messages when the deprecated function is used.
+
+## Design
+
+The `@deprecated` attribute can take upto two named parameters. The parameters affect the warning message issued by the tooling. The `msg` parameter is the warning message issued by the tooling when it identifies a use of the deprecated function The `alt` parameter is the name of the function that should be used in lieu of the deprecated function. The following table shows the messages issued by the different styles of this attribute when used on a function named `foo`.
+
+| Style                                  | Message                                                                |
+| -------------------------------------- | -----------------------------------------------------------------------|
+| `@deprecated`                          | `"Function 'foo' is deprecated."`                                      |
+| `@[deprecated {msg = msg}]`            | `"Function 'foo' is deprecated. <msg>"`                                |
+| `@[deprecated {alt = alt}]`            | `"Function 'foo' is deprecated. Use '<alt>' instead."`                 |
+| `@[deprecated {msg = msg, alt = alt}]` | `"Function 'foo' is deprecated. <msg> Use '<alt>' instead.`            |
+
+The `@deprecated` attribute can only be used on top-level functions. It does not apply recursively to the functions defined within the lexical scope of the attributed function.
+
+## Related Work
+
+Many mainstream languages provide a similar mechanism for deprecating language features. They differ in the kind of program elements that can be deprecated, the ability to convert deprecation warnings to errors, and the ability to specify the version which deprecated the API. 
+
+1. **C++** since C++14 provides a `[[deprecated(msg)]]` attribute that can be applied to declaration of a class, a typedef-name, a variable, a nonstatic data member, a function, a namespace, an enumeration, an enumerator, or a template specialization. The compiler emits a deprecation message, `msg`, when the deprecated program element is used.
+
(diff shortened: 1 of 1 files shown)


docs/syntax-attribute-functions-deprecated.md
```diff
@@ -0,0 +1,60 @@
...
+
+| Style                                  | Message                                                                |
+| -------------------------------------- | -----------------------------------------------------------------------|
+| `@deprecated`                          | `"Function 'foo' is deprecated."`                                      |
+| `@[deprecated {msg = msg}]`            | `"Function 'foo' is deprecated. <msg>"`                                |
+| `@[deprecated {alt = alt}]`            | `"Function 'foo' is deprecated. Use '<alt>' instead."`                 |
+| `@[deprecated {msg = msg, alt = alt}]` | `"Function 'foo' is deprecated. <msg> Use '<alt>' instead.`            |
```
vegorov-rbx, line 25: Shouldn't `msg` and `alt` values be strings?

docs/syntax-attribute-functions-deprecated.md
```diff
@@ -0,0 +1,60 @@
...
+
+The `@deprecated` attribute can only be used on top-level functions. It does not apply recursively to the functions defined within the lexical scope of the attributed function.
```
vegorov-rbx, line 27: Do not see the need to restrict to top level.

aviralg: Conceptually, the analysis feels simpler (tractable?) when you restrict to top-level functions. Also, I would be surprised if inner functions are used as the public API of a module. Having said this, I would not mind generalizing it to functions at any nesting depth.

docs/syntax-attribute-functions-deprecated.md
aviralg: Yes, I meant `msg` and `alt` to be placeholders for this string values. I will replace them with string to avoid confusion.

aviralg:
@alexmccord suggested renaming `msg` to `reason` and `alt` to `use`.

docs/syntax-attribute-functions-deprecated.md
vegorov-rbx: Yeah, most confusing is 'alt' since it could be taken as a symbolic reference of the replacement function.

docs/syntax-attribute-functions-deprecated.md
andyfriesen: Deprecation status is already a field on `Luau::FunctionType`, so it doesn't actually affect the implementation effort.

docs/syntax-attribute-functions-deprecated.md
```diff
@@ -6,25 +6,25 @@ This RFC proposes a `@deprecated` attribute to annotate Luau functions for depre
...
-| `@[deprecated {msg = msg}]`            | `"Function 'foo' is deprecated. <msg>"`                                |
-| `@[deprecated {alt = alt}]`            | `"Function 'foo' is deprecated. Use '<alt>' instead."`                 |
-| `@[deprecated {msg = msg, alt = alt}]` | `"Function 'foo' is deprecated. <msg> Use '<alt>' instead.`            |
+| Style                                           | Message                                                                |
+| ----------------------------------------------- | -----------------------------------------------------------------------|
+| `@deprecated`                                   | `"Function 'foo' is deprecated."`                                      |
+| `@[deprecated {reason = string}]`               | `"Function 'foo' is deprecated. <reason>"`                             |
+| `@[deprecated {use = string}]`                  | `"Function 'foo' is deprecated. Use '<use>' instead."`                 |
+| `@[deprecated {reason = string, use = string}]` | `"Function 'foo' is deprecated. Use '<use>' instead. <reason>`         |
```
alexmccord, line 25:
```suggestion
| `@[deprecated {reason = string, use = string}]` | `"Function 'foo' is deprecated. Use '<use>' instead. <reason>"`        |
```

docs/syntax-attribute-functions-deprecated.md
```diff
@@ -0,0 +1,60 @@
...
+
+This RFC proposes the introduction of a `@deprecated` attribute to mark deprecated Luau APIs. This has multiple benefits:
+1. It provides an explicit syntactic marker to identify deprecated functions in Luau code.
+2. The LSP interface can display the deprecated API in a visually distinct style in the autocompletion popup.
+3. The linter can issue a warning message when the deprecated function is used.
+
+## Design
+
+The `@deprecated` attribute can take upto two named string parameters. The parameters affect the warning message issued by the linter. The `reason` parameter gives the reason for deprecation and the `use` parameter gives the name of the function that should be used in lieu of the deprecated function. The following table shows the warning message issued by the different styles of this attribute when used on a function named `foo`.
```
alexmccord, line 18:
```suggestion
The `@deprecated` attribute can take up to two named string parameters. The parameters affect the warning message issued by the linter. The `reason` parameter gives the reason for deprecation and the `use` parameter gives the name of the function that should be used in lieu of the deprecated function. The following table shows the warning message issued by the different styles of this attribute when used on a function named `foo`.
```

vegorov-rbx:
I still wish for at least one real source code example.

docs/syntax-attribute-functions-deprecated.md
aviralg: Fixed

aviralg:
> I still wish for at least one real source code example.

I thought the proposal was simple enough that an example was not really needed. Added one now. Its pretty basic but I think it illustrates the idea.

vegorov-rbx:
> Added one now. Its pretty basic but I think it illustrates the idea.

Thanks.

> I thought the proposal was simple enough that an example was not really needed. 

It's just that is a new user copy-pasted the examples from style table, three out of four of them would have resulted in an error (since strings are not quoted).
With a code example it's much easier to start!

docs/syntax-attribute-functions-deprecated.md
```diff
@@ -0,0 +1,80 @@
...
+
+| Style                                           | Message                                                                |
+| ----------------------------------------------- | -----------------------------------------------------------------------|
+| `@deprecated`                                   | `"Function 'foo' is deprecated."`                                      |
+| `@[deprecated {reason = string}]`               | `"Function 'foo' is deprecated. <reason>"`                             |
+| `@[deprecated {use = string}]`                  | `"Function 'foo' is deprecated, use '<use>' instead."`                 |
+| `@[deprecated {reason = string, use = string}]` | `"Function 'foo' is deprecated, use '<use>' instead. <reason>"`         |
```
alexmccord, line 25:
```suggestion
| Style                                                 | Message                                                                |
| ----------------------------------------------------- | -----------------------------------------------------------------------|
| `@deprecated`                                         | `"Function 'foo' is deprecated."`                                      |
| `@[deprecated {reason = "my reason"}]`                | `"Function 'foo' is deprecated. <reason>"`                             |
| `@[deprecated {use = "bar()"}]`                       | `"Function 'foo' is deprecated, use '<use>' instead."`                 |
| `@[deprecated {reason = "my reason", use = "bar()"}]` | `"Function 'foo' is deprecated, use '<use>' instead. <reason>"`        |
```

docs/syntax-attribute-functions-deprecated.md
```diff
@@ -0,0 +1,80 @@
...
+
+If `foo` is a member of class `bar`, the warning messages above will start as `"Member 'bar.foo' is deprecated ...`.
+
+Consider the following example. Function `printTable` is marked deprecated.
+
+```lua
+@[deprecated {use = "table.foreach(table, print)", 
+              reason = "`printTable` forces tables to be printed in one style; `table.foreach` allows configurable printing."}]
```
alexmccord, line 33:
```suggestion
@[deprecated {
    use = "table.foreach(table, print)", 
    reason = "`printTable` forces tables to be printed in one style; `table.foreach` allows configurable printing.",
}]
```

alexmccord (commented):
Good point. Doing this is a bit better too.

alexmccord:
Also, table and class properties do have support for deprecated members.

```luau
local t = {}

@deprecated
function t:m() end -- it's the property `m` that's deprecated, not the `FunctionType` itself.
```

docs/syntax-attribute-functions-deprecated.md
```diff
@@ -0,0 +1,91 @@
...
+
+Many mainstream languages provide a similar mechanism for deprecating language features. They differ in the kind of program elements that can be deprecated, the ability to convert deprecation warnings to errors, and the ability to specify the version which deprecated the API. 
+
+1. **C++** since C++14 provides a `[[deprecated(msg)]]` attribute that can be applied to declaration of a class, a typedef-name, a variable, a nonstatic data member, a function, a namespace, an enumeration, an enumerator, or a template specialization. The compiler emits a deprecation message, `msg`, when the deprecated program element is used.
+
+2. **C#** provides a `[ObsoleteAttribute(msg, error)]` which can be applied to all program elements except assemblies, modules, parameters, and return values. The `msg` string is emitted by the compiler when the deprecated program element is used in code. The `error` flag converts the warning to an error.
+
+3. **Java** provides a `@Deprecated` annotation to deprecate a class, method, or field. The Java compiler issues warnings when these program elements are used. A detailed deprecation message can be supplied by a correspoding javadoc `@deprecated` tag.
```
vegorov-rbx, line 68:
```suggestion
3. **Java** provides a `@Deprecated` annotation to deprecate a class, method, or field. The Java compiler issues warnings when these program elements are used. A detailed deprecation message can be supplied by a corresponding javadoc `@deprecated` tag.
```

docs/syntax-attribute-functions-deprecated.md
```diff
@@ -0,0 +1,91 @@
...
+
+The description above shows the full-form of the deprecation annotation. In most cases, it is valid to leave the message parameter in favor of a generic deprecation message. Compared to most of these languages, the design proposed for deprecating Luau APIs is simpler. 
+1. Deprecation is not supported for non-function program elements because Luau only supports attributes on functions. 
+2. There is no way to convert deprecation warnings to errors. 
+3. There is no explicit attribute parameter for specifying version string. They can be accommodated in the `reason` parameter. If desired, a future RFC could introduce a `version` parameter to the `@deprecated` attribute.
+
+## Drawbacks
+
+Adding this attribute increases complexity of code. Once the attribute is released, we cannnot deprecate it.
```
vegorov-rbx, line 87:
```suggestion
Adding this attribute increases complexity of code. Once the attribute is released, we cannot deprecate it.
```

vegorov-rbx (approved):

PR #49, opened 2024-07-11
RFC: No Support for Function Inlining

dibrinsofor:
This RFC addresses the idea of an `@inline` keyword in Luau. It addresses common reasons where inlining would be typically be applicable but would not apply to Luau due to metamethod dynamic dispatching, further suggesting that this feature will not be introduced to the language.

[rendered](https://github.com/dibrinsofor/rfcs/blob/func-inline/docs/function-inlining.md)

### DIFF
--- a/docs/function-inlining.md
+++ b/docs/function-inlining.md
@@ -98,19 +98,6 @@ end
 ```
 In this example, we declare `Vec3` with an inlined `__add` (`+`) operator overload. To make this inlining work, we would need to do one of two things, (1) confirm the `rhs` and `lhs` share the same shape (i.e, they both have`x`, `y` and `z`) or (2) ensure the `rhs` has a similar operator overload. Due to dynamic dispatch, we cannot make either guarantee making this case unsuitable for overloads.
 
-
-4. Higher Order Functions
-```luau
-@inline
-local function add(x, y)
-  return x + y
-end
-
-local function accumulate(acc, i, lst, f)
-  return if i > #lst then acc else accumulate(f(acc, lst[i]), i + 1, lst, f)
-end
-```
-
 ## Design
 None (yet).
 
--- a/docs/function-inlining.md
+++ b/docs/function-inlining.md
@@ -6,9 +6,9 @@ Luau's bytecode compiler is smart about inlining functions. We would rather focu
 
 ## Motivation
 
-**context**: function inlining replaces function calls with the function's definition to reduce overhead costs.
+**context**: function inlining replaces function calls with the function's implementation to reduce overhead costs.
 
-Whether the goal is to improve instruction locality or perform context specific optimizations, <!-- force the compiler to optimize the function body in specific contexts --> function inlining can be a great resource for developers. At `-02` or higher, the compiler is able to automatically decide if functions ought to be inlined. In the currrent state of the language, exposing this to users directly could do more harm than good. Not only do these use cases only cater to a tiny subset of the language's users, it exposes our bytecode compiler to FILL when the larger subset who may not have a working understanding of the compiler adopts it. For Luau, this would happen at around the 5000th recursive call.
+Whether the goal is to improve instruction locality or perform context specific optimizations, <!-- force the compiler to optimize the function body in specific contexts --> function inlining can be a great resource for developers. At `-02` or higher, the compiler is able to automatically decide if functions ought to be inlined. In the currrent state of the language, exposing this to users directly could do more harm than good. Not only do these use cases only cater to a tiny subset of the language's users, it exposes our bytecode compiler to stack overflow errors when the larger subset who may not have a working understanding of the compiler adopts it. For Luau, this would happen at around the 5000th recursive call.
 
 ```luau
 local function sum(lst: {number})
@@ -23,10 +23,9 @@ end
 local s = sum({1, 2, 3})
 ```
 
-Consider this "inlined" recursive call, it attempts to sum up numbers in a list but it never terminates as `sum` is recursively called on `lst` which does not shrink. If Luau attempted to inline such a call, it would unroll this and one or two things could happen. If we did have an unroll limit then we would attempt our unrolling up until that point and end up making the program larger than it needs to be or thrashing our stack. The stack will grow beyond the system's stack limit and could potentially eat up resources meant for other processes.
-<!-- one or two also sounds funny, change -->
+Consider this "inlined" recursive call, it attempts to sum up numbers in a list but it never terminates as `sum` is recursively called on `lst` which does not shrink. If Luau attempted to inline such a call, it would unroll this up until the 5000th recursive call and end up thrashing our stack. The stack will essentially grow beyond the system's stack limit and could potentially eat up resources meant for other processes.
 
-More specifically, due to limited number of cases where the compiler supports inlining, exposing this feature would not be entirely useful. This section further explores common cases were `@inline` would not apply.
+More specifically, due to limited number of cases where the compiler can even consider inlining, exposing this feature would not be entirely useful. This section further explores common developer patterns that may adopt `@inline` in ways the compiler cannot handle.
 
 1. Exported a `table`
 ```luau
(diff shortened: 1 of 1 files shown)


docs/function-inlining.md
```diff
@@ -0,0 +1,40 @@
...
+
+Whether the goal is to improve instruction locality or perform context specific optimizations, <!-- force the compiler to optimize the function body in specific contexts --> function inlining can be a great resource for developers. Thankfully with the `-02` switch, Luau is able to automatically decide if code deserves to be inlined and optimize the code accordingly. To the best of our knowledge, exposing this to users directly could do more harm than good. Not only do these use cases only cater to a tiny subset of the language's users, it exposes our bytecode compiler to FILL when the larger subset who may not have a working understanding of the compiler adopts it.
+
+```luau
+@inline -- does not exist
+local function sum(lst: {[number]})
+    if #lst == 0 then
+        return 0
+    else
+        return lst[1] + sum(lst) -- bad rec
```
alexmccord, line 19:
```suggestion
        return lst[1] + sum(lst) -- bad rec
    end
```

docs/function-inlining.md
```diff
@@ -0,0 +1,40 @@
+# No support for user inlining (yet)
+
+## Summary
+<!-- https://luau-lang.org/performance#function-inlining-and-loop-unrolling -->
+Luau's bytecode compiler (or VM?) is smart about inlining functions. We would rather focus on improving this automation than risk the chance of its misuse.
```
alexmccord, line 5:
```suggestion
Luau's bytecode compiler is smart about inlining functions. We would rather focus on improving this automation than risk the chance of its misuse.
```

docs/function-inlining.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Motivation
+
+**context**: function inlining replaces function calls with the function's implementation to reduce overhead costs.
+
+Whether the goal is to improve instruction locality or perform context specific optimizations, <!-- force the compiler to optimize the function body in specific contexts --> function inlining can be a great resource for developers. Thankfully with the `-02` switch, Luau is able to automatically decide if code deserves to be inlined and optimize the code accordingly. To the best of our knowledge, exposing this to users directly could do more harm than good. Not only do these use cases only cater to a tiny subset of the language's users, it exposes our bytecode compiler to FILL when the larger subset who may not have a working understanding of the compiler adopts it.
+
+```luau
+@inline -- does not exist
```
alexmccord, line 14: True, it doesn't exist, but let's clip it out.
```suggestion
@inline
```

docs/function-inlining.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Summary
+<!-- https://luau-lang.org/performance#function-inlining-and-loop-unrolling -->
+Luau's bytecode compiler (or VM?) is smart about inlining functions. We would rather focus on improving this automation than risk the chance of its misuse.
+
+## Motivation
+
+**context**: function inlining replaces function calls with the function's implementation to reduce overhead costs.
+
+Whether the goal is to improve instruction locality or perform context specific optimizations, <!-- force the compiler to optimize the function body in specific contexts --> function inlining can be a great resource for developers. Thankfully with the `-02` switch, Luau is able to automatically decide if code deserves to be inlined and optimize the code accordingly. To the best of our knowledge, exposing this to users directly could do more harm than good. Not only do these use cases only cater to a tiny subset of the language's users, it exposes our bytecode compiler to FILL when the larger subset who may not have a working understanding of the compiler adopts it.
```
alexmccord, line 11: I'm not sure I understand "it exposes our bytecode compiler to FILL when the larger subset who may not have a working understanding of the compiler adopts it."

docs/function-inlining.md
```diff
@@ -0,0 +1,40 @@
...
+    if #lst == 0 then
+        return 0
+    else
+        return lst[1] + sum(lst) -- bad rec
+end
+
+local s = sum({1, 2, 3})
+```
+
+Consider this "inlined" recursive call, it attempts to sum up numbers in a list but it never terminates as `sum` is recursively called on `lst` which does not shrink. If Luau attempted to inline such a call, it would unroll this and one or two things could happen. If we did have an unroll limit then we would attempt our unrolling up until that point and end up making the program larger than it needs to be or thrashing our stack. The stack will grow beyond the system's stack limit and could potentially eat up resources meant for other processes.
```
alexmccord, line 25: This sounds like this applies if we did TCO (I dare not call it optimization because it really isn't, but alas), but we actually don't do this, so it'll stack overflow by 5000th recursive calls, iirc.

docs/function-inlining.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Design
+None (yet).
+<!-- additionally, we could just take the inline hints and ignore if it does not meet some criteria (unroll limit) -->
+
+## Drawbacks
+None (yet).
+
+## Alternatives
+<!-- copied from: https://roblox.atlassian.net/wiki/spaces/Client/pages/1542170530/VM+Roadmap+2021#VMRoadmap2021-Debugger -->
```
vegorov-rbx, line 38: Let's not leak internal resources in public GitHub.

docs/function-inlining.md
```diff
@@ -0,0 +1,40 @@
...
+Prior to now, the team has considered some designs to make function inlining work effectively in Luau and is still hopeful that this may be possible sometime in the future. Till then, we suggest opening github issues with code samples of legitimate cases that we can use to further improve Luau's automatic (function inlining and unrolling) optimizations.
```
vegorov-rbx, line 39: What do you mean 'hopeful and possible'? Automatic function inlining is already implement and is efficient.
Don't see a reason for this sentence to exist.

docs/function-inlining.md
```diff
@@ -0,0 +1,40 @@
+# No support for user inlining (yet)
+
+## Summary
+<!-- https://luau-lang.org/performance#function-inlining-and-loop-unrolling -->
+Luau's bytecode compiler (or VM?) is smart about inlining functions. We would rather focus on improving this automation than risk the chance of its misuse.
+
+## Motivation
```
alexmccord, line 7: I would like to see more cases where `@inline` doesn't apply:

1. Exported without a `table.freeze`.
2. "Object oriented" code.
3. Metamethod.

So 1 would look something like:

```luau
local mod = {}

function mod.sum(lst: {number})
  return mod.sumAcc(0, 1, lst)
end

function mod.sumAcc(acc: number, i: number, lst: {number})
  return if i > #lst then acc else mod.sumAcc(acc + lst[i], i + 1, lst)
end
```

docs/function-inlining.md
dibrinsofor: "FILL" is a placeholder

docs/function-inlining.md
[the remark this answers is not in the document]
dibrinsofor: should I still leave in a bad rec comment?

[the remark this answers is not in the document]
alexmccord: Nah. Let's clip it too. You can mention that recursion and inlining won't inline it or cause us to compile it into a loop.

docs/function-inlining.md
alexmccord: Oh, for pt 1, I want to mention that even if you added `table.freeze`, there's no guarantee that the property pointing to the function implementation did not get transmuted to point to a different function _before_ the table got frozen.

There might be a simpler example, but first one that comes to mind:

```luau
-- A.luau
local mod = {}

function mod.print_foo(t)
  print(t.foo)
end

return mod

-- B.luau
local A = require("./A")

local mod = {}

function mod.foo()
  -- `mod.foo` could theoretically be called without our knowledge
  -- which in turn is passed into `A.foo`, which we can't see the implementation of
  -- and so we must assume it mutates `mod`, making `mod.foo` not inlineable.
  A.foo(mod)
end

return table.freeze(mod)
```

docs/function-inlining.md
```diff
@@ -0,0 +1,109 @@
...
+@inline
+function Vec3:__add(rhs)
+  return Vec3.new(self.x + rhs.x, self.y + rhs.y, self.z + rhs.z)
+end
+```
+In this example, we declare `Vec3` with an inlined `__add` (`+`) operator overload. To make this inlining work, we would need to do one of two things, (1) confirm the `rhs` and `lhs` share the same shape (i.e, they both have`x`, `y` and `z`) or (2) ensure the `rhs` has a similar operator overload. Due to dynamic dispatch, we cannot make either guarantee making this case unsuitable for overloads.
+
+## Design
+None (yet).
+
```
alexmccord, line 103: Can clip this one, there's no design in an RFC that rationalizes against adding a feature. :)
```suggestion
```

docs/function-inlining.md
```diff
@@ -0,0 +1,109 @@
...
+## Drawbacks
+None (yet).
```
alexmccord, line 105: A drawback we could talk about is we are not giving the developers any control about raising the maximum cost of the attribute, but in exchange we don't have to deal with people who think `@inline` should _always_ inline.

docs/function-inlining.md
```diff
@@ -0,0 +1,97 @@
...
+
+@inline
+local function sumAcc(acc: number, i: number, lst: {number})
+  return if i > #lst then acc else sumAcc(acc + lst[i], i + 1, lst)
+end
+
+local s = sum({1, 2, 3})
+```
+
+Consider this "inlined" recursive call, it attempts to sum up numbers in a list but it never terminates as `sum` is recursively called on `lst` which does not shrink. If Luau attempted to inline such a call, it would unroll this up until the 5000th recursive call and end up thrashing our stack. The stack will essentially grow beyond the system's stack limit and could potentially eat up resources meant for other processes.
```
alexmccord, line 26: So, this function does terminate. If `i` is greater than the length of `lst`, return the value we've added up thus far. `i` is incremented on each call. As long as `sumAcc` was able to finish in fewer call frames than the available call stack, the function will successfully return the value.

Also, not really too sure about how inlining leads to unrolling, but unrolling shouldn't cause a stack overflow. Unrolling refers to an optimization that takes a loop that's statically known to only occur exactly N times every time, and just copy paste the body N amount of times.

```suggestion
Consider this "inlined" recursive call. Inlining a recursive function would mean either the bytecode size is infinite, or the generated code is effectively the same as a loop a la tail call optimization.
```

alexmccord:
If these feedback are resolved, this RFC will be in a great shape.

docs/function-inlining.md
dibrinsofor: do we not want to explain what the code does in detail?

alexmccord: Nah. I leave that as an exercise left to the reader. I want the non-code part of the RFC to focus on the _why_ the `inline` attribute can't work, not necessarily _what_ the code does.

alexmccord (approved):

aatxe (approved):

PR #50, opened 2024-07-23
RFC: Constraints on generic parameters

SecondNewtonLaw:
This RFC is a possible implementation on what could be an improvement to the current generics system, by allowing generic types to be constrained, just as in any other language's implementation of generics, which would allow for generics to be used in other seemingly unrelated places, like more efficient native code generation.

[rendered](https://github.com/SecondNewtonLaw/rfcs/blob/master/docs/type-constraints-on-generic-parameters.md)

### DIFF
--- a/docs/type-constraints-on-generic-parameters.md
+++ b/docs/type-constraints-on-generic-parameters.md
@@ -9,66 +9,170 @@ Modify the current generics implementation to allow generic types to be constrai
 When writing generic functions currently, it is as if you were simply using any unless you make your own type, being that the generic type has no constraints.
 
 Take this snippet for example:
+
 ```luau
+type Entity = {
+    -- ...
+}
+
 type Gun = {
     IsEmpty: (self: Gun) -> boolean,
     FireGun: (self: Gun) -> (),
     -- ...
 }
 
-local function fireGun<T>(self: T)
-    if not self:IsEmpty() then -- Type Error: Type 'T' does not have key 'IsEmpty'.
-        self:FireGun()         -- Type Error: Type 'T' does not have key 'FireGun'.
+type WeaponResult = {
+    HittedEntity: Entity,
+    DamageDealt: number,
+}
+
+type ShotgunWeaponResult =  {
+    PelletCount: number,
+    AmountOfPelletsHit: number,
+} & WeaponResult
+
+type Shotgun = {
+    -- ...
+} & Gun
+
+local function fireGun<T: Gun, U: WeaponResult>(self: T): U?
+    if not self:IsEmpty() then        -- No error
+        return self:FireGun()         -- No error
     end
+    
+    return nil    
 end 
 
 local gun: Gun
+local shotgun: Shotgun
 
-fireGun(gun)             -- Works as "intended".
-fireGun("Hello, world!") -- Will error you out, but it will compile as if it were valid.
+local fireResult = fireGun(gun) :: WeaponResult                 -- No error
+local shotgunFireResult = fireGun(shotgun) :: ShotgunWeaponResult             -- No error
 
+fireGun("Hello, world!")     -- Type Error: type 'string' and 'Gun' are not related to one another.
 ```
-On this snippet, we are trying to implement a theoretical Gun system, we may have many implementations for the Gun, like a Shotgun or a Pistol, and we may want to use generics to implement them, however, the current system for generics does not allow for this without getting type errors.
+
+On this snippet, we are trying to implement a theoretical Gun system, we may have many implementations for the Gun, like a Shotgun or a Pistol. In this implementation, we want to make a generic function that can take a gun of type `T` constrained by the base `Gun` type. But we also want to be able to return a result which is constrained by the type `WeaponResult`, however, the current system for generics does not allow for this without getting type errors.
 
 This is the main reasoning behind this modification. Many other languages have a feature to limit what types does a generic allow as a baseline. Languages like C# have this implemented on the form of the `where T : BaseClass` syntax, which delimits the type that `T` can be or inherit from. It supports cases in which you may want to use a strucutre, but not know its implementation details, like an interface.
+
+However, we can use this pattern to improve the reliability of code, a function such as the following that takes a string and returns a type, without having to type cast into `any` before type casting into `T`.
+
+
+```luau
+type GameService = {
+    -- ...
+}
+
+type GameManager = {
+    Disconnect: (self: GameManager) -> (),
+    -- ...
+} & GameService
+
+type ServiceContainer = {
+    ServiceList: { [string]: GameService },
+    Require: <T: GameService>(self: ServiceContainer, serviceName: string) -> T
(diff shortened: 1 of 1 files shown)


andyfriesen:
Constrained generics are absolutely something we're going to look at eventually, but this RFC needs a much more compelling example to properly justify them.

In your "Drawbacks" section, you point out a perfectly reasonable solution to your problem that doesn't require generics at all.

A more compelling example would be something like `function foo<T: Something>(t: T): T` where the type is generic and also appears in more than one part of the signature.  This function is inexpressible without a constrained generic.

SecondNewtonLaw:
> Constrained generics are absolutely something we're going to look at eventually, but this RFC needs a much more compelling example to properly justify them.
> 
> In your "Drawbacks" section, you point out a perfectly reasonable solution to your problem that doesn't require generics at all.
> 
> A more compelling example would be something like `function foo<T: Something>(t: T): T` where the type is generic and also appears in more than one part of the signature. This function is inexpressible without a constrained generic.

Thanks for the feedback, I have tried to add more complelling reasons, as well as more examples as to why this could be useful

docs/type-constraints-on-generic-parameters.md
```diff
@@ -0,0 +1,216 @@
...
+local gun: Gun
+local shotgun: Shotgun
+
+local fireResult = fireGun(gun) :: WeaponResult                     -- No error
+local shotgunFireResult = fireGun(shotgun) :: ShotgunWeaponResult   -- No error
+
+fireGun("Hello, world!")     -- Type Error: type 'string' and 'Gun' are not related to one another.
+```
+
+The syntax for defining a generic function would remain the same, except, instead of simply taking a `T`, we would take a `T : TypeConstraint`, the syntax remains similar to other features of the language, and it would work by checking if the type that `T` is inferred to be has any kind of relationship with `TypeConstraint`. However, we may also want to be explicit when we are doing generics on this way, therefore a syntax to properly and explicitly specify which types the function takes in its current call would be preferred, this way, we no longer have to use turbofish to type cast into the correct return type `U` constrained by `WeaponResult`.
```
Kampfkarren, line 135: I don't know why you're mentioning turbofish, that doesn't exist in Luau, and can't because it's ambiguous with `:: <T>() -> ()`.

No turbofish is also not an option for the exact same reasons it's necessary in Rust.

```luau
local a, b = the<guardian, stands>(resolute)
```

...is perfectly valid code today.

Kampfkarren: I remember discussing with Drak that `f.<T, U>()` isn't the worst syntax in the world, but also it feels out of scope for this RFC anyway?

SecondNewtonLaw: You're right, it feels out of scope for this RFC to add on about, and it was already somewhat addressed on the original generic functions rfc

SecondNewtonLaw: although I did not know that that snippet was valid ever, however, all I keep getting from attempting it is syntax errors

alexmccord:
I think I would also like to see 2 more things:

Type parameters are allowed to reference each other despite the lexical ordering, e.g. `<t: Thenable<a>, a>(...) -> (...)`. I think this is fine because types are not a runtime construct, they're a declarative language and as such, lexical ordering isn't super important.

A `where` clause that comes after the bindings list, but still within the `<>` part because of conflict with the rest of the grammar. This is something we'll need for adding constraints on some type functions but they don't directly show up as part of the parameters or returns, meaning they're internal constraints that the caller still need to know about. `<a, b where add<a, b>>(string, string) -> boolean`.

Here's a full EBNF describing the potential syntax, formally:

```ebnf
typaram ::= <NAME> [':' ty]
typarams ::= typaram {',' typaram}
tpparam ::= <NAME> '...' [':' tp]
tpparams ::= tpparam {',' tpparam}
params ::= (typarams tpparams?) | tpparams

tybound ::= ty [':' ty]
tybounds ::= tybound {',' tybound}
tpbound ::= tp [':' tp]
tpbounds ::= tpbound {',' tpbound}
bounds ::= (tybounds tpbounds?) | tpbounds

paramdecl ::= '<' params ['where' bounds] '>'
```

andyfriesen:
Something that this RFC doesn't touch on that merits scrutiny is the fact that constraints can also serve as witnesses.  Consider the following function:

```luau
function getKey<T, K: keyof<T>, V: index<T, K>>(t: T, k: K): V
    return t[k]
end
```

Ordinarily, when a type function is applied to a generic, we consider that function to be "stuck."  It cannot be reduced any further until the generic has been substituted for some concrete type.  In this case, however, the constraint `V : index<T, K>` indicates that `index<T, K>` is _not_ stuck.  It is `V`.

Kiiyoko:
Are you still planning on editing this RFC? If not, please let us know, otherwise, I have some feedback:
1. I believe that generic constraints should be available directly on type aliases as well. Currently, the syntax shown in the RFC only mentions generic function aliases like `type foo = <T: string | number>(T) -> T`. I should be able to embed the type `T` (with constraints) like `type foo<T: string | number> = (T) -> T`. 
2. The drawback isn't very compelling even after the initial round of edits. I think you should address [this reply](https://github.com/luau-lang/rfcs/pull/50#issuecomment-2245702202). 
3. It might be worth mentioning overloads as an alternative (with obvious drawbacks).
4. It's worth addressing the first part of [this reply](https://github.com/luau-lang/rfcs/pull/50#issuecomment-2250872992) since type ordering doesn't really matter in Luau.
5. Since calling functions with generic types are possible now with the `<<T>>()` syntax, the RFC should tackle this too.

SecondNewtonLaw:
> Are you still planning on editing this RFC? If not, please let us know, otherwise, I have some feedback:
> 1. I believe that generic constraints should be available directly on type aliases as well. Currently, the syntax shown in the RFC only mentions generic function aliases like `type foo = <T: string | number>(T) -> T`. I should be able to embed the type `T` (with constraints) like `type foo<T: string | number> = (T) -> T`. 
> 2. The drawback isn't very compelling even after the initial round of edits. I think you should address [this reply](https://github.com/luau-lang/rfcs/pull/50#issuecomment-2245702202). 
> 3. It might be worth mentioning overloads as an alternative (with obvious drawbacks).
> 4. It's worth addressing the first part of [this reply](https://github.com/luau-lang/rfcs/pull/50#issuecomment-2250872992) since type ordering doesn't really matter in Luau.
> 5. Since calling functions with generic types are possible now with the `<<T>>()` syntax, the RFC should tackle this too.

I thought this was partially already handled by another RFC. 

I will have another read to the whole PR when I get some free time and modify the existing RFC I drafted 😋

andyfriesen:
Constrained generics might be a nice fit for the "force this parameter to be a string singleton" problem.

The framing would be that we have a "strict subtype" constraint.

This would let us say "this parameter has to be a subtype of string, but can't actually be `string` itself."

PR #51, opened 2024-07-31
Draft and Discussion to figure out custom namecall self type definition

karl-police:
The RFC: https://github.com/karl-police/rfcs/blob/patch-1/docs/user_defineable_self_namecall_type.md

It's basically just all about

```lua
function Class:TestMethod()
	self = self :: ClassObject
end
```

and how ``:`` takes away the first argument, but I am not sure if this would cause issues for the typesolver. It seems to work fine when using ``--!strict`` when doing ``self = self :: ClassObject``

and when looking at the Bytecode

I don't see a instruction in the code where it is processing ``self = self`` which makes the current alternative within pure ``:`` functions possible, but annoying.

### DIFF
--- a/docs/user_defineable_self_namecall_type.md
+++ b/docs/user_defineable_self_namecall_type.md
@@ -1,0 +1,169 @@
+# User Defineable ``self`` Namecall Type
+
+## Summary
+
+
+The aim is to allow simple definition of ``self`` when defining functions with ``:``
+
+```lua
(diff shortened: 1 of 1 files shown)


andyfriesen:
Some of this is going to get a lot better when we implement https://github.com/luau-lang/rfcs/blob/master/docs/shared-self-types.md

PR #52, opened 2024-08-16
RFC: Tau constant

bjcscat:
[Rendered](https://github.com/bjcscat/rfcs/blob/constant-tau/docs/constant-tau.md)


bradsharp:
As a team I think we are on the fence about whether we should add `tau` to the standard library. It gets at a much deeper question though. Should the standard library contain derivates of fundamental constants, or only the fundamentals themselves? I think this is a question we'll need to discuss some more first as it governs the outcome of RFCs such as this. For now, I'd like to see more justification behind why `tau` itself would be a worthwhile addition, especially in relation to `pi`.

vegorov-rbx:
`tau` doesn't seem to have much value alone to be added to the Luau standard library.
It is derivative from the fundamental `pi` constant which we already have and it could be argued that other ratios are useful as well.

I understand for example that C library includes more 'helper' numbers and other `pi` ratios https://www.gnu.org/software/libc/manual/html_node/Mathematical-Constants.html as well as C++ https://en.cppreference.com/w/cpp/numeric/constants
But Lua/Luau `math` is pretty minimalistic and we also came far away in compiler technologies where computing the value could have had measurable runtime cost.

I would understand having a more comprehensive update to constants with a motivation for more things like `e` (which is computable and constant-foldable today, but not trivially), but a single `tau` alone seems like too much effort for little gain.

bjcscat:
I'm not going to contest that it's a small addition with marginal improvements in readability. If the use case isn't substantial enough to warrant inclusion if Luau is trying to remain minimal in this regard then I understand. 

For what its worth though, I would like to note that many other languages (specifically Java, Python, and .NET) have included tau constants in their standard libraries which indicates that there is a measure of support for inclusion of this constant in programming languages. Though I understand those languages don't share the same goal of minimalism as Luau.

I think that the inclusion of a tau constant would be a very minimal change a for small increase in readability in some specific places and I think that could be worth it.

vegorov-rbx:
> For what its worth though, I would like to note that many other languages (specifically Java, Python, and .NET) have included tau constants in their standard libraries which indicates that there is a measure of support for inclusion of this constant in programming languages.

I think this just comes from this strange effort to push for `tau` in each language even since that joke Tau manifesto came out.
Python docs for the constant are even written as a joke (but I do agree that Vi Hart's videos are great fun).

I think it's interesting to note that Rust, which is much more open to derivable math constants includes 9 ratios based on `pi` and not a single one based on `tau`, which is especially strange given that people talk about `tau` ratios being more easily understandable.
https://doc.rust-lang.org/stable/std/f64/consts/index.html
I think it points to the fact that `tau` has way more less use cases than `pi` ratios.

As an additional note, `pi` is never going away and with `tau` everyone has more things to remember and not even for a particularly good reason.

andyfriesen:
As a general principle, we include things into the Luau standard library because they are either actually critical for accomplishing some task, or at least they save users a bunch of work in certain common scenarios.

tau does not meet that bar.

Every programmer who has ever worked with pi is familiar with situations in which pi must be multiplied by two.  This isn't going to make anything more readable.

PR #53, opened 2024-08-23
RFC: `or return` syntax

ctrlaltmilk:
A small syntax extension to make handling `nil` values easier and more concise.

[Rendered](https://github.com/ctrlaltmilk/luau-rfcs/blob/or-return-syntax/docs/or-return-syntax.md)

### DIFF
--- a/docs/or-return-syntax.md
+++ b/docs/or-return-syntax.md
@@ -1,0 +1,71 @@
+# `or return` Syntax
+
+## Summary
+
+The `or return` syntax is a more compact, understandable way to end execution if an expression evaluates to `nil`.
+
+## Motivation
+
+Existing ways of handling errors or `nil` values in Luau tend to be either verbose or dense. Code with nested `if` statements can make indents unnecessarily large, which hinders readability. Checks after variable initialization are better, but still make code lengthy and annoying to write.
+
+`or return` would take this common idiom:
+```lua
+local var = foo()
+if not var then return false end
+```
+and compact it into a single, simpler statement:
+```lua
+local var = foo() or return false
+```
+
+This idiom in various forms is omnipresent in nearly all Luau code, with prime examples occuring in Roblox development docs. The [second scripting tutorial](https://create.roblox.com/docs/tutorials/scripting/basic-scripting/deadly-lava) has this piece of code:
+```lua
+local lava = script.Parent
+
+local function kill(otherPart)
+    local partParent = otherPart.Parent
+    local humanoid = partParent:FindFirstChild("Humanoid")
+    if humanoid then
+        humanoid.Health = 0
+    end
+end
+
+lava.Touched:Connect(kill)
+```
+With this RFC implemented, the code could shrink to this, which I personally feel is more understandable at a glance:
+```lua
+local lava = script.Parent
+
+local function kill(otherPart)
+    local partParent = otherPart.Parent
+    local humanoid = partParent:FindFirstChild("Humanoid") or return
+
+    humanoid.Health = 0
+end
+
+lava.Touched:Connect(kill)
+```
+
+## Design
+
+`or return` can be implemented as a binary operation with an optional right side, returning `nil` if no value is given.
+
+The parsing step would be fairly simple as well. Since the new syntax is a way to express an already-supported feature more concisely, it can be transformed back into its original form in the parser. This means no changes are necessitated on the compiler side. If specified in the middle of an expression, it can temporarily store the value on its left side to perform the nil check, then compute the rest of the expression using the original copy. For example,
+```lua
+local var = bar(foo or return)
+```
+would be expanded into
+```lua
+local tmp = foo
+if not foo then return end
+local var = bar(foo)
+```
+which would preserve the existing behavior of expressions, while avoiding the need for compiler or AST changes.
+
+## Drawbacks
+
+Implementing this as an operation might make parsing harder, especially since `or return` overlaps with `or`. The optional right argument could also be an issue in more complex expressions, though giving the operation a low priority could help by nudging the user towards grouping statements with parentheses.
+
+## Alternatives
+
+There are already alternatives built into the language - this is purely a convenience feature. Regardless, the other options are more effort to write, which I believe means this shortcut option is worthy of consideration.


dphfox:
I would really desperately want to see this - my code is littered with places this would be incredible - but a similar idea I mentioned a while back about "return in expressions" got shot down.

I really hope there is an expression of this design thst can be agreed upon.

docs/or-return-syntax.md
```diff
@@ -0,0 +1,71 @@
+# `or return` Syntax
+
+## Summary
+
+The `or return` syntax is a more compact, understandable way to end execution if an expression evaluates to `nil`.
```
alexmccord, line 5: Inconsistent with usual truthiness test. What if the expression evaluates to `false`?

bradsharp:
Would `or break` also make sense? I wonder if we could just allow `return` and `break` in expressions where they would otherwise make sense.

docs/or-return-syntax.md
```diff
@@ -0,0 +1,71 @@
...
+```
+With this RFC implemented, the code could shrink to this, which I personally feel is more understandable at a glance:
+```lua
+local lava = script.Parent
+
+local function kill(otherPart)
+    local partParent = otherPart.Parent
+    local humanoid = partParent:FindFirstChild("Humanoid") or return
+
+    humanoid.Health = 0
```
alexmccord, line 43: This is just `if local` which is proposed in a different RFC, so this isn't a strong motivator.

docs/or-return-syntax.md
```diff
@@ -0,0 +1,71 @@
...
+Existing ways of handling errors or `nil` values in Luau tend to be either verbose or dense. Code with nested `if` statements can make indents unnecessarily large, which hinders readability. Checks after variable initialization are better, but still make code lengthy and annoying to write.
+
+`or return` would take this common idiom:
+```lua
+local var = foo()
+if not var then return false end
+```
+and compact it into a single, simpler statement:
+```lua
+local var = foo() or return false
```
alexmccord, line 18: Doesn't compose with multrets.

docs/or-return-syntax.md
```diff
@@ -0,0 +1,71 @@
...
+end
+
+lava.Touched:Connect(kill)
+```
+
+## Design
+
+`or return` can be implemented as a binary operation with an optional right side, returning `nil` if no value is given.
+
+The parsing step would be fairly simple as well. Since the new syntax is a way to express an already-supported feature more concisely, it can be transformed back into its original form in the parser. This means no changes are necessitated on the compiler side. If specified in the middle of an expression, it can temporarily store the value on its left side to perform the nil check, then compute the rest of the expression using the original copy. For example,
```
alexmccord, line 53: Parser can't rewrite the AST beyond its local context, so AST changes are required for this feature to work.

docs/or-return-syntax.md
ctrlaltmilk: Probably should have clarified here - the main use case is for expressions that can return `nil`, but the already-established truthiness rules would be respected.

docs/or-return-syntax.md
ctrlaltmilk: `if local` does also solve this problem, but it also creates a new indentation level which is counter to the goals of this RFC.

docs/or-return-syntax.md
ctrlaltmilk: Why not? As far as I can tell, the only possible reason it wouldn't is if commas were used somewhere else in the expression, which is solved by using a pair of parentheses.

docs/or-return-syntax.md
ctrlaltmilk: That's understandable. Is there some form of AST expansion/lowering pass that this could fit into, or would this need to be supported in the compiler as well?

ctrlaltmilk:
> Would `or break` also make sense? I wonder if we could just allow `return` and `break` in expressions where they would otherwise make sense.

I did consider allowing `return` and `break` as expressions, but decided that would lead to confusing edge cases when called from a context where those statements wouldn't apply.

dphfox:
Here's an idea, `or else`.

It's never valid Luau to have the `or` keyword before an `else` keyword, so that sequence should be easy enough to detect. That would then allow proper support for `or else continue`.

And hey, it reads like posh English, that's fun too :)

witchiestwitchery:
> Here's an idea, `or else`.
> 
> ```Luau
> local humanoid = partParent:FindFirstChild("Humanoid") or else return
> ```
> 
> It's never valid Luau to have the `or` keyword before an `else` keyword, so that sequence should be easy enough to detect. That would then allow proper support for `or else continue`.
> 
> And hey, it reads like posh English, that's fun too :)

Feel like it should be orelse to be consistent with elseif

ccuser44 (changes requested):
This doesn't make to have in Luau and is syntactical bloat. And frankly, just looks kinda ugly.

Wunder-Wulfe:
I personally do not like the structure of this syntax. I would prefer something along the lines C# with their [Null-conditional Operators](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-).
```luau
local human: Humanoid? = hit.Parent?:FindFirstChildOfClass("Humanoid");

-- ...

local name: string? = script.Parent?.Name;

-- ...

local getFirst = list[key]?[1];
```

Or in the case of optimizing something more similar to the simple `if x==nil then x else y`, a structure like this:
```lua
local var = foo() if not nil else false;
```

surfbryce:
@Wunder-Wulfe Would probably be better that you make a RFC for this, probably paired with null assertion operators as well

alexmccord:
We already had the safe navigation operator RFC and we had to veto it out. 

1. https://github.com/luau-lang/luau/pull/501
2. https://github.com/luau-lang/luau/pull/142

gaymeowing:
> Not sure if it's ever been specifically considered, but we could adapt Rust's `let else` syntax, where the `else` branch must diverge. Maybe we don't have to make our else branches diverge like rust's if they evaluate to the same type (at typetime)? At runtime, if the expression isn't a return it just functions normally as a default value. In Luau it could be like:
> 
> ```luau
> local cat = animals_folder:FindFirstChild("Cat") else return nil
> print(cat.Position) -- indexing .Position doesn't error because cat can't be nil
> 
> -- runtime equivalent to
> local cat = animals_folder:FindFirstChild("Cat")
> if cat == nil then
>     return nil
> end
> 
> -- as default value
> 
> local cat = animals_folder:FindFirstChild("Cat") else default_cat
> 
> -- equivalent to
> local cat = animals_folder:FindFirstChild("Cat")
> if cat == nil then
>     cat = default_cat
> end
> ```

You've just described `or`

deviaze:
> You've just described `or`

iirc `or` doesn't work there w/ multirets and also evals if the LHS is `false` (not just nil)

PR #54, opened 2024-09-03
Add `lerp` to the standard `math` library

FunnyGlaggle:
[Add `lerp` to the standard `math` library](https://github.com/GlaggleGlaggler/rfcs/blob/master/docs/function-math-lerp.md)

### DIFF
--- a/docs/function-math-lerp.md
+++ b/docs/function-math-lerp.md
@@ -23,7 +23,7 @@ end
 ```
 
 When implementing a `lerp` function, we can judge how accurate the function is by the following criteria (credit to [this](https://github.com/rust-lang/rust/issues/86269#issuecomment-869108301) post about the nature of `lerp`). If the `lerp` function can meet all of these criteria, then the function is effectively perfect.
-- Is the function **exact**? This means that `math.lerp(x, y, 0) == x` and `math.lerp(x, y, 1) == ` exactly.
+- Is the function **exact**? This means that `math.lerp(x, y, 0) == x` and `math.lerp(x, y, 1) == y` exactly.
 - Is the function **consistent**? This means that `math.lerp(x, x, alpha) == x`, despite the value of `alpha`.
 - Is the function **monotonic**? This means that if `y` is strictly greater than `x`, `math.lerp(x, y, alpha)` should also strictly increase as `alpha` strictly increases, and vice versa for strictly decreasing. Now, due to the limitations of `double` precision, if `x` and `y` are very close together, there may be overlap where multiple values of `alpha` may correlate to the same result. This is ultimately unavoidable with any kind of `double` and `float` math, as we can observe the same behavior with `math.sin`, `math.cos`, etc.
 
--- a/docs/function-math-lerp.md
+++ b/docs/function-math-lerp.md
@@ -1,0 +1,36 @@
+# math.lerp
+
+## Summary
+
+Add `lerp` (linear interpolation) to the standard `math` library, which takes in two numbers, as well as a fractional value that interpolates between the two.
+
+## Motivation
+
+`lerp` is a very common mathematical formula and should be natively supported in Luau for ease-of-use and faster execution. Despite potential floating-point imperfections (which we will discuss later), the Luau `math` library already contains many floating-point-based formulas and we can design `math.lerp` to work in effectively 100% of use-cases.
+
+## Design
+
+The base logic without edge-cases will look as follows, but implemented in C++. As per already-existing `:Lerp` logic in the Roblox level above Luau, this function will be **unbounded**, meaning that `alpha` does not have to strictly be a number between 0 and 1, allowing for linear extrapolation. This also means that numbers that approach `inf` or `-inf` should return `inf` or `-inf` respectively to avoid overflow.
+
+```lua
+function math.lerp(x: number, y: number, alpha: number): number
+    if x == y then return x end
+    if alpha == 0 then return x end
+    if alpha == 1 then return y end
+    
+    return x + (y - x) * alpha
+end
+```
+
+When implementing a `lerp` function, we can judge how accurate the function is by the following criteria (credit to [this](https://github.com/rust-lang/rust/issues/86269#issuecomment-869108301) post about the nature of `lerp`). If the `lerp` function can meet all of these criteria, then the function is effectively perfect.
+- Is the function **exact**? This means that `math.lerp(x, y, 0) == x` and `math.lerp(x, y, 1) == ` exactly.
+- Is the function **consistent**? This means that `math.lerp(x, x, alpha) == x`, despite the value of `alpha`.
+- Is the function **monotonic**? This means that if `y` is strictly greater than `x`, `math.lerp(x, y, alpha)` should also strictly increase as `alpha` strictly increases, and vice versa for strictly decreasing. Now, due to the limitations of `double` precision, if `x` and `y` are very close together, there may be overlap where multiple values of `alpha` may correlate to the same result. This is ultimately unavoidable with any kind of `double` and `float` math, as we can observe the same behavior with `math.sin`, `math.cos`, etc.
+
+## Drawbacks
+
+As mentioned in the [Design](##design) section, the naïve implementation of `lerp` may introduce precision error, and all of the edge-cases will need to be accounted for.
+
+## Alternatives
+
+The only alternative solution would be to continue letting developers define their own lerp functions, placing it in a module and calling `require` all over the place, or placing it in a global scope such as `shared.lerp` or `_G.lerp` for easy access. However, this prohibits any Luau type-checking, and since `lerp` is an incredibly widely-used function, letting pure C++ handle the calculations can drastically improve performance.


docs/function-math-lerp.md
```diff
@@ -0,0 +1,36 @@
...
+- Is the function **monotonic**? This means that if `y` is strictly greater than `x`, `math.lerp(x, y, alpha)` should also strictly increase as `alpha` strictly increases, and vice versa for strictly decreasing. Now, due to the limitations of `double` precision, if `x` and `y` are very close together, there may be overlap where multiple values of `alpha` may correlate to the same result. This is ultimately unavoidable with any kind of `double` and `float` math, as we can observe the same behavior with `math.sin`, `math.cos`, etc.
+
+## Drawbacks
+
+As mentioned in the [Design](##design) section, the naïve implementation of `lerp` may introduce precision error, and all of the edge-cases will need to be accounted for.
+
+## Alternatives
+
+The only alternative solution would be to continue letting developers define their own lerp functions, placing it in a module and calling `require` all over the place, or placing it in a global scope such as `shared.lerp` or `_G.lerp` for easy access. However, this prohibits any Luau type-checking, and since `lerp` is an incredibly widely-used function, letting pure C++ handle the calculations can drastically improve performance.
```
vegorov-rbx, line 36: This is not true, module solution is the recommended one when working in Luau and it works with typechecking, so having custom implementations is not a problem at all. (I would say that having custom functions is the main mode of operation in Luau)

surfbryce: @vegorov-rbx I think "prohibits any Luau type-checking" was intended to be interpreted in the context of "placing it in a global scope such as `shared.lerp` or `_G.lerp`" rather than the requirement statement piece. The wording should definitely be changed however to clearly reference the latter alternatives rather than the former require alternative.

kalrnlo:
Could we get this merged? Doesn't seem like theres any objections to it.

vegorov-rbx:
Suggested wording change has never been applied. I would say there's no rush in getting it in.
Team is also focused on delivering other projects, so no need to rush things.

Additionally, the implementation is hard to follow with respect to the guarantees presented and whether or not we need them.
It doesn't match any of the popular implementations mentioned in the Rust issue.
It is also not as precise as the one from C++ (which I don't think is valuable to have).
We currently don't have time to do an analysis on numerical claims of that implementation, maybe the author can give a link to a paper that does that? I think that while special-casing at 3 points makes it exact, but might actually remove monotonicity.

Feels like that the developers will be better off implementing their usual custom implementation of a simple `x + (y - x) * alpha`.
Current implementation is complex enough with 3 condition statements to not be implemented in native codegen, so a custom function will win on performance.
My suggestion would be that developers who need precise numerical properties can have their own implementation with them, but that the one in Luau can have some properties missing, but be equal to the general expectation of what `lerp` does.

Here's the objection you wanted :)

docs/function-math-lerp.md
```diff
@@ -0,0 +1,36 @@
...
+- Is the function **exact**? This means that `math.lerp(x, y, 0) == x` and `math.lerp(x, y, 1) == y` exactly.
+- Is the function **consistent**? This means that `math.lerp(x, x, alpha) == x`, despite the value of `alpha`.
+- Is the function **monotonic**? This means that if `y` is strictly greater than `x`, `math.lerp(x, y, alpha)` should also strictly increase as `alpha` strictly increases, and vice versa for strictly decreasing. Now, due to the limitations of `double` precision, if `x` and `y` are very close together, there may be overlap where multiple values of `alpha` may correlate to the same result. This is ultimately unavoidable with any kind of `double` and `float` math, as we can observe the same behavior with `math.sin`, `math.cos`, etc.
+
+## Drawbacks
+
+As mentioned in the [Design](##design) section, the naïve implementation of `lerp` may introduce precision error, and all of the edge-cases will need to be accounted for.
```
AxisAngles, line 32:
```lua
-- lerp(a, b, t):
-- t can be much higher precision when near 0
-- so when t is near 0, we don't ever want to interpolate from a1 to a0
-- when t is closer to 0, we want higher accuracy around a
-- when t is closer to 1, we want higher accuracy around b
-- we want equations of the form A + t*B, as this guarantees monotonic behavior
-- at the interchange point, t = 1/2, we want to guarantee monotonic behavior
-- so we bound the result with respect to (a + b)/2

-- some things to consider:
-- (100000000 samples)

-- given x, y = sortabs(random()*randomsign(), random()*randomsign()):
--   x - y + y ~= x: ~49.2%
--   y - x + x ~= y: ~20.6%

-- given x, y = sortabs(random(), random()):
--   x - y + y ~= x: ~30.1%
--   y - x + x ~= y: ~3.98%

-- given x, y = sortabs(random()*1.5, random()*1.5):
--   x - y + y ~= x: ~25.4%
--   y - x + x ~= y: ~1.63%

-- given floorlog2(x) == floorlog2(y) == floorlog2(y - x)
--   x - y + y ~= x: 0%
--   y - x + x ~= y: 0%

local function lerp2(a0, a1, t)
	local m = (a0 + a1)/2
	if t <= 1/2 then
		local a = a0 + (a1 - a0)*t
		if a0 < a1 then
			return math.min(m, a)
		else
			return math.max(m, a)
		end
	else
		local a = a1 + (a1 - a0)*(t - 1)
		if a0 < a1 then
			return math.max(m, a)
		else
			return math.min(m, a)
		end
	end
end

-- but after testing, it appears that this never seems to expose un-monotonic behavior
-- Perhaps it has something to do with some specifics on how floating point rounds
local function lerp2(a0, a1, t)
	if t <= 1/2 then
		return a0 + (a1 - a0)*t
	else
		return a1 + (a1 - a0)*(t - 1)
	end
end
```

dphblox: so from that, seems like at *most* all we'd need is to conditionally swap `a` <-> `b` and invert `t` to reach most/all the guarantees we wanted? aka no reason to go more complex than `lerp2`?

dphblox: blending in @/vegorov-rbx's comments, would it even be worth special casing anything? would the only issue with using `a0 + (a1 - a0) * t` be that `t` has less precision around `1`?

AxisAngles: `t` has less precision around `1`, but there's no way to extract more precision from `t` in the first place, what they give is what we get to work with.

AxisAngles: From a user perspective, if I want more accuracy around `a1` than `a0`, I would need to find some accurate way of generating `s = 1 - t` directly, and then use `math.lerp(a1, a0, s)`

dphblox: Oh right this is about the precision of the multiplication with `(a1 - a0)`, right? We want to take advantage of the higher density of floating point numbers near 0 so we would rather "multiply down". e.g.: `x * 0.01` will be more precise than `x - (x * 0.99)`.

All makes sense - `lerp2` seems like the nicest "correct" solution. I suppose it just depends on whether we would want to use a conditional or whether that's deemed too heavy - will let the Luau folk speak to that.

FunnyGlaggle: Oh this is actually really clever, so the only "inaccuracy" we get is near 0.5

FunnyGlaggle: Thank you for this new insight, I will try and push this forward :)

zeux:
Some thoughts:

1. Do we need math.lerp? When this was raised at some point a couple years ago I was uncertain: math.lerp is useful wrt ergonomics, but a few uncertain properties around its implementation make it difficult to pick "the best one". Notably, Rust still hasn't added one. C++ has added a function that has guarantees that you'd want but the implementation is the one you'd never want if performance is even a little valuable.

However, the decision has been made to merge math.map; as such, I think math.lerp should be added: lerp is useful more often than map, math was added with no regard for any numerical properties, and while map can be converted into lerp by using 0,1 limits, that use is not ergonomic and not as performant (short of codegen cases where in theory map can be optimized optimally via partial constant folding).

So I support merging some RFC here in principle.

2. Proposed implementation is redundant: there is no need to check x==0 or x==y, this doesn't give any more guarantees. So I think this RFC should be amended to simplify the proposed lowering.

3. There's four properties I think we could consider:

a. exactness: lerp(0) = a, lerp(1) = b
b. monotonicity: lerp(t1) >= lerp(t2) for t1 >= t2
c. determinacy: lerp(a, b, t) is NaN only if t is infinite or a/b/t are NaN
d. boundedness: lerp(t) in [a, b]
e. consistency: lerp(x, x, t) == x for all t

C++ implementation satisfies all 5. However, the result is not useful wrt performance. In my opinion, determinacy is not particularly interesting as violating it requires fp overflow which is too strong of a constraint and not practically useful; exactness and consistency are important; boundedness is sometimes important; monotonicity is nice to have, as I do not know of practical applications where monotonicity is required.

This makes me wonder: why not choose `if t == 1 then b else a + (b - a) * t` as a suggested implementation? It is exact and consistent; it is bounded for reasonable values of a/b (is it always bounded?) and monotonic (for t in [0, 1]). If b-a overflows you get a bad lerp but that seems too strong of a constraint.

C++ implementation in https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0811r3.html is basically that, with an extra case for t=1+eps; I don't fully understand the value of monotonicity outside of the 0..1 interval, and as I mentioned above I don't know why monotonicity is important inside the interval, but certainly that property is not worth the extra branches.

I've empirically validated that the implementation above works well on 32-bit floats by bruteforcing one of a/b/t and fixing a few other values (for example, I validated boundedness by fixing a, fixing t at prevfloat(1), and bruteforcing all b >= a). This SO question https://math.stackexchange.com/questions/907327/accurate-floating-point-linear-interpolation has some discussion in the comments and comes to the same conclusion afaict.

I like the t<0.5 form in principle (note that it's mentioned in the comments of the issue linked above), but it has a steep implementation cost: you need to do two conditional selects on t<0.5 condition (switching a/b and t/t-1), plus compute t-1, and load two constants (0.5 & 1) into registers. A branch around 0.5 will behave well on microbenchmarks but cause branch mispredictions in some cases. Compared to this, t==1 is just one conditional select & load of just one constant, so it's cheaper.

4. Because the exact mechanics of lerp are tricky, I think it's probably fine to specify the above as a suggested implementation, and note that we want to guarantee exactness (and maybe consistency?), but not have explicit language around monotonicity et al. FWIW the formula `a + (b - a) * t` without corrections is not guaranteed to be bounded, however it's often bounded for interesting intervals that you choose to interpolate where boundedness is important (notably, intervals 0..X, -X..X, and in general +X..+Y seem to be bounded in practice?). But regardless it feels like it's fine to leave this unqualified for now.

zeux:
Oh and cost & properties of a few implementations; lerp 0 is `a + (b - a) * t`, lerp 1 is `a * (1 - t) + b * t`, lerp 2 is the implementation I suggest above, `lerp 3` is the implementation with `t<0.5` branch, `lerp 4` is `lerp 1` that enforces consistency via explicit `a == b` branch (from Rust discussion thread).

```
// lerp 0 (1mul): 3/3 insn; exact 0, monotonic 1, bounded 0, consistent 1
// lerp 1 (2mul): 5/5 insn; exact 1, monotonic 0, bounded ?, consistent 0
// lerp 2 (1mul t==1): 6/9 insn; exact 1, monotonic 1, bounded 1, consistent 1
// lerp 3 (1mul t<.5): 9/16 insn; exact 1, monotonic 1, bounded 1, consistent 1
// lerp 4 (2mul a==b): 7/9 insn; exact 1, monotonic 0, bounded ?, consistent 1
```

zeux:
I've submitted https://github.com/luau-lang/rfcs/pull/86 which intends to replace this RFC while providing the same functionality via `math.lerp` with the adjusted implementation following comments above and more fuzz testing.

vegorov-rbx:
Closing this as #86 has a better implementation if we were to include it.

PR #55, opened 2024-09-07
Add `math.map` to the standard library

MagmaBurnsV:
[Rendered](https://github.com/luau-lang/rfcs/blob/17bfcac8ea5f4bfff9206b0b4bb2fb1fa914a8b4/docs/function-math-map.md)

### DIFF
--- a/docs/function-math-map.md
+++ b/docs/function-math-map.md
@@ -1,0 +1,36 @@
+
+# math.map
+
+## Summary
+
+Add a `map` function to the `math` library, which maps a number from one range to another.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


docs/function-math-map.md
```diff
@@ -0,0 +1,36 @@
...
+function math.map(x: number, inmin: number, inmax: number, outmin: number, outmax: number): number
+	return outmin + (x - inmin) * (outmax - outmin) / (inmax - inmin)
+end
+```
+
+Just like common Luau implementations, `x` is allowed to be outside the input range. Since clamped mapping is not as widely used and can be easily replicated using `math.clamp`, it would not be supported.
+
+## Drawbacks
+
+This would add a new function to the native library, taking yet another fastcall slot.
```
vegorov-rbx, line 28: We do not start with adding every new function to fastcalls, so we will just skip this one, like random\noise.

vegorov-rbx:
Looks alright.
Let's wait for 2 week minimum for other feedback.

kalrnlo:
Could this be merged? Its been a month and theres been no objections.

vegorov-rbx (approved):

vegorov-rbx:
As a more simple addition, this should be available in next release after the sync.

AxisAngles:
https://gist.github.com/AxisAngles/e3c6ad3f57044480b69ff378477db03a

PR #56, opened 2024-09-11
RFC: Amended Require Syntax and Resolution Semantics

vrn-sn:
[Rendered](https://github.com/vrn-sn/rfcs/blob/alias-resolution-rfc/docs/amended-require-resolution.md)

We need to finalize a syntax for aliases in require expressions and determine intuitive resolution semantics as we prepare to build a package management system.

### DIFF
--- a/docs/amended-alias-resolution.md
+++ b/docs/amended-alias-resolution.md
@@ -12,10 +12,12 @@ In an [old RFC for require-by-string](https://github.com/luau-lang/luau/pull/969
 the path `require("my/module")` is resolved by first checking if `my` is an alias defined in a `.luaurc` file.
 If so, we perform a string substitution on `my`.
 If not, we try to resolve `my/module` relative to the requiring file.
+This is identical to approach (3a) below.
 
 This design changed in the next few months and became what we now have in [require-by-string-aliases.md](require-by-string-aliases.md):
 when requiring an alias, it must be prepended with the `@` prefix.
 Otherwise, we will assume the given string does not contain an alias and will resolve it relative to the requiring file.
+This is identical to approach (1a) below.
 
 There have been disagreements about which approach is more appropriate, so this RFC will serve as a way to document the ongoing discussion and the final decision we make.
 
--- a/docs/amended-alias-resolution.md
+++ b/docs/amended-alias-resolution.md
@@ -55,9 +55,9 @@ The main drawback of this approach is that projects with many _external_ depende
 
 This is similar to [Lune's approach](https://lune-org.github.io/docs/getting-started/2-introduction/8-modules#file-require-statements) and is identical to the approach currently described in [require-by-string-aliases.md](require-by-string-aliases.md).
 
-#### (1b) Make relative paths explicit with either `./` or `../` prefix
+#### (1b) Make relative paths explicit with `./` prefix
 
-We make usage of either the `./` or `../` prefix necessary in relative require statements.
+We make usage of the `./` prefix necessary in relative require statements.
 This means that any unprefixed string is always treated as an aliased path.
 ```luau
 -- Requires /My/Libraries/Directory/dependency.luau
@@ -70,7 +70,7 @@ require("./libs/dependency")
 require("@libs/dependency")
 ```
 
-The main drawback of this approach is that projects with many _internal_ dependencies will have many `./` and `../` symbols appearing within require statements.
+The main drawback of this approach is that projects with many _internal_ dependencies will have many `./` symbols appearing within require statements.
 
 This is similar to [darklua's approach](https://darklua.com/docs/path-require-mode/).
 
--- a/docs/amended-alias-resolution.md
+++ b/docs/amended-alias-resolution.md
@@ -105,14 +105,14 @@ These approaches were deemed undesirable during discussion and are no longer bei
 
 ### (2) Strongly recommend explicit prefixes
 
-We don't make the `@`, `./`, or `../` prefixes necessary like in approach (1c), but we strongly recommend them.
+We don't make the `@` or `./` prefixes necessary like in approach (1c), but we strongly recommend them.
 
 This can be accomplished by allowing unprefixed paths but throwing an error if they resolve to multiple files.
 There is a performance cost to this: currently, if we successfully match a given path to a file, we stop searching for other matches.
 If this approach were implemented, we would need to search for at most two matching files in case we need to throw an error.
 In the case of multiple nested directories with their own `.luaurc` files and `paths` arrays, this search could take linear time with respect to the total number of paths defined, in the worst case.
 
-Of course, developers could use the `@`, `./`, or `../` prefixes to explicitly specify a search type and eliminate this performance cost.
+Of course, developers could use the `@` or `./` prefixes to explicitly specify a search type and eliminate this performance cost.
 ```luau
 -- Error: ./libs/dependency.luau exists and .luaurc defines a distinct "libs" alias
 -- Would require ./libs/dependency.luau if "libs" alias did not exist
--- a/docs/amended-alias-resolution.md
+++ b/docs/amended-alias-resolution.md
@@ -19,7 +19,7 @@ Otherwise, we will assume the given string does not contain an alias and will re
 
 There have been disagreements about which approach is more appropriate, so this RFC will serve as a way to document the ongoing discussion and the final decision we make.
 
-## Design and Drawbacks
+## Possible Designs and Drawbacks
 
 Below, we have multiple different options for syntax and resolution semantics.
 For each design, assume that the following `.luaurc` file is defined:
@@ -74,9 +74,26 @@ The main drawback of this approach is that projects with many _internal_ depende
 
 This is similar to [darklua's approach](https://darklua.com/docs/path-require-mode/).
 
+#### (1c) All explicit prefixes necessary
+
+An alternative approach is to combine approaches (1a) and (1b) and always require prefixes for disambiguation.
+In this scenario, any unprefixed path will _always_ result in an error.
+```luau
+-- Error: must have either "@", "./", or "../" prefix
+require("libs/dependency")
+
+-- Requires ./libs/dependency.luau
+require("./libs/dependency")
+
+-- Requires /My/Libraries/Directory/dependency.luau
+require("@libs/dependency")
+```
+
+The main drawback of this approach is that it is not backwards compatible with any existing code that uses unprefixed require statements.
+
 ### (2) Strongly recommend explicit prefixes
 
-We don't make the `@`, `./`, or `../` prefixes necessary, but we strongly recommend them.
+We don't make the `@`, `./`, or `../` prefixes necessary like in approach (1c), but we strongly recommend them.
 
 This can be accomplished by allowing unprefixed paths but throwing an error if they resolve to multiple files.
 There is a performance cost to this: currently, if we successfully match a given path to a file, we stop searching for other matches.
@@ -99,31 +116,13 @@ require("@libs/dependency")
 
 This is a compromise between approaches (1a) and (1b) that sacrifices performance for compatibility and readability.
 
-## Alternatives
-
-### All explicit prefixes necessary
-
-An alternative approach is to combine approaches (1a) and (1b) and always require prefixes for disambiguation.
-In other words, this is similar to approach (2), but any unprefixed path will _always_ result in an error.
-This avoids the performance cost of approach (2), but it is not backwards compatible with existing code.
-```luau
--- Error: must have either "@", "./", or "../" prefix
-require("libs/dependency")
-
--- Requires ./libs/dependency.luau
-require("./libs/dependency")
-
--- Requires /My/Libraries/Directory/dependency.luau
-require("@libs/dependency")
-```
-
-### Use implicit resolution ordering to avoid `@` prefix
+### (3) Use implicit resolution ordering to avoid `@` prefix
 
 If we choose to remove the `@` prefix, we must define an implicit resolution ordering.
 In essence, aliases must either represent overrides or fallbacks for relative path resolution.
 A drawback of this approach is that implicit resolution ordering can be confusing, and people disagree on which ordering is best.
 
-#### Aliases as overrides
+#### (3a) Aliases as overrides
 
 If aliases were overrides, we would have the following behavior.
 This is identical to the [old RFC](https://github.com/luau-lang/luau/pull/969)'s behavior.
@@ -138,7 +137,7 @@ require("./libs/dependency")
 require("@libs/dependency")
 ```
 
-#### Aliases as fallbacks
+#### (3b) Aliases as fallbacks
 
 If aliases were fallbacks, we would have the following behavior.
 ```luau
--- a/docs/amended-alias-resolution.md
+++ b/docs/amended-alias-resolution.md
@@ -137,6 +137,18 @@ require("./libs/dependency")
 require("@libs/dependency")
 ```
 
+If the `libs` alias did not exist, however, we would have the following:
+```luau
+-- Requires ./libs/dependency.luau
+require("libs/dependency")
+
+-- Requires ./libs/dependency.luau
+require("./libs/dependency")
+
+-- Error: there is no "@libs" alias in .luaurc
+require("@libs/dependency")
+```
+
 #### (3b) Aliases as fallbacks
 
 If aliases were fallbacks, we would have the following behavior.
--- a/docs/amended-alias-resolution.md
+++ b/docs/amended-alias-resolution.md
@@ -98,6 +98,11 @@ Approaches (1a) and (1b) would be less clear about this, as `require("` could be
 
 The main drawback of this approach is that it is not backwards compatible with any existing code that uses unprefixed require statements and may result in code that looks cluttered.
 
+## Alternatives
+
+Below are options that were proposed in an earlier version of this RFC, with their numbered labels preserved.
+These approaches were deemed undesirable during discussion and are no longer being considered.
+
 ### (2) Strongly recommend explicit prefixes
 
 We don't make the `@`, `./`, or `../` prefixes necessary like in approach (1c), but we strongly recommend them.
@@ -122,12 +127,13 @@ require("@libs/dependency")
 ```
 
 This is a compromise between approaches (1a) and (1b) that sacrifices performance for compatibility and readability.
+Unfortunately, this approach could lead to unexpected behavior if users have globally defined `.luaurc` files that conflict with scripts that expect certain aliases to not be defined.
 
 ### (3) Use implicit resolution ordering to avoid `@` prefix
 
 If we choose to remove the `@` prefix, we must define an implicit resolution ordering.
 In essence, aliases must either represent overrides or fallbacks for relative path resolution.
-A drawback of this approach is that implicit resolution ordering can be confusing, and people disagree on which ordering is best.
+These approaches are not being considered since implicit resolution ordering can be confusing, and people may disagree on which ordering is best.
 
 #### (3a) Aliases as overrides
 
--- a/docs/amended-alias-resolution.md
+++ b/docs/amended-alias-resolution.md
@@ -91,7 +91,12 @@ require("./libs/dependency")
 require("@libs/dependency")
 ```
 
-The main drawback of this approach is that it is not backwards compatible with any existing code that uses unprefixed require statements.
+A benefit of this approach is that the role of autocomplete becomes fairly well-defined.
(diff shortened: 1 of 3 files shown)


docs/amended-alias-resolution.md
```diff
@@ -0,0 +1,165 @@
...
+require("@libs/dependency")
+```
+
+The main drawback of this approach is that projects with many _external_ dependencies will have many `@` symbols appearing within require statements.
+
+This is similar to [Lune's approach](https://lune-org.github.io/docs/getting-started/2-introduction/8-modules#file-require-statements) and is identical to the approach currently described in [require-by-string-aliases.md](require-by-string-aliases.md).
+
+#### (1b) Make relative paths explicit with either `./` or `../` prefix
+
+We make usage of either the `./` or `../` prefix necessary in relative require statements.
```
alexmccord, line 60: I personally like this because it resolves the footgun with missing `@` and for a while we actually had this. I hope we remove support for require by implicit relative path.

docs/amended-alias-resolution.md
```diff
@@ -0,0 +1,165 @@
...
+
+-- Requires /My/Libraries/Directory/dependency.luau
+require("@libs/dependency")
+```
+
+This is a compromise between approaches (1a) and (1b) that sacrifices performance for compatibility and readability.
+
+## Alternatives
+
+### All explicit prefixes necessary
```
alexmccord, line 104: Why is this under Alternatives instead of a design under ongoing consideration? "but it is not backwards compatible with existing code" is something we broke several times already in non-Roblox contexts.

vrn-sn: Honestly, this is a bit of an untraditional RFC because it's not exactly proposing anything yet, just laying out options. I'll bring in the alternatives into the main design section, and move the rejected ones back to Alternatives once we've all discussed the options.

docs/amended-alias-resolution.md
```diff
@@ -0,0 +1,165 @@
...
+require("libs/dependency")
+
+-- Requires ./libs/dependency.luau
+require("./libs/dependency")
+
+-- Requires /My/Libraries/Directory/dependency.luau
+require("@libs/dependency")
+```
+
+The main drawback of this approach is that projects with many _internal_ dependencies will have many `./` and `../` symbols appearing within require statements.
```
vegorov-rbx, line 73: Not sure why `../` is even up for discussion, there is no implicit mode for `../`.
To get up from the current 'folder' into the parent one, `../` is required in all alternatives here and in current implementation.

vrn-sn: Good point, just removed certain references to `../` that make this confusing.

dphfox:
I'm torn personally. Originally I envisioned this using @ personally (which I'd've been fine with - it's load bearing syntax!) but now that you mention it, it's normally nicer to explicitly mark relative requires. But of course, that's not how existing string requires work. Hmm.

I'd say prioritise backwards consistency. I don't think @ is that bad in the slightest, even though it's lamentable that people use relative paths as they do.

alexmccord:
It must be mentioned that I am of the opinion that having a require resolution order is a workaround for poor design, so I'm very firmly in the camp of explicit `@`, `./`, and `../`.

docs/amended-alias-resolution.md
vrn-sn: I believe both (1b) and (1c) force relative paths to be explicitly specified. I thought for a moment that (3a) also supported this, but I was missing a use case that I just added. So, (3a) also supports implicit relative paths.

bradsharp:
> I'd say prioritise backwards compatibility with Lua (we want to give everyone the option to integrate with the wider Lua ecosystem at some point, right?). I don't think @ is that bad in the slightest, even though it's lamentable that people use relative paths as they do.

Lua uses a different convention entirely with `.` as the separator. Additionally, a number of Lua libraries are written in C so I'm not sure how practical that is (but it's something we should definitely be thinking about).

bradsharp:
My preference is for 1b, relative requires must start with `./` which was the behavior proposed in [#969](https://github.com/luau-lang/luau/pull/969).

- Some applications treat `./foo` and `foo` differently when there's a different file directory and working directory
- Any path starting with `.` is always going to be a relative file path
- Prefixing all my external imports with `@` makes them look a bit ugly

Unfortunately, this also breaks some existing Luau code but I'm hoping since we're discussing this early enough that the impact is minimal.

dphblox:
Got it. I was under the impression Lua used normal fs paths, thanks for the correction :)

> It must be mentioned that I am of the opinion that having a require resolution order is a workaround for poor design

Firm agree. In general, it should be clear and unambiguous what you're trying to require, at the point of the require. I don't want to have to go hunting dependencies down.

> In the past Luau was more restrictive, requiring @ and ./ but at some point that changed to remove the ./ requirement. I think the question here is whether that should have happened

Yeah, I personally think `./` should be required to make the intent clear. At that point, I suppose you can go with or without `@` - since it's clearly and unambiguously referring to an alias - so I could live with either choice.

vrn-sn:
I want to wait for @vegorov-rbx and @aatxe to express their preferences on the designs, but it does seem like we're leaning in the direction of the explicit resolution methods (1a, 1b, 1c), rather than the implicit ones (3a, 3b). The "almost-explicit" (2) is really just a less performant version of (1c) that tries to minimize breaking backwards compatibility—if we don't care about this, then I think (1c) is strictly better than (2).

One confusing case with (2) that I can imagine is a developer's `.luaurc` file (perhaps globally defined) defining conflicting aliases with an external Luau packages' relative requires. (Although, maybe that would be the package developer's responsibility to avoid by using the explicit `./` prefix.) The point remains that (2) is not entirely environment-agnostic when it comes to unprefixed relative paths, so it could lead to some unexpected behavior.

I'm a fan of (1c). Despite it being a bit uglier, I think it's the responsibility of the language to be unambiguous, and the responsibility of an LSP to make it easy for the developer (i.e. autocomplete). With (1c), autocomplete can have a clear indication of when to branch to a new set of possibilities.
```luau
-- Autocomplete options: "@", "./", "../"
require("
```
```luau
-- Autocomplete options: all available aliases
require("@
```
```luau
-- Autocomplete options: all available modules in requirer's directory
require("./
```

In contrast, (1a) and (1b) are slightly more confusing in terms of autocomplete options.
```luau
-- Autocomplete options: need to show valid string path components and all three prefixes ("@", "./", "../")
-- "Valid string path components" means local modules for (1a) and aliases for (1b)
require("
```

aviralg:
1A, 1B, and 1C all seem fine to me. I think there is value in 1C since it is explicit and enables better autocomplete support. 
Personally, I don't find prefixing paths with either `@` or `./`, ugly. 
I am interested in knowing if compatibility with DarkLua and Lune matters to our users.

aatxe:
All of the (1) options are good with me, with a preference towards 1B or 1C. I'm okay with the idea that libraries are the default (and therefore aliases are the default), hence 1B, and 1C is more-or-less the same to me, just with library names being under `@`. The most important consideration for me is not having implicit resolution rules that people have to learn and internalize (3), and not having strange quirky behaviors depending on the environment (2), and that's why the (1) options are the best in my book.

dphblox:
Judging by the vibes in these replies, I guess the question is just whether we think relative paths are more frequently used than aliases. We could probably try and get some metrics for this if we wanted data.

Does anyone have strong feelings about relative paths without a prefix?

vrn-sn:
> I guess the question is just whether we think relative paths are more frequently used than aliases.

We also care about how we want to design our require syntax as we think about a package manager. If backwards compatibility isn't a high priority, then it might not matter if relative requires are used more frequently right now. For example, if our ultimate goal is to support modular workflows that involve aliased imports (facilitated by a package manager), then it may make more sense to lean away from (1a), which simplifies the syntax for relative paths instead.

My point is that our goals for future usage patterns may matter more than current usage patterns, depending on how highly we prioritize backwards compatibility.

dphblox:
Fair point, I hadn't considered that.

I guess something else we should remember is that we can always "reserve" the unprefixed variant and make this choice when we have more of an idea of how package management is going to affect all of this, and use both `@` and `./` for now.

Though personally, in light of this information, I'm already leaning towards mandating `./` for relative paths, since I agree that modular workflows would likely take over quite significantly once we have a proper package management system in place.

Dekkonot:
I think option 3 is unacceptable, but that thankfully seems to be a very popular opinion. In the same vein, 2 is just a worse version of 1c. If you make something optional, it's probably not going to be used by the bulk of users, so we should just bite the bullet and make it required.

I don't super care for requiring a prefix for relative requires, because I like being able to go `require("sibling")`. It's unintuitive to me at least to require it to be `require("./sibling")`. That's really my only complaint about option 1c though, so I'm favoring that one. If it turns out to be a problem, it can always be relaxed because it effectively reserves the unprefixed require syntax.

docs/amended-alias-resolution.md
```diff
@@ -0,0 +1,183 @@
...
+A benefit of this approach is that the role of autocomplete becomes fairly well-defined.
+When `require("` is typed, autocomplete has only three options to display: the `@`, `./`, and `../` prefixes.
+After a prefix has been typed, autocomplete will either display defined aliases or local modules, depending on the prefix.
+Approaches (1a) and (1b) would be less clear about this, as `require("` could be followed by a string component or one of the three prefixes.
+
+The main drawback of this approach is that it is not backwards compatible with any existing code that uses unprefixed require statements and may result in code that looks cluttered.
+
+### (2) Strongly recommend explicit prefixes
+
+We don't make the `@`, `./`, or `../` prefixes necessary like in approach (1c), but we strongly recommend them.
```
vegorov-rbx, line 103: `../` is always necessary in all choices.

vrn-sn: Fixed

vegorov-rbx:
Took a look at our internal Luau codebase.
It's a bit hard to categorize, but here are some results:

20090 total require statements.

- 8358 refer to an external package folder (would be explicit `@`).
- 6584 refer to an internal module coming described from the root (would be either an `@` alias to the root or the 'paths' will contain root path, so not a `./`)
- 2824 refer to an internal module not from root (would be a specific alias `@`)
- 1023 refer to a sibling (would be explicit `./`)
- 808 refer to an outer directory using `FindFirstAncestor`, better to rewrite into an `@` alias instead
- 60 refer to an outer directory (`../`)
- 433 remaining look mostly like something which could use an alias

vrn-sn:
I've also moved approaches (2), (3a), and (3b) to alternatives. It seems like we're in favor of the explicit options and now just need to choose from (1a), (1b), and (1c).

@dphblox and @Dekkonot both mentioned a great point about how (1c) effectively reserves unprefixed paths, so I just added a small section to (1c) about it being forwards-compatible with both (1a) and (1b).

vegorov-rbx:
Worth noting that you are not prevented from writing `./` even if it's not explicitly required.

dphblox:
On the topic of LSP support btw: I feel like we should be optimising these paths for reading, because autocomplete will likely do most of the work here for us.

As an example, suppose you write: `require("`

At that point, you'll immediately disambiguate what you mean with the next character you type, regardless of what option we go with. If you typed `.` for example, you'd immediately start getting file path suggestions. 

From that perspective, we're deciding what we're going to suggest to you first, if anything at all. If I type `R` next, will it show me `React` from my aliases, or `RenderUI` from my current folder, or will both be suggested with their respective prefixes? Do we want to keep this autocomplete predictable to aid muscle memory or do we want it to be more smart about what's around you?

I would say that predictable autocomplete is a good idea to pursue, because developers could exploit that muscle memory to work faster. But overall, all of these workflows are plenty fast, so it's less relevant compared to reading, which is both done more often and isn't accelerated by the computer in the same way.

JohnnyMorganz:
Is it worth noting how these different semantics work with the `paths` aspect that was mentioned in the original RFC (https://github.com/luau-lang/rfcs/blob/master/docs/require-by-string-aliases.md#paths)? As it doesn't look like it's been considered here, apart from a brief mention in option (2).

Particularly that right now unprefixed paths can either be relative files to the current file *or files resolved from the `paths` array*. What prefix would files resolved via paths be given for option (1b) and (1c), if any? (Given that the original RFC discusses paths in the context of relative requires)

kalrnlo:
> Judging by the vibes in these replies, I guess the question is just whether we think relative paths are more frequently used than aliases. We could probably try and get some metrics for this if we wanted data.
> 
> Does anyone have strong feelings about relative paths without a prefix?

Theres not really any good analytics for string requires, because it requires the use od

Dekkonot:
Anecdote: I'm writing a lot of Lune code for work right now, so I went through the effort writing requires "correctly" to see how I felt about it.

As predicted, including the `./` in requires is incredibly cumbersome, especially for symbols. I ended up writing `.modules/FileName` more often than `./modules/FileName`. It may be a skill issue on my part, but I'm really not in love with it.

I think I might have to change my stance to 1a just based on my own experience. The harm to intellisense isn't worth the hassle IMO.

dphfox:
Agreed on path resolution being weird. I really don't think we should be using path resolution at all. It strikes me as exactly the kind of implicit behaviour that people here are taking issue with - better to explicitly alias?

As for  `./` being hard to write in requires, what does the autocomplete look like today? It could be that LSP support isn't where it needs to be yet.

kalrnlo:
> Agreed on path resolution being weird. I really don't think we should be using path resolution at all. It strikes me as exactly the kind of implicit behaviour that people here are taking issue with - better to explicitly alias?
> 
> As for `./` being hard to write in requires, what does the autocomplete look like today? It could be that LSP support isn't where it needs to be yet.

Autocomplete doesnt exist, I mean it'll warn you for providing an invalid path but that's it.

dphblox:
I suppose to be clearer about why I don't like it, it's because I don't like it when projects share packages. It leads to all sorts of problems, especially when two projects need a package of the same name, but different versions (or worse - two unrelated packages with a name collision). I feel like `paths` doesn't serve much use within a single repository (just use aliases), and only becomes useful if you're trying to implement that sort of shared system. 

I suppose all of this'll end up being the responsibility for whatever package management system we land on though, so maybe it's OK to have? I'm not sure.

It feels a little confused to me, I guess?

JohnnyMorganz:
Agreed with all the points raised by @dphblox. Personally I'm not sold on the use cases of `paths`, they don't seem that strong. and I also feel like it reverts to that implicit ordering that people raised as a concern.

Maybe it's not *that* implicit, but the ordering still matters. For `./` paths it is quite direct what file it points to (1:1 mapping), similarly for aliases you can resolve a 1:1 mapping, but for paths you have to linearly search through all the defined paths to figure out what file it's pointing too (I care about this more from the perspective of a human reading it statically rather than a computer perspective - computers are fast and LSP can provide go to definition, but those tools aren't always available during a code review). Ordering of the paths in the array can affect resolution of files, when there are two files with the same (sub-)path across different paths. Would one need to go back to the .luaurc, visit one by one each directory path specified in order to then figure out where a require is actually pointing to?

The behaviour of paths can also be  replicated using aliases instead (e.g `@dep/` instead of a dependencies folder in a paths array). Sure, couple extra characters, but I think the explicitness is more valuable.

I would argue for actually removing paths support completely, but maybe that's too strong / it's too late.

JohnnyMorganz:
Not completely related to this RFC, so feel free to ignore, but also worth considering how `paths` would interact with a future package manager? 

I cannot see a world where paths would be usable when you want to have portable packages, apart from if the only paths listed there are within the scope of the package itself. But to be fair, the same restriction would probably apply to aliases too. But I think those types of aliases are still useful, whilst "only-within-package-scope" paths don't seem like it to me.

I haven't thought about that too deeply though (and I may be biased against paths at this point :P) so maybe you folks have already considered this.

alexmccord:
Can you give us a refresher on what the `paths` part is for? I don't know what it does, what problems it solves, or whether it's a historical foible that could be reworked/be made redundant with aliases, so I'm struggling to make an opinionated stance on how I feel about `paths` in relation to the require by string.

vrn-sn:
Discussed with @aatxe a bit just now.

The initial motivation for `paths` was likely to support something similar to Lua's `LUA_PATH`. At this point, we've diverged enough from Lua that I can see the argument for removing `paths` altogether. Parroting the good points made above, I believe any use case that involves `paths` can use `aliases` instead, and this also has the benefit of being more explicit.

Our package manager implementation could utilize `paths`, but after discussion, we'll be fine solely using aliases instead to point to dependencies. Also, from looking at the [analysis](https://github.com/luau-lang/rfcs/pull/56#issuecomment-2354123259) that @vegorov-rbx posted, the only case in which the `paths` variable might be appropriate would work fine implemented as an alias:

> 6584 refer to an internal module coming described from the root (would be either an @ alias to the root or the 'paths' will contain root path, so not a ./)

I'll refactor this RFC to propose removing the `paths` variable altogether. I also want to start to move toward one of the three approaches (1a, 1b, 1c), assuming `paths` no longer exists.

docs/amended-require-resolution.md
```diff
@@ -0,0 +1,196 @@
...
+        "libs": "/My/Libraries/Directory"
+    }
+}
+```
+Additionally, assume that we have a file named `dependency.luau` located in these two locations:
+1. `./libs/dependency.luau`
+2. `/My/Libraries/Directory/dependency.luau`
+
+As part of the push to make require statements more explicit, we also propose removing `paths`, as aliases serve a similar purpose and are more explicit.
+The following options operate under the assumption that `paths` is no longer supported in `.luaurc` files, so we do not need to provide a syntax that corresponds to `paths`.
```
vrn-sn, line 40: I've added a couple of sentences describing the proposal to remove `paths` altogether.

vrn-sn:
To help this proposal move forward now, I think the best course of action is to formally propose (1c) and move (1a) and (1b) to alternatives. Approach (1c) is the most restrictive of the three options and prevents us from making a decision about unprefixed paths before we know what the most common use case will be (say, in the context of a package manager).

Despite being the most restrictive, it feels like a good compromise to me because of this forward compatibility with both (1a) and (1b).

I get your point here @Dekkonot:
> As predicted, including the `./` in requires is incredibly cumbersome, especially for symbols. I ended up writing `.modules/FileName` more often than `./modules/FileName`. It may be a skill issue on my part, but I'm really not in love with it.
> 
> I think I might have to change my stance to 1a just based on my own experience. The harm to intellisense isn't worth the hassle IMO.

Still, I'm a bit wary about going in the direction of (1a), building a package manager, and then getting similar complaints about the mandated `@` prefix (assuming aliased modules will be more commonly used). If anything, workflows in the context of package management seem to fit better with (1b), but I think it's too early to make a definite decision right now.

I think the best option is to choose the approach that gives us the most flexibility with unprefixed paths. Unless someone has strong opposing thoughts on this, I think we should move forward with (1c).

vrn-sn:
Reworked the RFC: would appreciate PR review now. I'll plan on merging in these changes in a week if there is no opposition.

alexmccord:
One last thing I'm not currently seeing in the RFC. If we still support both `.luau` and `.lua` extensions, then there is still a require resolution ordering and that has two possible variants.

Given this scenario:
```
src/
  mod/
    init.luau
  mod.luau
  mod.lua
```

If I have `require("./src/mod")`, what does this resolve to?

aatxe:
Two sane behaviors IMO are: prefer luau to lua, or error about the conflict.

Dekkonot:
My preference would be an error in that situation, because I cannot imagine it being deliberate.

vrn-sn:
I've added a section about throwing an error in any case in which there's ambiguity, like @alexmccord's example:

> Given this scenario:
> 
> ```
> src/
>   mod/
>     init.luau
>   mod.luau
>   mod.lua
> ```
> 
> If I have `require("./src/mod")`, what does this resolve to?

I think that's in line with making everything else explicit that we've discussed in this PR.

dphfox:
On the topic of ambiguity, would this also error if multiple files shared the same name? This isn't possible on disk but is possible in the DataModel.

vrn-sn:
@dphfox and I discussed this internally:

> Right now when performing an ambiguous indexing operation in Roblox, it always has to succeed for backcompat reasons, but perhaps we could take this opportunity to explicitly disallow it for this new syntax, and for any path resolution style APIs going forward?

We're going to leave Roblox-related details out of the scope of this RFC, but the goal is to keep things as compatible as possible. In Luau's implementation, I don't think we need to add an additional check for this, as the file system implicitly enforces it for us.

Dekkonot:
That's a fair attitude to take. It's an implementation detail that Roblox is a graph. Basically everywhere else Luau exists it shouldn't matter.

andyfriesen (approved):

PR #58, opened 2024-10-04
RFC: Type Metamethod Support

metatablecat:
Suggests allowing developers to use the __type metamethod mechanism.

[Rendered Link](https://github.com/metatablecat/rfcs/blob/type-metamethod/docs/type-metamethod-support.md)


vegorov-rbx:
We specifically excluded support for __type override for the reasons mentioned in the drawbacks.
As such, we do not plan to accept this RFC.

alexmccord:
This also breaks type refinements as a feature.

vegorov-rbx:
It also break type guards in native codegen.
Injecting a prefix dynamically kills typeof performance and guaranteeing statically that type will have prefix and will not change is impossible because metatables are mutable.

alexmccord:
Yeah. The feature already exists in the form of tagged unions, use them.

PR #59, opened 2024-10-09
Fix table formatting in user-defined-type-functions.md

joonyoo181:
The table formatting in .md does not translate well to .html:

https://rfcs.luau-lang.org/user-defined-type-functions.html#types-api-reference

If you open the full `types` API Reference, you can see that .html dumped the raw text. This PR fixes this visual error by formatting our API Reference using table tags in HTML.

A downside is the amount of code needed to support this visual change. If someone were to modify the content of this Reference table/maintain the RFC, it may be tedious to read through all the tags. However, considering that merged RFCs are finalized, I don't think this poses much of an issue :)

### DIFF
--- a/docs/user-defined-type-functions.md
+++ b/docs/user-defined-type-functions.md
@@ -98,103 +98,380 @@ This section details the initial programming interface we propose for `type`s wh
 
 <details><summary>Expand for full type API reference.</summary>
 
-### `types` Library
+<h3>types Library</h3>
 
-| Library Properties | Type | Description |
-| ------------- | ------------- | ------------- |
-| `unknown` | `type` | an immutable instance of the built-in type `unknown` |
-| `never` | `type` | an immutable instance of the built-in type `never` |
-| `any` | `type` | an immutable instance of the built-in type `any` |
-| `boolean` | `type` | an immutable instance of the built-in type `boolean` |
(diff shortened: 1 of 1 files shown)


vegorov-rbx:
We use Kramdown for html generations and it supports tables: https://kramdown.gettalong.org/quickref.html#tables

I suggest fixing whatever syntax the markdown source has that is not being recognized.
I would prefer not to merge your PR with raw html.

joonyoo181:
From what I see, Kramdown does not seem to support having markdowns under HTML tags. For example,

```markdown
<details>

### HEADING 1
Here are more details for heading 1

</details>
```

does not parse well when converted to HTML.

I guess in that case we have two options:
1) if we want to keep the "collapsing" feature of this RFC, we need to convert the contents between <detail> tags to raw HTML (what this PR currently does)
2) remove the "collapsing" feature of this RFC (aka remove the details and the summary tag)

I am good with both options. Which one would you prefer?

vegorov-rbx:
Removing the collapsing header seems like the best option as that removes even more HTML.

joonyoo181:
Perfect, I've removed the HTML headers.

vegorov-rbx (approved):

vegorov-rbx:
Thank you.

PR #60, opened 2024-10-19
RFC: Named Function Expressions

jackdotink:
Adds named function expressions:

```luau
foo(function factorial(n)
	return if n == 1 then 1 else n * factorial(n - 1)
end)
```

[Rendered](https://github.com/jackdotink/rfcs/blob/master/docs/named-function-expressions.md).

### DIFF
--- a/docs/named-function-expressions.md
+++ b/docs/named-function-expressions.md
@@ -6,7 +6,7 @@ This proposes that the current function expressions, which are always anonymous 
 
 ## Motivation
 
-This would make some code more readable, enable luau to give function expressions names other than `<anonymous>` in stack traces and with `debug.info`, enable function expressions to be recursive (call themselves), and help in miscellaneous profiling and debugging tools.
+This would make some code more readable, enable Luau to give function expressions names other than `<anonymous>` in stack traces and with `debug.info`, enable function expressions to be recursive (call themselves), and help in miscellaneous profiling and debugging tools.
 
 ## Design
 
--- a/docs/named-function-expressions.md
+++ b/docs/named-function-expressions.md
@@ -1,0 +1,59 @@
+# Named Function Expressions
+
+## Summary
+
+This proposes that the current function expressions, which are always anonymous (unnamed) be allowed to have names.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


aviralg:
While I sympathize with the problem of not having callback names in the debugger, I am not in favor of solving it this way.

The following piece of code currently creates a closure and binds it to `factorial` in the global scope:

```
function factorial(n)
	return if n == 1 then 1 else n * factorial(n - 1)
end
```

Per your suggestion, if used in a function argument position, the same syntax will no longer create a binding to the closure in the global scope, but only in the closure's body. This introduces different semantics for the same syntax. 

Secondly, function definitions now also become expressions, which allows for weird syntax like this. The same function can  now be accessed with two names in the body but one name in the enclosing scope.

```
local foo = function factorial(n)
	return if n == 1 then 1 else n * factorial(n - 1)
end
```

In my opinion, this proposal introduces bad syntax and semantics.

alexmccord:
How about `@debugname`?

WheretIB:
It gets pretty complicated at the part where referring to the name is allowed.
That means it needs an allocation of a local inside an expression context which may already contain temporary values before it on the stack, something currently not defined to work reliably in the compiler.
It's also strange to have this local be mutable and requires introduction of potential upvalue close opcodes inside the argument list.
Similar issues are problematic in the proposals about block expressions.

jackdotink:
@aviralg 
> Per your suggestion, if used in a function argument position, the same syntax will no longer create a binding to the closure in the global scope, but only in the closure's body. This introduces different semantics for the same syntax.

While this is somewhat true, its also true to say that `1 + 1` is valid syntax in expression position, but is a syntax error in statement position.

> Secondly, function definitions now also become expressions, which allows for weird syntax like this. The same function can now be accessed with two names in the body but one name in the enclosing scope.

This specific example is false. In this example, `local foo = function() foo() end` calling `foo` is a runtime error of `attempt to call a nil value`. Other examples do exist yes, but they're more contrived and already possible.

@WheretIB 
> It gets pretty complicated at the part where referring to the name is allowed.

If this truly is not possible I would like to open an RFC for table function fields, so please let me know.

jackdotink:
There has not been an update in a couple weeks and the last I heard this was unreasonable to implement. I'll close this now.

PR #62, opened 2024-10-22
Add 'Status: Implemented' to require-by-string RFCs

vrn-sn:
Updating require-by-string RFCs, as Luau's CLI tool now supports all aspects of these RFCs.


JohnnyMorganz:
Is this true for luau-analyze? Or is that not considered wrt implementation status

vrn-sn:
Ah, good point. I'll reopen this PR once the alias support is merged in for `luau-analyze`.

bradsharp (approved):

PR #63, opened 2024-10-26
RFC: switch for statements

dumpstring:
[switch for statements](https://github.com/dumpstring/rfcs/blob/master/docs/switch-for-statements.md)

### DIFF
--- a/docs/switch-for-statements.md
+++ b/docs/switch-for-statements.md
@@ -1,0 +1,55 @@
+# switch for statements
+
+## Summary
+
+This RFC proposes a new `switch` syntax for the Luau language, adding `switch` statements for clearer, more concise branching logic.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


checkraisefold:
another alternative could be something similar to rust match statement?

nothing1649:
not really sure how i feel about using existing for/do keywords in this context, at first glance i thought that these were loops and not case statements

checkraisefold:
Actually I feel like using `for` and `do` is very ambiguous; what happens if you want to run a for loop or use a `do end` inside the switch case? I think that becomes very hard to interpret, and maybe hard to parse. There definitely needs to be an alternative keyword here

alexmccord:
Parsing a `match` or `switch` statement with the syntax one normally would expect, is undecidable in Luau's current syntax. Nothing gives any piece of information implying that we're, with great certainty, parsing a `match` or whatever.

Sorry.

alexmccord (commented):
Decided to give my opinions even though it's undecidable to parse it.

deviaze:
Luau could benefit from an expressive, Rust-like `match` expression syntax, but I don't see this feature as proposed being helpful enough to provide benefit to Luau users. `switch` statements with fallthrough aren't a great part of C, so I don't think bringing them as-is is a good idea for Luau.

PR #66, opened 2024-11-04
RFC: buffer swap variable byte range

00826:
[Rendered](https://github.com/00826/buffer-swap-rfc/blob/master/docs/buffer-swap-rfc.md)!

### DIFF
--- a/docs/buffer-swap-rfc.md
+++ b/docs/buffer-swap-rfc.md
@@ -1,0 +1,44 @@
+# buffer.swap(b: buffer, x: number, y: number, range: number?)
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


zeux:
`buffer.copy` with a single temporary buffer will allow avoiding the use of strings, and reduce the waste here to an extra allocation, which can be shared between multiple swap calls depending on implementation. I think this is preferable to adding a rarely useful function to the standard library, which also doesn't have an equivalent for table, or indeed in C standard library.

00826:
thank you mr zeux for the insight! very much appreciated

PR #69, opened 2024-11-06
RFC: Metatable type functions

ffrostfall:
[Rendered](https://github.com/ffrostfall/rfcs/blob/metatable-type-functions/docs/metatable-type-functions.md)

### DIFF
--- a/docs/metatable-type-functions.md
+++ b/docs/metatable-type-functions.md
@@ -72,7 +72,7 @@ type MetatableType = getmetatable<setmetatable<{}, metatable>>
 ```
 
 If the type passed into the type function doesn't have a metatable and it *could* have a metatable (so primitives, singletons, classes, and tables), then it will return nil.
-If the type passed in is an intersection/union, then it will return an interesction/union respectively.
+If the type passed in is an intersection or union, then it will return an intersection or union respectively.
 
 ## Drawbacks
 
--- a/docs/metatable-type-functions.md
+++ b/docs/metatable-type-functions.md
@@ -32,7 +32,7 @@ type Identity = setmetatable<{ time: number }, { __index: clock }>
 
 In the following code example, `Identity` should evaluate to `{ sound: string, @metatable: { __index: animal } }`:
 
-```lua
+```luau
 local animal = {}
 type Identity = setmetatable<{
 	sound: string
--- a/docs/metatable-type-functions.md
+++ b/docs/metatable-type-functions.md
@@ -22,6 +22,8 @@ local clock = {}
 type Identity = setmetatable<{ time: number }, { __index: typeof(clock) }>
 ```
 
+If the second argument of the type function isn't a table, then it should error.
+
 ### `getmetatable` Type Function
 
 [Issue #1435](https://github.com/luau-lang/luau/issues/1435) is caused by a lack of special behavior for `getmetatable` when the type's metatable has a `__metatable` field. This is fixable in a variety of ways, however a type function has multiple benefits and exposes more type-level expression for users. It fits in nicely with the new solver, as type functions are a cleaner alternative to magic functions.
--- a/docs/metatable-type-functions.md
+++ b/docs/metatable-type-functions.md
@@ -45,7 +45,7 @@ type Identity = setmetatable<{
 
 In the following code example, `ReversedIdentity` should evaluate to `{ __index: animal }`:
 
-```lua
+```luau
 local animal = {}
 type Identity = setmetatable<{
 	sound: string
--- a/docs/metatable-type-functions.md
+++ b/docs/metatable-type-functions.md
@@ -74,4 +74,6 @@ The name of `setmetatable` has an interpretation of implying a side effect; tabl
 
 ## Alternatives
 
+Do nothing, and rely on unique metatable type syntax to achieve what a `setmetatable` type function would achieve. However, this still has the issue of `getmetatable`. It is flat out incorrect at the moment, and a type function would be an ideal solution to that problem. If a type function is introduced, then any user could reasonably expect a counterpart `setmetatable` type function. Both metatable type functions and metatable syntax can exist at the same time, and that might be ideal.
+
 Do nothing. `typeof` solves the current issues with accessibility to the current object-oriented programming patterns, and the issue relating to getmetatable can be solved by overloads.
--- a/docs/metatable-type-functions.md
+++ b/docs/metatable-type-functions.md
@@ -17,7 +17,7 @@ type Identity = typeof(setmetatable({} :: { time: number }, { __index = clock })
 
 could be reduced to:
 
-```lua
+```luau
 local clock = {}
 type Identity = setmetatable<{ time: number }, { __index: clock }>
 ```
--- a/docs/metatable-type-functions.md
+++ b/docs/metatable-type-functions.md
@@ -10,7 +10,7 @@ Implement type functions for `getmetatable` and `setmetatable`.
 
 There is currently no way for users to apply metatable type information to a type without the usage of `typeof()`. This isn't ideal, as it adds verbosity and boilerplate to common patterns such as object-oriented programming. For example, the following:
 
-```lua
+```luau
 local clock = {}
 type Identity = typeof(setmetatable({} :: { time: number }, { __index = clock }))
 ```
--- a/docs/metatable-type-functions.md
+++ b/docs/metatable-type-functions.md
@@ -44,7 +44,7 @@ type Identity = setmetatable<{
 ```
 
 When a non-table type (number, boolean) is passed into the type function, it should result in an error. This includes classes.
-If the type is an intersection/union of tables, then the metatable should be applied to every 
+If the type is an intersection or union of tables, then the metatable should be applied to every component type.
 
 ### `getmetatable` Type Function
 
--- a/docs/metatable-type-functions.md
+++ b/docs/metatable-type-functions.md
@@ -58,7 +58,7 @@ type ReversedIdentity = getmetatable<Identity>
 
 Due to `__metamethod`, additional behavior needs to be met. In the following code example, `MetatableType` should evaluate to `"No metatable here!"`:
 
-```lua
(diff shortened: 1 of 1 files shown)


docs/metatable-type-functions.md
```diff
@@ -0,0 +1,79 @@
...
+}, {
+	__index: animal
+}>
+```
+
+### `getmetatable` Type Function
+
+In the following code example, `ReversedIdentity` should evaluate to `{ __index: animal }`:
+
+```lua
```
aatxe, line 48:
```suggestion
```luau
```

docs/metatable-type-functions.md
```diff
@@ -0,0 +1,79 @@
...
+}, {
+	__index: animal
+}>
+
+type ReversedIdentity = getmetatable<Identity>
+```
+
+Due to `__metamethod`, additional behavior needs to be met. In the following code example, `MetatableType` should evaluate to `"No metatable here!"`:
+
+```lua
```
aatxe, line 61:
```suggestion
```luau
```

docs/metatable-type-functions.md
```diff
@@ -0,0 +1,79 @@
...
+
+Implement type functions for `getmetatable` and `setmetatable`.
+
+## Motivation
+
+### `setmetatable` Type Function
+
+There is currently no way for users to apply metatable type information to a type without the usage of `typeof()`. This isn't ideal, as it adds verbosity and boilerplate to common patterns such as object-oriented programming. For example, the following:
+
+```lua
```
aatxe, line 13:
```suggestion
```luau
```

docs/metatable-type-functions.md
```diff
@@ -0,0 +1,79 @@
...
+local clock = {}
+type Identity = typeof(setmetatable({} :: { time: number }, { __index = clock }))
+```
+
+could be reduced to:
+
+```lua
```
aatxe, line 20:
```suggestion
```luau
```

docs/metatable-type-functions.md
```diff
@@ -0,0 +1,79 @@
...
+
+[Issue #1435](https://github.com/luau-lang/luau/issues/1435) is caused by a lack of special behavior for `getmetatable` when the type's metatable has a `__metatable` field. This is fixable in a variety of ways, however a type function has multiple benefits and exposes more type-level expression for users. It fits in nicely with the new solver, as type functions are a cleaner alternative to magic functions.
+
+## Design
+
+### `setmetatable` Type Function
+
+In the following code example, `Identity` should evaluate to `{ sound: string, @metatable: { __index: animal } }`:
+
+```lua
```
aatxe, line 35:
```suggestion
```luau
```

docs/metatable-type-functions.md
```diff
@@ -0,0 +1,79 @@
...
+type Identity = setmetatable<{ time: number }, { __index: clock }>
+```
+
+### `getmetatable` Type Function
+
+[Issue #1435](https://github.com/luau-lang/luau/issues/1435) is caused by a lack of special behavior for `getmetatable` when the type's metatable has a `__metatable` field. This is fixable in a variety of ways, however a type function has multiple benefits and exposes more type-level expression for users. It fits in nicely with the new solver, as type functions are a cleaner alternative to magic functions.
+
+## Design
+
+### `setmetatable` Type Function
```
aatxe, line 31: What is the semantics of `setmetatable` when the first argument is not a table (or union/intersection of tables)? Presumably, reduction failure? This should be specified.

docs/metatable-type-functions.md
```diff
@@ -0,0 +1,79 @@
...
+```lua
+local animal = {}
+type Identity = setmetatable<{
+	sound: string
+}, {
+	__index: animal
+}>
+```
+
+### `getmetatable` Type Function
```
aatxe, line 44: What is the semantics of `getmetatable` when the argument does not have a metatable? What about when it cannot have one, e.g. because it is neither a table nor class? What does it do when given a union of tables? Should it check if the tables all have the same metatable and return the one or error otherwise? Should it return a union of the metatables across the union?

aatxe (changes requested):

docs/metatable-type-functions.md
aatxe: My general impression is that it makes the most sense to air conservatively here, we can always specify more behavior later. Possible answer: `getmetatable` will work on a table, and a class, and give you their respective metatables, but nothing else, and will respect `__metatable`.

docs/metatable-type-functions.md
```diff
@@ -0,0 +1,79 @@
...
+### `setmetatable` Type Function
+
+In the following code example, `Identity` should evaluate to `{ sound: string, @metatable: { __index: animal } }`:
+
+```lua
+local animal = {}
+type Identity = setmetatable<{
+	sound: string
+}, {
+	__index: animal
```
aatxe, line 40: Do you mean for this to be `typeof(animal)`? Or for `animal` to be a type?

docs/metatable-type-functions.md
```diff
@@ -0,0 +1,79 @@
...
+### `getmetatable` Type Function
+
+In the following code example, `ReversedIdentity` should evaluate to `{ __index: animal }`:
+
+```lua
+local animal = {}
+type Identity = setmetatable<{
+	sound: string
+}, {
+	__index: animal
```
aatxe, line 53: Do you mean for this to be `typeof(animal)`? Or for `animal` to be a type?

aatxe (changes requested):

docs/metatable-type-functions.md
ffrostfall: Yeah, I did. Oops

docs/metatable-type-functions.md
```diff
@@ -0,0 +1,79 @@
...
+```luau
+local clock = {}
+type Identity = typeof(setmetatable({} :: { time: number }, { __index = clock }))
+```
+
+could be reduced to:
+
+```luau
+local clock = {}
+type Identity = setmetatable<{ time: number }, { __index: clock }>
```
aatxe, line 22: Did you mean to use `typeof(clock)` here?

docs/metatable-type-functions.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+could be reduced to:
+
+```luau
+local clock = {}
+type Identity = setmetatable<{ time: number }, { __index: typeof(clock) }>
+```
+
+If the second argument of the type function isn't a table, then it should error.
```
aatxe, line 25: This should be in the design section.

docs/metatable-type-functions.md
```diff
@@ -0,0 +1,87 @@
...
+```luau
+type metatable = {
+	__metatable: "No metatable here!"
+}
+
+type MetatableType = getmetatable<setmetatable<{}, metatable>>
+```
+
+If the type passed into the type function doesn't have a metatable and it *could* have a metatable (so primitives, singletons, classes, and tables), then it will return nil.
+If the type passed in is an intersection/union, then it will return an interesction/union respectively.
```
aatxe, line 75:
```suggestion
If the type passed in is an intersection or union, then it will return an intersection or union respectively.
```

docs/metatable-type-functions.md
```diff
@@ -0,0 +1,87 @@
...
+local animal = {}
+type Identity = setmetatable<{
+	sound: string
+}, {
+	__index: typeof(animal)
+}>
+```
+
+When a non-table type (number, boolean) is passed into the type function, it should result in an error. This includes classes.
+If the type is an intersection/union of tables, then the metatable should be applied to every
```
aatxe, line 47: Stopped writing midsentence?
```suggestion
If the type is an intersection or union of tables, then the metatable should be applied to every component type.
```

aatxe (changes requested):

aatxe (approved):

docs/metatable-type-functions.md
```diff
@@ -0,0 +1,86 @@
...
+type Identity = setmetatable<{
+	sound: string
+}, {
+	__index: typeof(animal)
+}>
+
+type ReversedIdentity = getmetatable<Identity>
+```
+
+Due to `__metamethod`, additional behavior needs to be met. In the following code example, `MetatableType` should evaluate to `"No metatable here!"`:
```
vegorov-rbx, line 63:
```suggestion
Due to `__metatable`, additional behavior needs to be met. In the following code example, `MetatableType` should evaluate to `"No metatable here!"`:
```

docs/metatable-type-functions.md
```diff
@@ -0,0 +1,86 @@
...
+```luau
+type metatable = {
+	__metatable: "No metatable here!"
+}
+
+type MetatableType = getmetatable<setmetatable<{}, metatable>>
+```
+
+If the type passed into the type function doesn't have a metatable and it *could* have a metatable (so primitives, singletons, classes, and tables), then it will return nil.
+If the type passed in is an intersection or union, then it will return an intersection or union respectively.
```
aatxe, line 74:
```suggestion
If the type passed in does not support metatables in the type system (i.e. at the time of writing, is not any of a primitive, a singleton, a class, a table, or a union or intersection thereof), it should error when passed to `getmetatable`. 
If the type passed in is an intersection or union of things that support metatables, then it will return an intersection or union of those metatables respectively.
If no metatable is present, but the type system supports metatables for that type, it should return the `nil` type indicating that there is no metatable present.
```

vegorov-rbx:
If `__metatable` handling is specified for `getmetatable` to match the runtime better (where it returns a field), should `setmetatable` also specify the behavior when called on a type with `__metatable` present? (So, error that metatable cannot be set).

ffrostfall:
> If `__metatable` handling is specified for `getmetatable` to match the runtime better (where it returns a field), should `setmetatable` also specify the behavior when called on a type with `__metatable` present? (So, error that metatable cannot be set).

I think it should. The purpose of types is to prevent errors, so...

PR #70, opened 2024-11-06
RFC: Unique syntax for metatable types

ffrostfall:
[Rendered](https://github.com/ffrostfall/rfcs/blob/unique-metatable-type-syntax/docs/metatable-type-syntax.md)

### DIFF
--- a/docs/metatable-type-syntax.md
+++ b/docs/metatable-type-syntax.md
@@ -6,7 +6,7 @@ Add unique syntax to allow for specifying the metatable of a type.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


Wunder-Wulfe:
I believe this would be very helpful for manually defining classes without using this structure:
```luau
type Class = setmetatable<
    {
       KeyType: ValueType;
    },
    classMeta
>;
```
Though, the way metatables are currently presented to the user in the type engine is with `@metatable`, so it may be more convenient to copy that syntax as well:
image
```luau
type Class = {
    KeyType: ValueType;
    @metatable classMeta;
}
```

aatxe:
The main concern about `@metatable` would be conflicting with possible additions for attributes on types.

PR #71, opened 2024-11-09
RFC: Allow `xpcall` error handlers to yield

ccuser44:
RFC: Allow `xpcall` error handlers to yield

[Rendered](https://github.com/ccuser44/rfcs/blob/patch-1/docs/xpcall-error-handler-yield.md)

### DIFF
--- a/docs/xpcall-error-handler-yield.md
+++ b/docs/xpcall-error-handler-yield.md
@@ -1,0 +1,41 @@
+# Feature name
+
+## Summary
+
+Allow `xpcall` error handlers to yield.
+
+## Motivation
+
+Lua 5.2+ `xpcall` can yield both the main function and error handlers. Luau can only yield the main function but not the error handler.
+
+The Luau website says that `xpcall` yielding semantics have been ported from Lua 5.2. However, this isn't exactly correct, only the main function can yield, such an oversight can lead to potential confusion with users. This also adds a layer of incompatibility with vanilla Lua that's not needed.
+
+The biggest issue is that the yield semantics are very un-intuitive, causing potential issues with novice users. Especially when the debug semantics aren't very user-friendly.
+
+Furthermore, if one were to use a feature that would require yielding in the error handler, they would have to implement some kind of event wrapped functionality to be able to use the desired feature or use a different alternative to `xpcall` altogether.
+
+## Design
+
+Allow the second argument of `xpcall`, which is the error handler function to yield.
+
+## Drawbacks
+
+Potential performance implications and increased complexity in call semantic handling. Probably very minor though.
+
+## Alternatives
+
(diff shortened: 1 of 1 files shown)


bradsharp:
An additional drawback is that a number of users likely use `xpcall` to prevent yielding of code. That said, this can likely be reimplemented with:
```
local function noyield(f, ...)
  local t = coroutine.create(f)
  local r = table.pack(coroutine.resume(t, ...))
  if coroutine.status(t) == "suspended" then
    coroutine.close(t)
    error("Method is not allowed to yield")
  end
  return table.unpack(r)
end

ccuser44:
> An additional drawback is that a number of users use `xpcall` to prevent yielding of code. That said, this can be reimplemented with:
> ```lua
> local function noyield(f, ...)
>   local t = coroutine.create(f)
>   local r = table.pack(coroutine.resume(t, ...))
>   if coroutine.status(t) == "suspended" then
>     coroutine.close(t)
>     error("Method is not allowed to yield")
>   end
>   return table.unpack(r)
> end
> ```

I see, that may be a problem. However I still absolutely think that the `xpcall` error handler should be able to yield.
First of all this behavior is totally undocumented. Relying on it is hacky and unreliable too as the yield behavior of both `xpcall` (in Luau) and `pcall` (before Luau when `ypcall` was deprecated).
Another is parity with vanilla Lua, the current behavior only partially mirrors that of vanilla Lua.
Also the debug semantics of not allowing the error handler to yield are quite un-user friendly.

bradsharp:
I'd be in favor of this but we need to carefully consider how we'd roll this out without impacting users.

vegorov-rbx:
I don't think it will be "Probably very minor though.", seems like a major increase in complexity of the VM where the whole Lua 5.2 implementation uses data structures incompatible with Luau VM.

Given limited use case and already existing yielding limitations, I don't think it will be worth adding this. Also don't want to accept an RFC to later revert as unimplementable.

Just because compatibility box in the docs in incorrect, doesn't mean we have to change Luau, we can just note the difference in the documentation correctly.

zeux:
+1 for just revising the docs. Yielding in xpcall error handler is non-trivial with the current implementation to the extent I remember it, and rewriting yield support to match Lua 5.2 is out of the question for many reasons (performance being one of them, but also just a lot of complexity; Luau implementation diverges from Lua 5.2 for a reason).

I also think xpcall is a "framework" utility function, not an app utility function - it's somewhat dangerous (eg xpcall error handler can't really error from a logical perspective), and should be wielded with care. Regular user code should not use xpcall.

vegorov-rbx:
Closed by an update to our compatibility table as the concern about the behavior difference is valid https://github.com/luau-lang/site/pull/36

PR #72, opened 2024-11-11
RFC: Result type function

gaymeowing:
[Rendered](https://github.com/witchiestwitchery/rfcs/blob/master/docs/result-type-operator.md)

### DIFF
--- a/docs/result-type-operator.md
+++ b/docs/result-type-operator.md
@@ -103,7 +103,7 @@ local success, result = pcall(mrow, "cat food", ":3")
 
 ## Alternatives
 
-Allow for type pack unions to be written by developers, with the syntax probably looking like this:
+Allow for type-pack unions to be written by developers, with the syntax probably looking like this:
 ```luau
 local function mrow(meow: string, mrrp: string): (true, string) | (false, nil)
 	-- code here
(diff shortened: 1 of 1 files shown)


docs/result-type-operator.md
```diff
@@ -0,0 +1,121 @@
...
+		return { ok = true, value = `{meow} {mrrp}` }
+	else
+		return { ok = false, value = nil }
+	end
+end
+```
+
+## Design
+
+The functionality of the `result` type function is special, with it being an exception as it'll create a type-pack union. Thus using the result type function will not require the developer to write an overloaded function type.
```
aatxe, line 69: Type pack unions do not exist in Luau. So, there's no way for a type pack function to produce a type pack union.

gaymeowing: Yes, but result would be special cased. As its the only use for typepack unions.

aatxe: There's not a way for us to do that right now. Type functions and type pack functions have to return types or type packs, and type packs are not unions. Doesn't mean we can't do this at all, but it does mean that this is essentially the same amount of work as proposing general support for type pack unions.

vegorov-rbx:
Going to close this, primarily because it depends on functionality of a type system which doesn't exist and is not defined.
While this can be re-introduced at a later point, if the language type system will have support for type pack unions, it will likely come with a way to define those for any situation and not specifically to how 'result' is defined here.

PR #73, opened 2024-11-14
Allow `table.clone` to copy tables with locked metatables

ccuser44:
Allow `table.clone` to copy tables with locked metatables.

[Rendered](https://github.com/ccuser44/rfcs/blob/patch-2/table.clone-locked-metatables.md)

### DIFF
--- a/table.clone-locked-metatables.md
+++ b/table.clone-locked-metatables.md
@@ -1,0 +1,21 @@
+# Feature name
+
+## Summary
+
+Allow `table.clone` to copy tables with locked metatables.
+
+## Motivation
+
+Currently `table.clone` cannot create shallow copies of a table if it has a locked metatable. The dictionary definition for shallow copying is to copy all the fields of a table to a new table but not doing so recursively. There is no mention of metatables in the first place so most users expect said behavior just to work, if they haven't taken a look at the manual of course.
+
+## Design
+
+When `table.clone` attempts to clone a table with a locked metatable the table is shallow copied with the exception of not assigning the metatable to the shallow copy.
+
+## Drawbacks
+
+There are no drawbacks to this. You can already create a shallow copy of a table with a locked metatable (without assigning the metatable to the copy) so there is no downside in allowing `table.clone` to do the same.
+
+## Alternatives
+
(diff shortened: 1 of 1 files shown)


vegorov-rbx:
This RFC seems to completely ignore the reasoning around not allowing locked metatable copy from the `table.clone` RFC: https://github.com/luau-lang/rfcs/blob/master/docs/function-table-clone.md

ccuser44:
> This RFC seems to completely ignore the reasoning around not allowing locked metatable copy from the `table.clone` RFC: https://github.com/luau-lang/rfcs/blob/master/docs/function-table-clone.md

It does not, you misunderstood what this RFC proposes. This RFC allows cloning tables that have locked metetables **by not applying the metatable to the clone if it's locked!**


The bahavior of `table.clone` would change from:

```lua
function table.clone(t)
assert(type(t) == "table")
local nt = {}

for k, v in pairs(t) do
nt[k] =v
end

if type(getmetatable(t)) == "table" then
setmetatable(nt, getmetatable(t))
end

return nt
end
```

to

```lua
function table.clone(t)
assert(type(t) == "table")
local nt = {}

for k, v in pairs(t) do
nt[k] =v
end

local mtLocked = getmetatable(t) ~= nil and not pcall(setmetatable, t, getmetatable(t))

if not mtLocked and type(getmetatable(t)) == "table" then
setmetatable(nt, getmetatable(t))
end

return nt
end
```

vegorov-rbx:
I don't see any mention of the decisions made in the original RFC in your RFC, thus you have ignored previous work.

Your text even has "There is no mention of metatables in the first place" while there is plenty of mention of the metatables in the `table.clone` RFC.
If you find that's something is missing in the documentation, documentation can be updated, not the behavior.

ccuser44:
> Your text even has "There is no mention of metatables in the first place" while there is plenty of mention of the metatables in the `table.clone` RFC.

If you had read my RFC properly at all you would have found out I said there was no mention of metatables in the dictionary definition of a shallow copy. I guess my wording was a bit too confusing.

vegorov-rbx:
I don't see what dictionary definition has to do with Luau semantics, which we can define in a way we want.
Do not submit such low effort RFCs in the future.

PR #74, opened 2024-11-19
Add RFC for 2-component vector constructor

petrihakkinen:
Add RFC for a new vector constructor with two arguments. The constructor is especially useful for 2D graphics and UI code because it improves code ergonomics and efficiency of creating vectors.

[Rendered](https://github.com/petrihakkinen/rfcs/blob/petrih-vector2-constructor/docs/vector-library-vector2-constructor.md)

### DIFF
--- a/docs/vector-library-vector2-constructor.md
+++ b/docs/vector-library-vector2-constructor.md
@@ -1,0 +1,37 @@
+# 2-component vector constructor
+
+## Summary
+
+Add a constructor for initializing vectors from two components.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


vegorov-rbx:
Sounds reasonable.
The meat of the proposal seems to be just accepting two arguments.

I don't particularly like mentions of implementation-defined optimizations as a requirement in the body of the RFC.
Different Luau implementation might not even have fastcalls and we didn't ever specify existing vector constants as an RFC before and tricks like LBF_BIT32_EXTRACTK can be implemented if we find that useful without going through an RFC.
I would relax that wording to the motivation section as a motivation that faster things can be done, not that they must be done.

petrihakkinen:
Makes sense. I will relax the wording on implementation details.

petrihakkinen:
Thanks for the approval. I can start working on the implementation this week or next, unless you plan to work on this internally at Roblox? Does this need a feature flag (although it's a non-breaking change)?

vegorov-rbx:
I will not be able to start on this in the coming weeks.

The change will require a feature flag.

PR #75, opened 2024-11-19
Add RFC for shorter vector constructor

petrihakkinen:
Add RFC for replacing the current vector constructor (vector.create) with a less verbose version (vector).

[Rendered](https://github.com/petrihakkinen/rfcs/blob/petrih-shorter-vector-constructor/docs/vector-library-shorter-constructor.md)

### DIFF
--- a/docs/vector-library-shorter-constructor.md
+++ b/docs/vector-library-shorter-constructor.md
@@ -1,0 +1,33 @@
+# Shorter vector constructor
+
+## Summary
+
+Replace the current vector constructor `vector.create` with a less verbose, more ergonomic `vector`.
+
+## Motivation
+
+While the current vector constructor `vector.create` is consistent with other functions in the vector library, it is quite verbose. A shorter version would improve the readability and writability of code.
+
+Especially scripts that contain a lot of data such as positions, directions, matrices or colors could be significantly less verbose with a shorter syntax.
+
+## Design
+
+The `__call` metamethod in the vector library table will be set to the already existing vector constructor. Thus calls to `vector` will create a new vector value. Any existing code that uses `vector.create` can swap in `vector` with identical behavior.
+
+The new constructor `vector` shall be backed with a built-in fastcall so that `vector` and `vector.create` will have identical performance characteristics. As a result vectors will also be added to the constant table by the compiler when the new shorter version of `vector` is used.
+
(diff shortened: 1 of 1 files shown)


petrihakkinen:
After a long think I decided to propose the deprecation and eventual removal of `vector.create` in favor of `vector` in the RFC because I think it would be better for the language in the long run to have only one constructor. If you think this is a too big breaking change we can modify the RFC to keep the old version as an alias. I don't have a strong opinion but we'd prefer some officially sanctioned way to disable `vector.create` (or any other built-in) locally for our codebase in case `vector` is added but `vector.create` is not deprecated.

jackdotink:
Personally I would like to see special syntax for creating vectors, such as `<x, y, z>`.

vegorov-rbx:
I don't think we will be adopting this.
Reasoning for current constructor was already discussed in the vector library RFC and we have selected that option to go forward.

In a way, this RFC was already rejected in the discussion of that RFC.

alexmccord:
> I don't think we will be adopting this. Reasoning for current constructor was already discussed in the vector library RFC and we have selected that option to go forward.
> 
> In a way, this RFC was already rejected in the discussion of that RFC.

I don't think this accurately conveys what actually happened with the syntax for vector creation. We talked about a whole bunch of them:

1. `vector(x, y, z)`
2. `[x, y, z]` following usual notation in linalg
3. `<x, y, z>`
4. `|x, y, z|`

I feel a strong urge to mention that I never liked the last two, but I digress.

The reason why we split out the vector creation shorthand from the RFC was to allow it to go through, _and then_ let someone talk about the shorthand separately if they're annoyed by `vector.create(x, y, z)` being the way to create them.

So I don't think it's fair to automatically reject this RFC.

vegorov-rbx:
I did not mean to reject it automatically, I reject it based on what is proposed.

bradsharp:
Also want to mention `table.create` as additional art.

ccuser44 (approved):

Wunder-Wulfe:
This could be enhanced further depending on shorthand usage, examples:
```luau
v = vector.create; -- or vec = vector.create
print(v(1,2,3) + v(3,2,1));
```

vegorov-rbx:
Closing this one as there was no consensus to accept the RFC as it stands today.
Alternatives can still be explored in the future.

PR #80, opened 2024-11-23
RFC: Absolute Require Syntax

bradsharp:
This RFC proposes additional syntax for require by string that allows modules to be resolved relative to the root of a project or library. It is based on #78 which proposed the same changes but also removed relative paths.

[Rendered](https://github.com/luau-lang/rfcs/blob/rfc-absolute-require/docs/absolute-requires.md).

### DIFF
--- a/docs/absolute-requires.md
+++ b/docs/absolute-requires.md
@@ -1,8 +1,8 @@
-# Require Resolution with Libraries
+# Absolute Requires
 
 ## Summary
 
-This proposal describes a new syntax to support require-by-string in a format that can work for filesystems and embedded applications such as the Roblox platform, replacing the existing syntax we have today.
+This proposal describes a new absolute syntax to support require-by-string in a format that can work for filesystems and embedded applications such as the Roblox platform, replacing the existing syntax we have today.
 
 ## Motivation
 
@@ -77,11 +77,9 @@ However, one subtle change is that you won't be allowed to require init.luau dir
 
 There are no changes to aliases in this proposal.
 
-### Relative Paths
+### Comparison with Relative Paths
 
-Support for relative paths will be dropped as any libraries written with relative paths would still be incompatible for the aforementioned reasons.
-
-### Comparison with Current Behavior
+We will continue to support relative paths, however recommend updating to absolute paths to guarantee portability between different environments.
 
 ```
 ├── folder
@@ -94,7 +92,7 @@ Support for relative paths will be dropped as any libraries written with relativ
 └── init.luau
 ```
 
-| **From**          | **To**                   | **Current Syntax**                              | **Proposed Syntax**  |
+| **From**          | **To**                   | **Relative Path**                               | **Absolute Path**    |
 |-------------------|--------------------------|-------------------------------------------------|----------------------|
 | init.luau         | module.luau              | ./module                                        | module               |
 | init.luau         | folder/init.luau         | ./folder                                        | folder               |
@@ -106,11 +104,11 @@ Support for relative paths will be dropped as any libraries written with relativ
 | folder/init.luau  | library/folder/file.luau | ./../library/folder/file ../library/folder/file | library/folder/file  |
 | library/init.luau | library/folder/file.luau | ./folder/file                                   | folder/file          |
 
-As paths are now relative to the root directory the only ones that change are those in files under a subdirectory. All other paths remain the same as the current syntax.
+As paths are relative to the root directory with the absolute syntax the only ones that change are those in files under a subdirectory. All other paths remain the same as the current syntax.
 
 ## Drawbacks
 
-While still fairly new, some users have already adopted the current syntax, which will mean they need to update their code to use the proposed syntax. We could maintain support for relative paths however this would be inadvisable due to the compatibility issue mentioned prior.
+As relative paths are still available, we haven't completely solved the problem. Users could still opt to use them, leading to the issue mentioned above. However, there may be options (such as tooling) to guide users away from relative paths over time to help address this.
 
 ## Alternatives
 
--- a/docs/require-by-string-2.md
+++ b/docs/require-by-string-2.md
@@ -1,0 +1,123 @@
+# Require Resolution with Libraries
+
+## Summary
+
+This proposal describes a new syntax to support require-by-string in a format that can work for filesystems and embedded applications such as the Roblox platform, replacing the existing syntax we have today.
+
+## Motivation
+
+It should be possible for all users of Luau to write code that is capable of running, without modification, on many different platforms. Unfortunately, we identified an issue with the current syntax that would make it difficult for users of certain tools to adopt. As paths are relative to the file calling require, any tools that adjust the file hierarchy would also need to update require paths to match. While we know there may not be a perfect solution that works for all use-cases, we think the current syntax should be amended to support these cases or at least not require them to translate require paths in files.
+
+## Design
+
+### Overview
+
+When calling require without a prefix, the path will be resolved as an absolute path where the root directory is defined as follows:
+
+* If the file does not belong to a library, then the entry point's directory
+* If the file belongs to a library, then the directory containing the library
+
+For example, here's a simple file system:
+
+```
+├── folder
+│   └── file.luau
+├── module.luau
+└── init.luau
+```
+
+And here's how you would navigate between files using the proposed syntax when init.luau is the entry point:
+
+| **From**         | **To**           | **Path**    |
+|------------------|------------------|-------------|
+| init.luau        | module.luau      | module      |
+| init.luau        | folder/file.luau | folder/file |
+| module.luau      | folder/file.luau | folder/file |
+| folder/file.luau | module.luau      | module      |
+
+With this syntax, paths are the same no matter where a file is in a project, meaning they can be restructured by the user or with tools, without creating any issues.
+
+### Libraries
+
+While this RFC does not call out the exact definition of a library, there will be a way to designate a directory as such. Here is an example of a filesystem including one:
+
+```
+├── library (defined in some way)
+│   ├── folder
+│   │   └── file.luau
+│   └── module.luau
+└── init.luau
+```
+
+And here's how you would navigate between these files:
+
+| **From**                 | **To**                   | **Path**       |
+|--------------------------|--------------------------|----------------|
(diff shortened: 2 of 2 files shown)


docs/absolute-requires.md
```diff
@@ -0,0 +1,121 @@
...
+
+As paths are relative to the root directory with the absolute syntax the only ones that change are those in files under a subdirectory. All other paths remain the same as the current syntax.
+
+## Drawbacks
+
+As relative paths are still available, we haven't completely solved the problem. Users could still opt to use them, leading to the issue mentioned above. However, there may be options (such as tooling) to guide users away from relative paths over time to help address this.
+
+## Alternatives
+
+We have already considered (and even approved) some alternatives to this approach, however there are drawbacks with each of them that ultimately led us in this direction.
```
alexmccord, line 115: Can I see this section reworked? As it is, it requires you to read all the RFCs and filtering out irrelevant information wrt what this RFC is trying to solve. It's also not really "alternatives" either.

alexmccord:
Had to read all the related RFCs that was spawned in the last week surrounding `init.luau` and why it was problematic. This [comment](https://github.com/luau-lang/rfcs/pull/78#issuecomment-2494788964) explained the problem the best, and gave me a lot of context about why we're giving this issue any attention. Decided to collate all this information so that I can look at this problem clearly and what I think the resolution is.

Given this filesystem:

```
foo.luau
mod/
  foo.luau
  init.luau
```

With such contents:

```
% cat foo.luau
local m1 = require("./mod/init")
local m2 = require("./mod")
print(m1 == m2)
% cat mod/foo.luau
local m = {}
function m.id(x) return x end
return m
% cat mod/init.luau
return require("./foo")
```

Piping this through Rojo gives this DataModel structure:

```
foo
mod <- ModuleScript whose source is init.luau
  foo
```

To make matters worse, there is supposed to be a bijectivity between the require by string and require by DM paths:

| Prefix by string | Prefix by DM path
|------------------|-------------------
| `./`             | `script`
| `../`            | `script.Parent`
| `@mod/`          | `game:GetService("ServerScriptService"):FindFirstChild("mod")`

**This presents a counterexample**: `require(script.foo) != require("./foo")` due to Rojo's flattening, which implicitly causes `./` in the DM to actually function as `../` in the filesystem.

Consequently `./` in `init.luau` is fixed relative to the directory that contains `mod` instead of `mod` folder itself. I can now see how the bijectivity property between the filesystem and the DataModel are broken, and that it is caused by Rojo flattening a folder with `init.luau` into a `ModuleScript` that takes the content of `init.luau`.

Following this line of thinking, you can see another diverging behavior in the other direction: `foo.luau` cannot require `mod/init.luau` because `mod/init.luau` does not exist in the DataModel, due to Rojo flattening this structure.

In Lune, if you have the above filesystem structure and its contents and you run `foo.luau`, it prints `true`. Running this through the DataModel, you get an error saying that `foo.luau` could not require `mod/init.luau` because it doesn't exist.

This leads me to conclude that it might have been a benign design bug in Rojo's mapping between filesystem <-> DM. I say benign because it has always existed and was never shown to be a problem _until now_, just like `if x == nill then y else z` with the intent of checking if `x == nil`: it was not a problem until Luau got a type system that flags for unknown globals like `nill`. The same can be said here too, Rojo's mapping feature was designed _for_ that world where require by string did not exist, but that doesn't mean Rojo can't change.

The question I have now: is this a language design issue, or is this an issue elsewhere? I don't want to want to jump to the conclusion about whose issue it is, but it doesn't feel like a language issue to me.

### Perspective: Luau needs to change something

If you adopt the view that the language has to change to fix the problem, you get these RFCs:

1. [RFC 76](https://github.com/luau-lang/rfcs/pull/76) proposes to redefine `./` in any `init.luau` to work one level up in the filesystem and adds a new prefix (`self` or `script` or whatever) whose semantics are equivalent to `require(script.foo)`.
2. [URI style strings](https://github.com/luau-lang/rfcs/pull/76#issuecomment-2489623405) which is pre-RFC, but it takes the view that the language has to change to fix the problem.
3. [File + directory](https://github.com/luau-lang/rfcs/pull/76#issuecomment-2495292328) is essentially the same as RFC 76, only that `init.luau` has the same name as the directory, and sits as sibling to the directory.
4. [RFC 78](https://github.com/luau-lang/rfcs/pull/78) proposed to add absolute paths relative to the root directory, and also proposed to remove relative paths. A nuclear RFC.
5. [RFC 79](https://github.com/luau-lang/rfcs/pull/79) proposes to basically freeze all changes to require by strings. A nuclear RFC.
6. [RFC 80](https://github.com/luau-lang/rfcs/pull/80) proposes the same as RFC 78, minus the proposal to remove relative paths. Also a nuclear RFC, but to a lesser degree than the previous two.

The last 3 RFCs listed here got me looking at this and think about what I could do to step in and defuse the situation.

My stance is that none of the above fixes the problem where the filesystem and the DM are observationally different.

### Perspective: Rojo needs to change something

If you shift perspective and think that Rojo should change to fix the problem, you will be able to see better progression towards a real resolution by adding a constraint where the filesystem and the DM must be **structure preserving** to maintain bijection.

Currently, it conflates _flattening_ with _mapping_, where a folder containing `init.luau` is flattened into a ModuleScript _without_ `init.luau`. This is what causes the relative prefixes to work one level up in the DM as opposed to the intended level.

By removing the flattening part out of Rojo, you end up with this structure, exactly equivalent to one in the filesystem.

```
foo
mod <- Folder instance
  foo
  init
```

Way better.

Indeed, we have two issues that arises from doing this:

1. Require by DM paths can no longer require `mod` itself. I consider this to be a _missing feature_ on Roblox's behalf.
2. Fixes relative require by strings using `./` and `../`, but breaks relative requires by `script`. This one appears paradoxical, but can also be resolved on Roblox's behalf.

This means Rojo was bludgeoned into a corner, and only one way out left: Roblox.

### Perspective: Roblox needs to change something

If you again shift perspective and think that Roblox is missing something, then you can  ponder why flattening was added in Rojo in the first place.

The reason was purely a syntactic convenience. Without it, you had to settle on a convention on which script was the entry point for a collection of modules. You would also have to deal with one extra `.Parent` in DM paths that were already verbose to begin with. This motivates Rojo to want some way to rearrange the DM for your convenience.

This boils down to a missing instance type, call it `Library`. It has the following interactions:

1. If `require` resolves to a `Library` instance, resolve to the `init` script within the library. Both require by string and require by DM paths should do this for `Library` instances.
2. A ModuleScript named `init` that is a child of `Library` has `script` bound to the `Library` instance itself, rather than the `init` ModuleScript.

This effectively does the flattening _for_ Rojo, without making flattening a _destructive_ operation done by Rojo, allowing Rojo to represent a full and faithful bijection of the original filesystem used to map over into Roblox for the purpose of resolving requires without any observable differences.

With a `Library` instance, we'd end up with this:

```
foo
mod <- Library instance
  foo
  init
```

Rojo no longer has to flatten this out, since Roblox supports it first-class, and restores all backward incompatibilities that was not resolvable with `Folder`. Only at this point can Rojo remove the flattening and change how folders with `init.luau` are mapped over to the DM.

This feature exists with `require(id)` using `MainModule` as the entry point, but does not exist in the DM paths variant. This isn't anything unique, even require by string has this feature. It's the require by DM paths that's the oddball here.

### The Proposal

I think we should take this issue as a growing pain _towards_ having a `Library` instance whose purpose is to represent what the filesystem looks like in a structure preserving way. It should get prioritized, and as soon as it is released, Rojo can switch over to that, resolving the require by string inconsistency once and for all.

Until such an instance exist, Rojo is literally incapable of doing anything. It can't change the way it maps the filesystem to the DM without breaking one require style or the other. This means that require by string and require by DM paths are _mutually exclusive_ to each other if you're using Rojo.

**Libraries gives Rojo a way out of this mutual exclusion problem.**

Then it becomes obvious that all RFCs proposing to change the language can be thrown out, closed, and forgotten about. All of this just stems from the lack of confluence in semantics between each _style_ of requires, which needed to be resolved.

jackdotink:
The idea that Roblox could change to support the filesystem is indeed a nice idea, but you also have to consider the users. To some extent you can rely on the "if we build it they will come" mentality, but you have to remember this is a convention that has existed in Roblox for over a decade.

Is it really worth breaking that convention, and the way Rojo syncs projects (which will make string requires unusable without significant modification to existing Rojo projects), to better align with the file system? I don't think that it is.

I think a better idea is to change the way we think about the file system to instead match how the datamodel functions. This is helpful for Rojo, as it won't have to change, but also helpful for Roblox when they create their own tooling to put projects in the filesystem.

alexmccord:
Yeah, I checked in with Dekkonot and when you add in a few other edge cases that Rojo has to account for, the libraries suggestion is not enough, and there's no good solution other than to give up the view that the DM could model the filesystem faithfully. I'm beginning to think that URIs are a good idea as long as it doesn't interact with the language design, because this is still a Roblox-ism.

vegorov-rbx:
> 1. Require by DM paths can no longer require `mod` itself. I consider this to be a _missing feature_ on Roblox's behalf.

This is not correct, requiring a Folder instance will require 'folder/init' module in the current Roblox implementation.

Changing the value of `script` is likely not possible, for internal reasons.

vegorov-rbx:
> This works using require by string. Not by DM paths, which was what the sentence says.

Thanks for clearing that up, I incorrectly assumed it talked about string path in a DM.

Dekkonot:
Happy new year! Let's get back to arguing about require by string! I'll leave the discourse at the door though, because it doesn't really help anyone ha ha. :wink:

I cannot support an RFC that's contingent upon a feature that doesn't exist. Libraries sound like a great idea conceptually, but unless there's an RFC defining what one _is_, I don't think it's reasonable to merge anything that mentions them. Especially not this one: this RFC doesn't make any sense because the entire idea of a root is defined relative to libraries. I know that this is a draft right now, but it's a draft that says nothing other than "libraries solve this" without addressing what libraries are.

I'd also like to get it down on paper: I'm not willing to change Rojo over this unless I can be convinced it's the only way, and I remain unconvinced. It's not particularly relevant to this RFC outside of what @alexmccord said, but I think it's worth saying just so people quit thinking it. I don't carry veto power for Rojo's structure, so take it with a grain of "I may be overruled" but if I had my way we would not change.

PR #81, opened 2024-11-23
RFC: Do not change init.luau or relative require semantics

ffrostfall:
[Rendered](https://github.com/ffrostfall/rfcs/blob/master/docs/do-not-change-init.luau-or-relative-require-semantics.md)


dphfox:
Dupe of #79?

ffrostfall:
Oh, true. I didn't notice that.

PR #83, opened 2024-11-29
RFC: Add a `@mustuse` function attribute

CompeyDev:
This RFC proposes a `@mustuse` attribute, which may be applied to a function, enforcing the usage of a function's return value.

[Rendered](https://github.com/CompeyDev/luau-rfcs/blob/rfc-mustuse-attribute/docs/syntax-attribute-functions-mustuse.md).

### DIFF
--- a/docs/syntax-attribute-functions-mustuse.md
+++ b/docs/syntax-attribute-functions-mustuse.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-This RFC proposes a `@mustuse` attribute, which may be applied at the type or function level, that enforces the usage of a function's return value. 
+This RFC proposes a `@mustuse` attribute, which may be applied to a function, that enforces the usage of a function's return value. 
 
 If the return value is not used, a lint should be raised during analysis time. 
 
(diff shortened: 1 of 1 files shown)


dv-extrarius:
I know it's used in several languages, but to me "mustuse" is sort of confusing - it sounds like you have to use the function. In C++ there is the gcc attribute warn_unused_result and the standard nodiscard (don't discard the return value or you get a warning), both of which seem better to me

CompeyDev:
> I know it's used in several languages, but to me "mustuse" is sort of confusing - it sounds like you have to use the function. In C++ there is the gcc attribute warn_unused_result and the standard nodiscard (don't discard the return value or you get a warning), both of which seem better to me

I would also prefer `@nodiscard` on second thought, since `@mustuse` does initially feel like a dead code lint of sorts.

Dekkonot:
I'd like it if there was a 'reason' argument supported for this attribute, as it may not be obvious at a glance _why_ a value must be used.

Also, this seems like it'd need a presence in the type system. That's not currently implemented, so this RFC is somewhat hiding a much more complicated proposal behind it. I suggest you (or someone else) write an RFC for attributes to exist in the type system.

CompeyDev:
> I suggest you (or someone else) write an RFC for attributes to exist in the type system.

I was looking to do that as an extension to this RFC once it is merged.

PR #84, opened 2024-12-01
RFC: Do Not Release Unfinished Features

jackdotink:
[Rendered](https://github.com/jackdotink/rfcs/blob/do-not-release-unfinished-features/docs/do-not-release-unfinished-features.md).

Recently Luau has released the vector primitive and vector standard library feature, even though type system support is not ready. This RFC seeks to prevent future situations like this.

### DIFF
--- a/docs/do-not-release-unfinished-features.md
+++ b/docs/do-not-release-unfinished-features.md
@@ -1,0 +1,23 @@
+# Do Not Release Unfinished Features
+
+## Summary
+
+Luau should not ship incomplete or unusable features that are not ready for users, even if part of the feature is ready.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


ffrostfall:
I think this RFC is in bad faith. Vectors not having proper types was an oversight. One which only had an issue [5 days ago](https://github.com/luau-lang/luau/issues/1549), where, given the fact that it is a holiday, is clearly not enough time to fix this.

Luau didn't intentionally release this feature "incomplete". It was an oversight, quite a large oversight, but still an oversight.

It's not even like a new feature was actually added. The vector primitive has always been there. The vector library was a feature designed to let you interact with the vector primitive which was previously inaccessible. Given that the vector primitive has existed for months, and the library was released like, 2 weeks ago? I think it's an understandable mistake that the Luau team forgot that the vector primitive, **which was added a long time ago,** did not have a metatable type on it.

aatxe:
RFCs are not a venue for complaining about being unhappy with our release schedule, and we are not going to explicitly define a notion of feature releases to mean "whenever it's enabled on Roblox."

PR #85, opened 2024-12-05
RFC: Support for Generic Types and Packs in User-Defined Type Functions

vegorov-rbx:
I particular, this enables user-defined type functions to be used on Roblox DataModel types, but overall value for Luau that is full of generic functions is also great.

[Rendered](https://github.com/luau-lang/rfcs/blob/udtf-generics/docs/support-for-generic-function-types-in-user-defined-type-functions.md)

### DIFF
--- a/docs/support-for-generic-function-types-in-user-defined-type-functions.md
+++ b/docs/support-for-generic-function-types-in-user-defined-type-functions.md
@@ -2,14 +2,14 @@
 
 ## Summary
 
-Add support for generic function types in user-defined type function to allow more types to be accepted by them, including tables and classes which might contain generic function types.
+Add support for generic function types in user-defined type functions to allow more types to be accepted by them, including tables and classes which might contain generic function types.
 
 ## Motivation
 
 Generic functions are very common in Luau, especially when inferred by the typechecking engine. They are also often present in classes.
 
 Omitting generic types from appearing in user-defined type functions limits their use.
-Even if there is no direct manipulation of generics, an serialization error appears immediately.
+Even if there is no direct manipulation of generics, a serialization error appears immediately.
 
 ## Design
 
@@ -31,12 +31,12 @@ For example, consider a type `<T, U>(T, U, <T>(T) -> U) -> ()`
 From the typechecking perspective, inner and outer function 'T' types are different, but they are still represented by the same runtime generic named 'T'.
 
 The ambiguity is resolved purely on the scoping of the generic in question.
-When runtime type is returned back to the typechecking engine, only those generics that are in current set of function type generic lists are allowed to be mentioned.
-When inner generic function list places name 'T' again, it represents a different typechecking type and hides outer 'T'.
+When the runtime type is returned back to the typechecking engine, only those generics that are in the current set of function type generic lists are allowed to be mentioned.
+When the inner generic function list places the name 'T' again, it represents a different typechecking type and hides the outer 'T'.
 
 So we can say generics are defined by the lexical scope and that actually matches how they are written out in code.
 
-Just like in the syntax of the language, reusing the same generic name, even between types and packs in a single list in a duplicate generic definition error.
+Just like in the syntax of the language, reusing the same generic name, even between types and packs in a single list results in a duplicate generic definition error.
 
 ### Examples
 
@@ -107,7 +107,7 @@ end
 
 We can assign a different `tag` to a generic pack instead of the `ispack` property.
 
-It is also possible to split generic type and packs list into two.
+It is also possible to split a single list containing both generic types and packs into two.
 This was done in an initial prototype implementation as was inconvenient to specify.
 One benefit was that you didn't have to split them manually when reading them.
 
--- a/docs/support-for-generic-function-types-in-user-defined-type-functions.md
+++ b/docs/support-for-generic-function-types-in-user-defined-type-functions.md
@@ -9,15 +9,15 @@ Add support for generic function types in type functions to allow more types to 
 Generic functions are very common in Luau, especially when inferred by the typechecking engine. They are also often present in classes defined by the environment.
 
 Omitting generic types from appearing in type functions limits their use.
-Today, even if there is no direct manipulation of generics, a serialization error appears immediately.
-Ideally, we want use-defined type functions to support any possible type, aside from temporary implementation-detail types of the type solver.
+Today, even if there is no direct manipulation of generics, a serialization error appears immediately for any generic function or class containing them.
+Ideally, we want user-defined type functions to support any possible type, aside from temporary implementation-detail types of the type solver.
 
 ## Design
 
 We will introduce a single 'generic' kind of runtime type for both generic types and generic packs.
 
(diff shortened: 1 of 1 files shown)


docs/support-for-generic-function-types-in-user-defined-type-functions.md
```diff
@@ -0,0 +1,120 @@
...
+
+Add support for generic function types in type functions to allow more types to be accepted by them, including tables and classes which might contain generic function types.
+
+## Motivation
+
+Generic functions are very common in Luau, especially when inferred by the typechecking engine. They are also often present in classes defined by the environment.
+
+Omitting generic types from appearing in type functions limits their use.
+Today, even if there is no direct manipulation of generics, a serialization error appears immediately.
+Ideally, we want use-defined type functions to support any possible type, aside from temporary implementation-detail types of the type solver.
```
aatxe, line 13:
```suggestion
Ideally, we want user-defined type functions to support any possible type, aside from temporary implementation-detail types of the type solver.
```

docs/support-for-generic-function-types-in-user-defined-type-functions.md
```diff
@@ -0,0 +1,120 @@
...
+## Summary
+
+Add support for generic function types in type functions to allow more types to be accepted by them, including tables and classes which might contain generic function types.
+
+## Motivation
+
+Generic functions are very common in Luau, especially when inferred by the typechecking engine. They are also often present in classes defined by the environment.
+
+Omitting generic types from appearing in type functions limits their use.
+Today, even if there is no direct manipulation of generics, a serialization error appears immediately.
```
aatxe, line 12:
```suggestion
Today, even if there is no direct manipulation of generics, a serialization error appears immediately for any generic function or class containing them.
```

docs/support-for-generic-function-types-in-user-defined-type-functions.md
```diff
@@ -0,0 +1,120 @@
...
+Ideally, we want use-defined type functions to support any possible type, aside from temporary implementation-detail types of the type solver.
+
+## Design
+
+We will introduce a single 'generic' kind of runtime type for both generic types and generic packs.
+
+While this doesn't match up the typechecking engine representation directly, we avoid having to deal with two different userdata types.
+Being content of a generic type or pack is not known upfront, assigning separate tags is not that different from having an extra property on it.
```
aatxe, line 20:
```suggestion
Since the value of a generic type or pack is not known upfront, assigning separate tags is not that different from having an extra property on it.
```

docs/support-for-generic-function-types-in-user-defined-type-functions.md
```diff
@@ -0,0 +1,120 @@
...
+
+We already have variadic types (`...number`) that get represented as a regular runtime type and only gain that special meaning when placed in a 'tail' spot.
+We also have a different representation of a table with a metatable between type functions and typechecking.
```
aatxe, line 23: Maybe worth adding something like "These variations were consciously chosen to improve the developer experience for user-defined type functions."

docs/support-for-generic-function-types-in-user-defined-type-functions.md
```diff
@@ -0,0 +1,120 @@
...
+
+Names are the way generic identity is specified.
+Generic types are immutable and two generics with the same name are equal, even if they are defined in different function type scopes.
+
+For example, consider a type `<T, U>(T, U, <T>(T) -> U) -> ()`
+
+From the typechecking perspective, inner and outer function 'T' types are different, but they are still represented by the same runtime generic named 'T'.
+
+The ambiguity is resolved based on the scoping of the generic in question.
+When the runtime type is returned back to the typechecking engine, only those generics that are in the current set of function type generic lists are allowed to be mentioned.
```
aatxe, line 36:
```suggestion
When the runtime type is returned back to the typechecking engine, only those generics that are in the current set of function type generic lists are allowed to be mentioned. Usage of ill-scoped generics will result in serialization errors.
```

I don't know if this is how you implemented it, but I hope so.

docs/support-for-generic-function-types-in-user-defined-type-functions.md
```diff
@@ -0,0 +1,120 @@
...
+When the inner generic function list places the name 'T' again, it represents a different typechecking type and hides the outer 'T' until the scope of the inner type ends.
+
+So we can say generics are defined by the lexical scope, which matches how they are written out in source code.
```
aatxe, line 39:
```suggestion
As such, generics are defined by their lexical scope, matching how they are written out in source code.
```

docs/support-for-generic-function-types-in-user-defined-type-functions.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+## Updates to the library and type userdata
+
+### `types` Library
+
+| New/Update | Library Functions | Return Type | Description |
+| ------------- | ------------- | ------------- | ------------- |
+| New |  `generic(name: string, ispack: boolean?)` | `type` | returns an immutable instance of a generic type or a pack; name cannot be empty |
+| Update | `newfunction(parameters: { head: {type}?, tail: type? }, returns: { head: {type}?, tail: type? }, generics: {type}?)` | `type` | New `generic` argument is added to specify generic types and packs in a single list |
```
aatxe, line 86: This is sort of unfortunate parameter ordering, but at the same time, I think it's probably the one we'd pick based on usefulness even if we had generics all along.

docs/support-for-generic-function-types-in-user-defined-type-functions.md
```diff
@@ -0,0 +1,120 @@
...
+
+It is also possible to split a single list containing both generic types and packs into two.
+This was done in an initial prototype implementation and was inconvenient to specify.
+One benefit was that list didn't have to be split manually when reading types or packs individually.
+
+## Drawbacks
+
+Generics do bring additional complexity to type functions, especially with the way we can now hold a generic type pack as a separate item (while `type T = U...` impossible to define in the syntax of the language).
+
+The way they bring in lexical scoping consideration is also new for type function runtime.
```
aatxe, line 120:
```suggestion
The way they bring in lexical scoping considerations is also new for the type function runtime.
```

aatxe (approved):
Overall looks good, just some editing nits.

docs/support-for-generic-function-types-in-user-defined-type-functions.md
vegorov-rbx: Doesn't exactly follow the order we have when writing source code, but making a 3-arg overload with a different order seemed like a path to confusion.

docs/support-for-generic-function-types-in-user-defined-type-functions.md
vegorov-rbx: Added a few lines!

aatxe: Looks good!

PR #86, opened 2024-12-11
RFC: Add function math.lerp

zeux:
[Rendered link](https://github.com/luau-lang/rfcs/blob/math-lerp/docs/function-math-lerp.md)

Closes #54 (by replacing it)


petrihakkinen:
Good stuff! I support the addition of `math.lerp` but I haven't had time to read through the details yet. However, we'd be even more interested in `vector.lerp` because the performance gap between a built-in and Lua implementation is likely much larger. This is a bit tangentual to the RFC, but I think we should consider whether `math.lerp` and `vector.lerp` can share the same properties (exactness, monotonicity, etc.). For example, is it possible to guarantee them and still have a performant implementation using simd math. And what would it mean in practice if `vector.lerp` and `math.lerp` have different properties (other than internal precision).

zeux:
The considerations wrt numerical properties are the same and can be achieved with the same implementation, with comparable performance (the extra cost is similarly small and the delta is likely outweighed by fastcall cost in VM implementations). With vector lerp there's a question as to whether the lerp factor is a scalar or a vector itself; this doesn't change the numerics but does change the interface. However, this is not related to `math.lerp` so I'll leave other people to explore this in other RFCs :)

vegorov-rbx (approved):

PR #87, opened 2024-12-13
RFC: Generic Constraints

Kiiyoko:
Hi! This is an RFC to implement extending generics/constraining generics, like in other languages. This is meant to build upon #50's weakpoints.

[Rendered](https://github.com/Kiiyoko/luau-rfcs/blob/master/docs/generic-constraints.md)

### DIFF
--- a/docs/generic-constraints.md
+++ b/docs/generic-constraints.md
@@ -58,4 +58,4 @@ local function callbackProperty<T, K: keyof(T)>( object: T, key: K, callback: (i
 - Adding an extra use to `&` could make its usage more confusing to novices.
 ## Alternatives
 - Don't do this; this would make it impossible for functions like above to be able to be inferred correctly. Just let people explicitly type their variables instead of inferring types. This makes code more verbose and would likely not allow for full optimization.
-- Use overloaded functions as previously mentioned, but this wouldn't allow the usage of a generic with correct type inference and would require users to add a new overload for each key.
+- Use overloaded functions as previously mentioned, but this wouldn't allow the usage of a generic, would require users to add a new overload for each key.
--- a/docs/generic-constraints.md
+++ b/docs/generic-constraints.md
@@ -39,7 +39,7 @@ end
 
 type getProperty<T, K: keyof<T>> = ( object: T, key: K ) -> index<T, K>
 ```
-This would match the current expectation of `:` being used to type annotate a variable. This would be backwards compatible without performance implications. As a note, the ordering of `T` and `K` is arbitrary and can be switched if desired.
+This would match the current expectation of `:` being used to type annotate a variable. This would be backwards compatible without major performance implications. As a note, the ordering of `T` and `K` is arbitrary and can be switched if desired.
 
 2. Use a `where` clause.
 ```luau
@@ -97,8 +97,6 @@ In this example, `T` must either be a number or vector, and setting annotating v
 > This drawback is difficult justify in Luau's current state. The introduction of type negation could solve this, but is outside the scope of this RFC. If type negation were to be added, the proposed syntax would allow for both of these types of constraints.
 - This would complicate Luau's grammar and add an extra keyword.
 > Any addition would add further complexity to the language. Adding the `where` keyword would also introduce another conditional keyword to the language.
-- The introduction of bounded polymorphism would greatly complicate the compiler.
-> I am not personally familiar with the internals of the typechecker.
 
 ## Alternatives
 1. Don't do this; this would make it impossible for functions like above to be able to be automatically inferred correctly. Just let people explicitly annotate their variables instead of inferring types. This makes code more verbose.
--- a/docs/generic-constraints.md
+++ b/docs/generic-constraints.md
@@ -2,8 +2,8 @@
 ## Summary
 Introduce syntax to extend/constrain the type of a generic.
 ## Motivation
-Luau currently does not provide a way in order to constrain the type of a generic without direct assertions.
-```lua
+Luau currently does not provide a way in order to constrain the type of a generic without direct assertions, and even so, would not properly infer the right type.
+```luau
 local qux = { foo = 10, bar = "string" }
 
 local function getProperty<T>( object: T, key: keyof<T> )
@@ -17,13 +17,13 @@ local bar = getProperty( qux, "bar" )
 ```
 This is wrong! We would expect foo to be a number, with bar being a string.
 In the following snippet, it is impossible to tell whether or not the key variable is the same as the one being used to index T in the callback.
-```lua
+```luau
 local function callbackProperty<T>( object: T, key: keyof<T>, callback: (index<T, keyof<T>>) -> () )
-...
+....
 ```
 ## Design
 The design of this would take inspiration from TypeScript's `extends`, where instead of the additional keyword, we would use `&`.
-```lua
+```luau
 local qux = { foo = 10, bar = "string" }
 
 local function getProperty<T, K & keyof<T>>( object: T, key: K )
@@ -37,7 +37,7 @@ local bar = getProperty( qux, "bar" )
 ```
 This would correctly infer the type of the each key's value.
 The following snippet would also thus be correctly inferred.
-```lua
+```luau
 local qux = { foo = 10, bar = "string" }
 
 local function callbackProperty<T, K & keyof<T>>( object: T, key: K, callback: (index<T, K>) -> () )
@@ -49,9 +49,9 @@ callbackProperty( qux, "foo", function( foo )
 end)
 ```
 An alternative syntax could be the following, but does not satisfy current requirements about `:` only being used inside `()` and `{}`, and personally does not look as good.
-```lua
-local function getProperty<T, K: keyof<T>>( object: T, key: K ) ... end
-local function callbackProperty<T, K: keyof<T>>( object: T, key: K, callback: (index<T, K>) -> () ) ... end
+```luau
+local function getProperty<T, K: keyof<T>>( object: T, key: K ) .... end
+local function callbackProperty<T, K: keyof<T>>( object: T, key: K, callback: (index<T, K>) -> () ) .... end
 ```
 ## Drawbacks
 - I am not personally familiar with the internals of the typechecker, but this has a chance to further complicate type inference.
--- a/docs/generic-constraints.md
+++ b/docs/generic-constraints.md
@@ -7,7 +7,7 @@ Luau currently does not provide a way in order to constrain the type of a generi
 local qux = { foo = 10, bar = "string" }
 
 local function getProperty<T>( object: T, key: keyof<T> ): index<T, keyof<T>>
-  return object[key]
+	return object[key]
 end
 
 local foo = getProperty( qux, "foo" )
@@ -34,7 +34,7 @@ There are a few options for the syntax of generic constraints.
 1. Treat the generic parameter as a variable--annotate it as declaring a new variable.
 ```luau
 local function getProperty<T, K: keyof<T>>( object: T, key: K ): index<T, K>
-  return object[ key ]
+	return object[ key ]
 end
 
 type getProperty<T, K: keyof<T>> = ( object: T, key: K ) -> index<T, K>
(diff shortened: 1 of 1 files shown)


aatxe (changes requested):
Thanks for the RFC! Bounded polymorphism is definitely a feature we are looking to add to Luau going forward, and we're excited to be able to do so as soon as possible. I've left a lot of feedback throughout the RFC, but overall, I think the biggest issues here are problems with the actual syntax being proposed, and the actual RFC itself not doing a good job of exploring options and capturing the design space here.

Here are some things that I think this RFC should be shooting for:
1. We should aim to capture at least a few different viable syntax options for bounded generics, including giving consideration to the idea of separate where clauses. It's often the case in languages with bounded polymorphism that the set of constraints on a type can get very large, and having them all inline with its definition can itself become cumbersome. If this RFC wants to argue that such a thing is unnecessary or that it's worth adding this feature without where clauses, that's fine, but the discussion should be present within the RFC text.
2. We should aim to clearly describe the set of bounds that we are trying to add to generic type parameters. Within the scope of this RFC right now, that is solely subtyping constraints. We should describe that explicitly, and include any additional options listed as either part of the RFC or as alternatives that we are consciously choosing not to incorporate (and with an explanation of why not). Some of these alternatives could include inhabitance constraints, equality constraints, inequality constraints, not-subtyping constraints, etc. They all have different use cases, and we should make an argument for why we are or are not including them.
3. We should aim to capture the design of the feature with simple, clear examples paired with crisp English explanations of the intended behavior. You can work up to your example that relies on two type functions, for instance, by talking about the difference between a function that e.g. takes and returns a union like `number | string` versus a function that takes and returns a generic type _bounded_ by `number | string`. This distinction is what actually underlies the problem you're hoping to solve with your use of `keyof`.

docs/generic-constraints.md
```diff
@@ -0,0 +1,64 @@
...
+Type interference believes that either value could be a number or string as keyof<T> is too broad.
+
+```luau
+local function callbackProperty<T>( object: T, key: keyof<T>, callback: (index<T, keyof<T>>) -> () )
+....
+```
+It is impossible to tell whether or not the key variable is the same as the one being used to index T in the callback, and thus Luau infers that it could be a number | string.
+
+## Design
+The design of this would take inspiration from TypeScript's `extends`, where instead of the additional keyword, we would use `&`.
```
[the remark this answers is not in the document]
Kiiyoko: Agreed. Will use a `:` to indicate its type rather than an `&`--not sure why I insisted so heavily on the ampersand.

docs/generic-constraints.md
```diff
@@ -0,0 +1,64 @@
...
+This would also be correctly inferred as index<T, K> would correctly result in the type of the field.
+
+An alternative syntax could use `:` instead of `&`. However, this would not match the current semantics of `:` and using `&` implies a union of types.
+```luau
+local function getProperty<T, K: keyof<T>>( object: T, key: K ) .... end
+local function callbackProperty<T, K: keyof<T>>( object: T, key: K, callback: (index<T, K>) -> () ) .... end
+```
+## Drawbacks
+- I am not personally familiar with the internals of the typechecker, but this has a chance to further complicate type inference.
+- Adding an extra use to `&` could make its usage more confusing to novices.
```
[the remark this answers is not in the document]
Kiiyoko: Agreed; the proposed syntax will be changed in order to moreso match current type meanings.

Kiiyoko:
Hi, thank you so much for the feedback. It has helped a lot in trying to rewrite some sections of the RFC.

> 1. We should aim to capture at least a few different viable syntax options for bounded generics, including giving consideration to the idea of separate where clauses. It's often the case in languages with bounded polymorphism that the set of constraints on a type can get very large, and having them all inline with its definition can itself become cumbersome. If this RFC wants to argue that such a thing is unnecessary or that it's worth adding this feature without where clauses, that's fine, but the discussion should be present within the RFC text.
> 2. We should aim to clearly describe the set of bounds that we are trying to add to generic type parameters. Within the scope of this RFC right now, that is solely subtyping constraints. We should describe that explicitly, and include any additional options listed as either part of the RFC or as alternatives that we are consciously choosing not to incorporate (and with an explanation of why not). Some of these alternatives could include inhabitance constraints, equality constraints, inequality constraints, not-subtyping constraints, etc. They all have different use cases, and we should make an argument for why we are or are not including them.

This was incredibly important feedback; I have expanded upon `where` clauses signficantly, and they should allow even more constraints to exist with the way that these constraints are set up.

> 3. We should aim to capture the design of the feature with simple, clear examples paired with crisp English explanations of the intended behavior. You can work up to your example that relies on two type functions, for instance, by talking about the difference between a function that e.g. takes and returns a union like `number | string` versus a function that takes and returns a generic type _bounded_ by `number | string`. This distinction is what actually underlies the problem you're hoping to solve with your use of `keyof`.

I have added a few examples of this. Thank you so much for the direction.

And overall, I am very sorry for my vocabulary when writing the RFC. I have read through the documentation for Luau quite a bit and I'm surprised that I mixed up `&` and `|`. If you have any other commentary I would love to hear it.

docs/generic-constraints.md
```diff
@@ -0,0 +1,142 @@
...
+local function getProperty<T>( object: T, key: keyof<T> ): index<T, keyof<T>>
+	return object[key]
+end
+
+local foo = getProperty( qux, "foo" )
+-- foo: number | string
+local bar = getProperty( qux, "bar" )
+-- bar: number | string
+```
+Type interference believes that either value could be a number or string as `keyof<T>` is a union type of either foo or bar.
```
alexmccord, line 18:
```suggestion
Type inference believes that either value could be a number or string as `keyof<T>` is a union type of either foo or bar.
```

docs/generic-constraints.md
```diff
@@ -0,0 +1,142 @@
...
+local function callbackProperty<T>( object: T, key: keyof<T>, callback: (property: index<T, K>) -> () )
+....
+```
+This does not work in Luau.
+> Ideally, this behaviour should work out of the box, but, could be solved with updates to the type engine.
+
+## Design
+### Syntax
+There are two new options for syntax that come with this generic constraints.
+1. Treat the generic parameter as a variable--annotate it as declaring a new variable.
```
alexmccord, line 34: > "Treat the generic parameter as a variable--annotate it as declaring a new variable."

It actually _is_ a variable, but it ranges over types instead of values, so `X: T` is a natural consequence from that.

docs/generic-constraints.md
```diff
@@ -0,0 +1,142 @@
...
+```
+This would allow users to specify the constraints of the separately from the generic declaration itself. This would be reminiscent to users of C#. Imposing multiple constraints on different generics can be done by delimiting with `,`. This would be backwards compatible, without major performance implications, as it could just be a conditional keyword. 
+
+The `where` clause syntax can be further expanded to pose custom bounds on generics.
+```luau
+local function add<T, K where add<T, K>>( a: T, b: K )
+	return a + b
+end
+```
+The `add` function would only accept `T` and `K` if `add<T, K` is not `never`. This is currently available in Luau, but this behaviour is locked behind `add<T, K>` being the return type.
```
alexmccord, line 60:
```suggestion
The `add` function would only accept `T` and `K` if `add<T, K>` is not `never`. This is currently available in Luau, but this behaviour is locked behind `add<T, K>` being the return type.
```

docs/generic-constraints.md
```diff
@@ -0,0 +1,142 @@
...
+-- Attempt 3
+local sumSpecific: ( ( number, number ) -> ( number ) ) & ( ( vector, vector ) -> ( vector ) ) = function( a, b )
+	return a + b -- okay!
+end
+```
+This is fine, but is not concise whatsoever, and does not allow for the traditional function declaration syntax.
+```luau
+local function sumSpecific<T: number | vector>( a: T, b: T ): T
+	return a + b
+end
```
alexmccord, line 94: It's a bit hard to do this. A counterexample:

```luau
function f() -- f : () -> number | string
  return if math.random() > 0.5 then 5 else vector.create(0, 0, 0)
end

sumSpecific(f(), f())
```

Which would still infer `T` to be `number | vector`, and provide a `number` for `a` and `vector` for `b`.

docs/generic-constraints.md
```diff
@@ -0,0 +1,142 @@
...
+sumSpecific( 42, 143 ) -- okay!
+
+sumSpecific( vector.create( 42, 42, 42 ), 143 ) -- error! a and b do not match.
+```
+In this example, `T` must either be a number or vector, and setting annotating variables `a` and `b` as `T` would solve this. This can also be done by overloading the function but is more verbose and less elegant than this solution. This cannot be done with Luau's `add` type function, as it would allow all 'addable' types, and could not guarantee the return type being the same as both inputs.
+
+#### Other Constraints
+```luau
+local function multiply<T, K where T: number, mul<T, K>>( a: T, b: K )
+	return a + b
```
alexmccord, line 106:
```suggestion
	return a * b
```

docs/generic-constraints.md
```diff
@@ -0,0 +1,142 @@
...
+sumSpecific( vector.create( 42, 42, 42 ) + vector.create( 143, 143, 143 ) ) -- okay!
+sumSpecific( 42, 143 ) -- okay!
+
+sumSpecific( vector.create( 42, 42, 42 ), 143 ) -- error! a and b do not match.
+```
+In this example, `T` must either be a number or vector, and setting annotating variables `a` and `b` as `T` would solve this. This can also be done by overloading the function but is more verbose and less elegant than this solution. This cannot be done with Luau's `add` type function, as it would allow all 'addable' types, and could not guarantee the return type being the same as both inputs.
+
+#### Other Constraints
+```luau
+local function multiply<T, K where T: number, mul<T, K>>( a: T, b: K )
```
alexmccord, line 105: So if we look closer, we can see that (assuming `return a + b` was meant to be `return a * b`) the function `multiply`'s return type would be inferred as `mul<T, K>`, which is covariant and all its dependent type variables are contravariant: this renders `mul<T, K>` in the `where` clause to be _redundant_. That is, the example is equivalent to

```suggestion
local function multiply<T, K where T: number>( a: T, b: K ): mul<T, K>
```

docs/generic-constraints.md
```diff
@@ -0,0 +1,142 @@
...
+end
+
+multiply(10, vector.create(10, 10, 10)) -- okay!
+multiply(10, 41) -- okay!
+multiply(10, "hello!") -- error! add<10, "hello"> is not okay.
+```
+The `multiply` function would be bounded by `T: number`, along with `mul<T, K>`. This can be used to create any type of constraint if user-defined type functions work with generics in the future.
+
+## Drawbacks
+- There are no built-in inequality constraints nor not-subtyping constraints.
```
alexmccord, line 116: Negation types! :)

alexmccord (commented):
Two comments:

Firstly, generic bounds has direct consequences with lack of reexports. Suppose you have `export type Foo<T> = T` in a module, and you decide to update it to `export type Foo<T: string> = T`. This now requires you to update _all_ reexports variation of `Foo`, and _all_ type functions that depends on `Foo`.

This is drastic, since that means you have `N * (M1+...+Mi)` amount of aliases and functions to update where `N` represents the number of modules along the reexport chain, and `M1+...+Mi` represents the sum of type aliases and functions that depends on `Foo`. This [program](https://github.com/JohnnyMorganz/wally-package-types) exists because of the lack of good reexport story.

Obviously, this problem is orthogonal to this RFC, but I think it deserved to be raised to perhaps motivate the story of reexports.

Secondly, I noticed the lack of scoping rules in generic bounds. Type inference is by design _unordered_ to reach fixpoint, so scoping rules generally does not care about the ordering of aliases, I think this should also extend to type parameters too, that is: `<r: add<a, b>, a, b>(a, b) -> r` is just as valid as `<a, r: add<a, b>, b>(a, b) -> r` as it is to `<a, b, r: add<a, b>>(a, b) -> r`.

<sup><sup>btw you can write `git switch -c branch-name`</sup></sup>

PR #88, opened 2024-12-14
Allow table.clone to copy tables with locked metatables

ccuser44:
Allow `table.clone` to copy tables with locked metatables.

[Rendered](https://github.com/ccuser44/rfcs/blob/patch-2/docs/table.clone-locked-metatables.md)

This is a remake of the original poorly done RFC #73 . The issues in the original RFC have been fixed and the ambiguities have been clarified.

### DIFF
--- a/docs/table.clone-locked-metatables.md
+++ b/docs/table.clone-locked-metatables.md
@@ -65,7 +65,6 @@ However, this is not an issue as the changes proposed in this RFC do not allow f
 
 Another potential drawback that could be levied at this proposal is a reduced debuggability of not hard-failing when the expected behavior would be to create a shallow copy with the metatable of the original table.
 However, this point is most certainly moot in practice, as most users very likely are not even aware, or desire the fact that `table.clone` can assign the metatables of the original table to the clone.
-<!-- This was the point about the dictionary definition of shallow copying made in the original RFC -->
 
 ## Alternatives
 
--- a/docs/table.clone-locked-metatables.md
+++ b/docs/table.clone-locked-metatables.md
@@ -7,7 +7,7 @@ Allow `table.clone` to copy tables with locked metatables.
 ## Motivation
 
 As proposed by the [`table.clone` RFC](function-table-clone.md), the function `table.clone` cannot create shallow copies of a table if it has a locked metatable.
-Due to this ugly limitation, it is extremely un-ergonomic to create shallow copies of arbitary tables.
+Due to this ugly limitation, it is extremely un-ergonomic to create shallow copies of arbitrary tables.
 The current behavior of using `table.clone` on a table with a locked metatable is a hard-fail approach where `table.clone` spits out an error instead of having a soft-fail approach wherein a clone is generated without a metatable.
 This hard-fail approach severely hinders the usefulness of `table.clone` due to in practice users having to re-implement the function in many scenarios.
 
@@ -19,18 +19,18 @@ ie. The behavior (as described by Lua pseudocode) would go from:
 
 ```lua
 function table.clone(t)
-assert(type(t) == "table")
-local nt = {}
+	assert(type(t) == "table")
+	local nt = {}
 
-for k, v in pairs(t) do
-nt[k] =v
-end
+	for k, v in pairs(t) do
+		nt[k] =v
+	end
 
-if type(getmetatable(t)) == "table" then
-setmetatable(nt, getmetatable(t))
-end
+	if type(getmetatable(t)) == "table" then
+		setmetatable(nt, getmetatable(t))
+	end
 
-return nt
+	return nt
 end
 ```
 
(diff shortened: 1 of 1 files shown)


dphfox:
Would it be intuitive for a metatable to only sometimes be copied?

ccuser44:
> Would it be intuitive for a metatable to only sometimes be copied?

Yes

docs/table.clone-locked-metatables.md
```diff
@@ -0,0 +1,73 @@
+# Feature name
+
+## Summary
+
+Allow `table.clone` to copy tables with locked metatables.
+
+## Motivation
+
+As proposed by the [`table.clone` RFC](function-table-clone.md), the function `table.clone` cannot create shallow copies of a table if it has a locked metatable.
+Due to this ugly limitation, it is extremely un-ergonomic to create shallow copies of arbitrary tables.
```
vegorov-rbx, line 10: Please refrain from using such language in the text of the RFC.

docs/table.clone-locked-metatables.md
```diff
@@ -0,0 +1,73 @@
...
+		setmetatable(nt, getmetatable(t))
+	end
+
+	return nt
+end
+```
+
+## Drawbacks
+
+There are no drawbacks to this.
```
vegorov-rbx, line 60: This can be removed as you do have a drawback listed directly below.

docs/table.clone-locked-metatables.md
```diff
@@ -0,0 +1,73 @@
...
+
+The original RFC puts forth a rationale for restricting cloning tables with locked metatables due to security reasons.
+This is a valid concern. The issue at hand being an avenue to assign metatables to unintended tables.
+However, this is not an issue as the changes proposed in this RFC do not allow for such behavior.
+
+Another potential drawback that could be levied at this proposal is a reduced debuggability of not hard-failing when the expected behavior would be to create a shallow copy with the metatable of the original table.
+However, this point is most certainly moot in practice, as most users are very likely not even aware, or desire the fact that `table.clone` can assign the metatables of the original table to the clone.
```
vegorov-rbx, line 67: It's not a moot point, the description of what `table.clone` does is immediately clear from the first line of the documentation.

This RFC doesn't even change the behavior of non-protected metatable copies, so unaware users are still better to learn what the functions they use do.

I don't think it's fair to dismiss the issue of the table clone not working as the original in an extended amount of cases.

docs/table.clone-locked-metatables.md
```diff
@@ -0,0 +1,73 @@
...
+## Summary
+
+Allow `table.clone` to copy tables with locked metatables.
+
+## Motivation
+
+As proposed by the [`table.clone` RFC](function-table-clone.md), the function `table.clone` cannot create shallow copies of a table if it has a locked metatable.
+Due to this ugly limitation, it is extremely un-ergonomic to create shallow copies of arbitrary tables.
+The current behavior of using `table.clone` on a table with a locked metatable is a hard-fail approach where `table.clone` spits out an error instead of having a soft-fail approach wherein a clone is generated without a metatable.
+This hard-fail approach severely hinders the usefulness of `table.clone` due to in practice users having to re-implement the function in many scenarios.
```
vegorov-rbx, line 12: Since there are many scenarios, can one or two examples be included in the motivation?

Examples can also note how dropping the metatable is handled in them (given that it will be dropped only if it's protected).

docs/table.clone-locked-metatables.md
ccuser44: Done

docs/table.clone-locked-metatables.md
ccuser44: Done

ccuser44:
I haven't finished this yet because I haven't resolved 2 of the issues levied yet.

docs/table.clone-locked-metatables.md
ccuser44: Done.

ccuser44:
Done

zeux:
I don't fully understand the motivation here. Why would JSONEncode need to clone the table, as opposed to using rawget et al? Why would you want to JSON serialize something with a locked metatable?

The downside is that this RFC proposes changing `table.clone` behavior from "it returns a table with the same structure or fails" to "it returns a table with the same structure, or a table that has different behavior if the original metatable was locked". It's... not obvious that this is a good idea.

> I also don't fully recall the scope of potential for the exploits against "clone that preserves locked metatable". Understanding this would probably require analyzing Roblox internal codebase, which only Luau team members could do.

One alternative I could see that is perhaps more sensible is `table.clone(t, newmt)`, as it is unclear that table.clone with new behavior without an immediate call to setmetatable is useful. This comes with its own challenges, namely that you'd want to allow `newmt = nil`, so the implementation will have to distinguish `newmt = none` vs `newmt = nil`, which is possible but something that is generally frowned upon.

But beyond everything else I think this RFC needs a use case that is actually clear and not adequately solved today.

karl-police:
Does default Lua allow table.copy to copy with metatables, shouldn't it be an optional thing that can be copied?

vegorov-rbx:
> Does default Lua allow table.copy to copy with metatables, shouldn't it be an optional thing that can be copied?

Lua doesn't have `table.copy`/`table.clone` functions.

karl-police:
Should it still be a second argument to copy metatables?

cheesycod:
From a cursory glimpse at this RFC, would this not be a sandboxing hazard if table.clone'd tables shadow cloned the table and then simply dropped the metatable?

vegorov-rbx:
Closing this for the general lack of support/interest from people and the main issue that this change makes `table.clone` silently return a table which is not a clone instead of a more clear error we have today.

PR #89, opened 2025-01-02
RFC: Add a global `eprint` function for printing to `stderr`

Dekkonot:
[Rendered link](https://github.com/Dekkonot/luau-rfcs/blob/function-eprint/docs/function-eprint.md)

Let me know if I've missed anything obvious or if you have questions.

### DIFF
--- a/docs/function-eprint.md
+++ b/docs/function-eprint.md
@@ -23,7 +23,7 @@ A global `eprint` would be added to the global environment. It would have the sa
 - Python has `print` designed around this passing stream handles: `print(...)` prints to `stdout` and `print(..., file=sys.stderr)` prints to `stderr`
 - C has `fprintf` that works similiar to Python's `print`: `printf(stdout, ...)` and `printf(stderr, ...)`
 
-These are just examples. Based on further research on this that it seems roughly split between functions that accept a stream handle to write in (like C or Python) and those that use globals (like JavaScript or Rust). Given Luau currently has no concept of stream handles and `print` already exists, it seems more reasonable to propose the latter.
+These are just examples. Based on further research on this, it seems roughly split between functions that accept a stream handle to write in (like C or Python) and those that use globals (like JavaScript or Rust). Given Luau currently has no concept of stream handles and `print` already exists, it seems more reasonable to propose the latter.
 
 By default, `stderr` is generally not buffered in the same way as `stdout`. This RFC acknowledges this, but does not specify whether `stderr` should be changed to buffer or not: this is likely better left to the embedder rather than decided by Luau itself.
 
--- a/docs/function-eprint.md
+++ b/docs/function-eprint.md
@@ -20,7 +20,7 @@ A global `eprint` would be added to the global environment. It would have the sa
 
 - JavaScript has `console.log` as an equivalent to `print` and `console.error` as an equivalent to `eprint`.
 - Rust has `println!` and `eprintln!` macros that perform equivalently to `print` and the proposed `eprint`
-- Python has `print` designed around this: `print(...)` prints to `stdout` and `print(..., file=sys.stderr)` prints to `stderr`
+- Python has `print` designed around this passing stream handles: `print(...)` prints to `stdout` and `print(..., file=sys.stderr)` prints to `stderr`
 - C has `fprintf` that works similiar to Python's `print`: `printf(stdout, ...)` and `printf(stderr, ...)`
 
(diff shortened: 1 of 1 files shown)


bradsharp:
I've wanted something similar to this in the past for various reasons however as we start expanding these types of capabilities I would rather see them as part of a library than a set of globals. Something akin to console.log and console.err would be my preference.

Dekkonot:
> I've wanted something similar to this in the past for various reasons however as we start expanding these types of capabilities I would rather see them as part of a library than a set of globals. Something akin to console.log and console.err would be my preference.

Like I said in the RFC, this is something I considered! I think it's a bit more complicated though, because ideally that library would have actual IO stream capabilities, and that brings with it some complications. It's may be still worth it just to avoid the global, but I'm less enthralled by it otherwise because it's just functions that act like `print` but in a library.

It's also just slightly more annoying. Call me lazy but I'd rather call `printf` or `eprint` than `console.printf` or `console.eprint`.

ccuser44 (commented):
In vanilla Lua `warn` prints to stderr exactly like the proposed function does. I think `warn` could be a suitable alternative, at the very least it should be mentioned in the alternatives section in the RFC.

Another mention I would drop here is that I think `eprint` is quite an un-intuitive name for the proposed function. I'd much prefer `printerror` or `printe`. Preferrably `printerror`.

One positive consequence of this RFC is a way of actually allowing printing of errors in Roblox (that is of Roblox decides to override this function, which, if implemented they should override).

gaymeowing:
I feel like this would be better done as an overload to error, but either way this is needed. As currently I'm having to use warn, and then call error because warn doesnt halt thread execution and doesnt look like print.

Also this is especially useful in roblox,  as you cant click on a instance full name to find an instance from the output in the explorer. So you end up having to do what I described earlier, if you want a nicer debugging experience.

```luau
warn("[SCRIPT NAME]: instance is not of type 'PVInstance\n\tinstance:'", instance)
error("[SCRIPT NAME]: an error occurred, please check previous warning from this script for details.")
```

vegorov-rbx:
I don't think this function makes much sense for Luau as an embeddable library.
It would add another function that has to be overwritten immediately to re-sandbox or to make actually work - stdout/stderr is actually often not available (in many environments in a way that is useful), including Roblox where stdout/stderr should not be available.

It is true that the proposal is similar to the `warn` function in Lua 5.4, but Luau is based mainly on Lua 5.1 and we take additions that are worthwhile.
In a sense, `print` is also not that great for embedding, but it is present in Lua 5.1 so is there for running existing things smoothly.

In Luau REPL, `warn` can be added as an 'embedder extension', similar to how custom `warn` already exists at Roblox.
Other runtimes which have more IO functionality like Lune or lrt can add it as it makes sense there.

But I don't think this is that useful for base Luau, so I would vote against it.

Dekkonot:
It's fair to view this as a embedding issue because it is one, but I dont think that's enough of a problem to justify rejecting this and all future IO related RFCs on merit. Particularly as efforts continue to support cross-runtime modules, I think having a standard way to print debug info that doesn't abort execution will be valuable.

Presumably people are already having to set up IO for `print` and `error`, so it's clearly feasible. It might be prudent to introduce a better method for this but I've never embedded Luau outside of toy projects so I'm not sure what the current process looks like. For all I know you don't need to handle IO for these at all.

I'm not particularly attached to the name `eprint`; it could be named `warn`, I just actually forgot that function existed in Lua 5.4. My only concern with that would be overlap with the Roblox and Lune globals of the same name, which doesn't seem like that big of a problem.

zeux:
`error` does not do IO. There is only one function right now from baselib that usually needs be overridden, which is `print`. (technically there's also `require` but it is not provided by baselib atm, unsure what the plans are wrt require-by-string)

vegorov-rbx:
> technically there's also `require` but it is not provided by baselib atm, unsure what the plans are wrt require-by-string

Overall plan for `require` is to have a separate C++ library you can link against with a separate `luaopen_require` for embedders (with a potential argument to provide non-file IO callbacks for VFS).

vegorov-rbx:
Closing this at this time.

While this feature has merit, we don't have a defined path for RFCs that align features of 'general-purpose' Luau runtimes (as opposed to embedded Luau).

Expectations of features like this and other IO (we have considered having `os` library file access in Luau REPL before) are also unclear wrt embedding scenarios.
If we dismiss concerns as 'embedders who don't need it will clear not include the method', it has to be understood that it's also the position that Roblox can take and not implement `eprint` at all (I think it's understandable as `os` file access would not be implemented at Roblox for example).

PR #90, opened 2025-01-13
RFC: Explicit type parameter instantiation

Kampfkarren:
[Rendered](https://github.com/Kampfkarren/rfcs-1/blob/type-annotations-in-function-calls/docs/parameterized-function-expressions.md)

TL;DR -- Adds support for calling `f<<T>>()` where `f` is a function that takes a type parameter `T`, similar to Rust's turbofish (`f::<T>()`) or C++'s `f<T>()`. This is done at the expression level, so `f<<T>>` is valid as a value.

### DIFF
--- a/docs/parameterized-function-expressions.md
+++ b/docs/parameterized-function-expressions.md
@@ -1,0 +1,179 @@
+# Parameterized function expressions
+
+## Summary
+
+We will add support for the ability to bind a function defined with generics to specific types at the time of expression. This behavior is analagous to C++'s `f<T>` and Rust's `f::<T>`, where `f` is defined as a function accepting a type parameter `T`.
+
+## Motivation
+
+There are several cases where Luau either makes it overly difficult to declare a type in a function call or overly verbose. For example, let's consider something akin to React's bindings:
+
+```lua
+local function createBinding<T>(value: T): React.Binding<T>
+    -- code
+end
+```
+
+If we wanted a binding to a number, then Luau can trivially infer this type:
+
+```lua
+-- 1999 is a number, so moneyBinding is React.Binding<T>
+local moneyBinding = React.createBinding(1999)
+```
+
+In other cases, the type is more complex. For example, if we want money to be `number?`, then we cannot simply pass in `nil`, as that will infer as `React.Binding<nil>`. Furthermore, while there are theoretically cases like these where Luau can expand out a `nil` into `number?`, it is still beneficial to want this to be restricted at the type to prevent confusing debugging or potential for the type to infer wider than intended.
+
+There are two ways to do this. One is to declare the type of `moneyBinding` itself to `React.Binding`. This would look like...
+
+```lua
+local moneyBinding: React.Binding<number?> = React.createBinding(nil)
+```
+
+This is not only quite verbose, but requires that the user know the exact type of the return value of `React.createBinding`, *and* requires that it really be as trivial as just passing in the types. For example, `React.Binding` could be an even more complex type that asks for some sort of initial state type, or setter type, and so on. In other cases, it might even be defined as an inline table, or as a non-exported type.
+
+The other alternative is something like:
+
+```lua
+local moneyBinding = React.createBinding(nil :: number?)
+```
+
+However, `::` has the ability to cast things in unsound manners. Both these calls are accepted with `::` when we would not want it in this case:
+
+```lua
+local function f(x: number?)
+    React.createBinding(x :: number) -- Allowed
+end
+
+type Object = { a: number }
+React.createBinding({} :: Object) -- Allowed
+```
+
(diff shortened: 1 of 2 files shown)


docs/parameterized-function-expressions.md
```diff
@@ -0,0 +1,179 @@
...
+The syntax proposed in this RFC would close the door on the ability to add a bitwise left-shift operator to Luau. This is because this would make the following usage ambiguous:
+
+```lua
+-- Assuming we have right-shift, because there's no way we add one without the other
+return f<<a, b>>(c)
+```
+
+This is either a parameterized function call of `f(c)` with `a` and `b` as types, or a return of two values--`f << a`, and `b >> (c)`.
+
+Because there would be no ability to add bitwise shifts to the language, it is unlikely that any bitwise operators would be added, as it would be a stark omission.
```
alexmccord, line 123: Should we see a RFC that basically puts the final nail in the coffin for bitwise ops? Time and time again, people want to see it, but `bit32` is better because it guarantees better performance characteristics anyway.

aatxe:
High-level nitpick but can we call this `RFC: Explicit Type Instantiation` or `RFC: Explicit Type Parameter Instantiation`

docs/parameterized-function-expressions.md
```diff
@@ -0,0 +1,179 @@
...
+local a = f<<T, U>>()
+
+-- Calling through index
+local b = t.f<<T>>()
+local c = t:f<<T>>()
+local d = t["f"]<<T>>()
+
+-- Saving `f<<T>>` as a variable to called later
+local e = f<<T>>
+```
```
aatxe, line 96: Might want to lead with this, or something else similar to be clear that the proposal is allowing the expression `prefixexp '<<' TypeList '>>'` as its own expression, and not just  function calls having an optional new piece of syntax.

docs/parameterized-function-expressions.md
```diff
@@ -0,0 +1,179 @@
...
+For unambiguous alternatives, `f.<T>` and `f:<T>` are among the most discussed. This would look like:
+
+```lua
+local moneyBinding = React.createBinding.<number>()
+
+-- or...
+local moneyBinding = React.createBinding:<number>()
+```
+
+The downside of these is that they blur the lines between runtime and static, in the sense that `React.createBinding.` starts out as a runtime concept, followed by the purely static `<number>`. As for `:`, it carries the baggage of `x:y()` which will perform a runtime mutation of the function call in the form of adding on `self`.
```
aatxe, line 149: I would not describe this as a runtime mutation. It's very confusing. There's no mutation happening, `x:y()` is just syntactic sugar for `x.y(x)`.

docs/parameterized-function-expressions.md
```diff
@@ -0,0 +1,179 @@
...
+
+There is also not necessarily a reason that we have to provide symmetrical operators, so something like `f!T, U()` is reasonably parseable, but is not obviously better.
```
aatxe, line 151: Also asymmetrical operators have the drawback of being even more Different:tm: from other instances of type application syntax in languages. This is something where you're providing a collection of arguments, and culturally, folks have decided that collections are grouped with symmetric brackets of some sort.

andyfriesen:
I'm honestly inclined to go for `::<>` as the syntax so we don't close the door to ever adding bitshifting operators.

The ambiguity issue is real, but I think this is obscure enough that we can ask people to use whitespace to differentiate between `foo::<T>()` and `foo :: <T>() -> ()`

Kampfkarren:
Intentionally adding an ambiguous syntax for the sake of potentially adding something that has been denied in the past makes me really uneasy 😕

docs/parameterized-function-expressions.md
Kampfkarren: Yeah I'm not sure the right way to say what I want to say--what I want to say is that `:` is inherently more "interesting" than `.`

aatxe:
> I'm honestly inclined to go for `::<>` as the syntax so we don't close the door to ever adding bitshifting operators.
> 
> 
> 
> The ambiguity issue is real, but I think this is obscure enough that we can ask people to use whitespace to differentiate between `foo::<T>()` and `foo :: <T>() -> ()`

Given the cultural opposition to significant whitespace in our community, I think we have to categorically rule out this possibility, even if it's unlikely to actually affect anyone.

I also think the long-standing arguments against bitshift operators are quite strong. We do not have integers, they are not useful without integers, and the identity of the language would have to change pretty dramatically for us to add separate integers, I'd say.

andyfriesen:
I talked with @aatxe and @vegorov-rbx about this offline.  I'm convinced.  Let's go forward with the syntax `f<<T>>()`

andyfriesen (approved):

PR #91, opened 2025-01-14
RFC: Function default arguments

Bottersnike:
[Rendered](https://github.com/Bottersnike/luau-rfcs/blob/function-default-arguments/docs/function-default-arguments.md)

This RFC proposes adding default argument values, removing the need for `if arg == nil then arg = default end` cases at the start of every function.

TL;DR

```lua
function random_point()
    return { x=math.random(), y=math.random() }
end
function print_manhattan_distance(message_prefix = "", point = random_point(), origin = { x=0, y=0 })
    print(message_prefix .. math.abs(point.x - origin.x) + math.abs(point.y - origin.y))
end
```

along with all of the semantics and type inference you might expect from that example.

I have a working implementation of this feature [on my luau fork](https://github.com/luau-lang/luau/compare/master...Bottersnike:luau:function-default-arguments). The brave adventurer may wish to run a build and play around but otherwise this serves more to show that this could be reasonably add into the language with minimal work required.

### DIFF
--- a/docs/function-default-arguments.md
+++ b/docs/function-default-arguments.md
@@ -1,0 +1,200 @@
+# Function default arguments
+
+## Summary
+
+Add default values to arguments in function definitions, used when a parameter is unspecified or `nil`.
+
+## Motivation
+
+Frequently when writing code, having a default value for an unspecified argument is desired. This can be found in a range of languages, such as C++ and Python. Luau has no first-party support for this, instead filling omitted arguments as `nil` allowing programmers to begin their functions with a series of statements such as:
+
+```lua
+arg = arg or default
+```
+or
+```lua
+if arg == nil then arg = default end
+```
+or
+```lua
+arg = if arg == nil then default else arg
+```
+
+While effective this produces a degree of noise at the start of functions. Notably also the `or` short-circuit approach will coalesce all falsey values to the default value, rather than just `nil` values.
+
+The luau type system also does not currently narrow these types correctly. For example,
+```lua
+function demo_type(arg: number?)
+    if arg == nil then arg = 0 end
+    local x = arg
+    return x
+end
+```
+will result in `x` having an inferred type of `number?` and the function having a return type of `number?`. Resolving this involves creating a new variable name for the parameter post-defaulting rather than shadowing the name. This RFC explicitly hides the optional nature of arguments from the function body.
+
+Further, this defaulting behavior is hidden from tooling that may want to inspect function signatures, such as tooltips in IDEs. A handful of examples from the Roblox standard library that could benefit have been included here:
+
+```lua
+function color3.new(x = 0, y = 0, z = 0)
+function Vector3.new(x = 0, y = 0, z = 0)
+function CFrame.fromEulerAngles(rx: number, ry: number, rz: number, order = Enum.RotationOrder.XYZ)
+function DataStoreService:GetDataStore(name: string, scope = "global", options: DataStoreOptions?)
+function DataStoreService:ListDataStoresAsync(prefix = "", pageSize = 0, cursor = "")
+```
+
+A slightly more complicated Roblox example could call API functions in the default value, for example:
+
+```lua
+local Players = game:GetService("Players")
+function give_players_item(item, players = Players:GetPlayers())
+```
+
+## Design
+
+This proposal at its core suggests the following modification to the luau grammar:
+
+```diff
+- parlist = bindinglist [',' '...' [':' GenericTypePack | Type]]
++ parlist = bindinglistwithdefault [',' '...' [':' GenericTypePack | Type]]
+
+(* snip *)
+
++ bindinglistwithdefault = binding ['=' exp] [',' bindinglistwithdefault]
+```
+
+This allows any function argument to be provided with any expression as its default argument. To explain the proposed semantics, the following demonstration function will be used
+
+```lua
+local ARG2_CONSTANT = 2
+function demo(
+    arg1: number,
+    arg2: number = ARG2_CONSTANT,
+    arg3 = {x=0, y=0},
+    arg4 = print("arg4 default"),
+    arg5: number
+)
+    print(arg1, arg2, arg3, arg4, arg5)
+end
+```
+
+### Semantics: Default expressions are evaluated at call-time
+In python, a common foot-gun is defining a function as `def example(arg=[])` which will use a singleton list instance across all calls of the function. This RFC suggests expressions are evaluated strictly when the function is called, matching the semantics of the `if`-based existing methods.
+
+This is shown both by `arg3` creating a new table every call, and `arg4` outputting `"arg4 default"` to the console.
+
+### Semantics: Default expressions are only evaluated when necessary
+If a value is provided for an argument, the default expression is not evaluated.
+
+Proving a value for `arg4` will silence the `"arg4 default"` output in the console.
+
+### Semantics: Passing `nil` is equivalent to an unspecified argument
+Following similar semantics to assigning a table entry to `nil`, passing `nil` as a function argument causes the default value to be used.
+
+This can be seen by passing `nil` as any of `arg2`, `arg3` or `arg4`.
+
+As a result of this...
+
+### Semantics: Non-default arguments are allowed to follow default arguments
+Unlike in other languages, there is no strict requirement that all arguments after the first default argument must also be default arguments. If we wished to call our demonstration function providing a value for only `arg1` and `arg5` we could call it as
+
+```lua
+demo(1, nil, nil, nil, 5)
+```
+
+This further allows for any arbitrary mixture of provided and non-provided arguments.
+
+### Semantics: Default value expressions cannot access function arguments
+All required default arguments are evaluated before function arguments are bound to the local scope.
+
+For example
+
+```lua
+function foo(a, b = a)
+    print(a, b)
+end
+foo(1)
+```
+
+will output `1 nil` rather than the potentially expected `1 1`. When evaluating the default value for `b`, the parameter `a` has not yet entered scope and so `= a` attempts to reference `a` from higher scopes. This behavior is analogous to the statement `local a, b = 1, a`.
+
+It is noted that this may not be desired behavior in all cases; one can imagine a function such as the below where it may be beneficial to be able to access the arguments during defaulting:
+
+```lua
+function table.remove(t: Array, pos: number = #t)
+```
+
+This RFC does not currently consider this, though future works may re-investigate this behavior.
+
+### Type Semantics: Type annotations take precedence over default types when performing inference
+When inferring types, an explicit annotation has precedence over the inferred type of the default argument. For example, in the following function the error should be that a `Type 'string' could not be converted into 'number'`.
+
+```lua
+function demo_2(a: number = "demo") end
+```
+
+### Type Semantics: Inferred types from default values take precedence over inferred types from the function body
+If a type has been inferred from the default value, a conflicting type within the function body will cause an error. The following code is invalid
+
+```lua
+function demo_3(a = "demo")
+    a = 1  -- Invalid
+end
+```
+
+### Type Semantics: Inferred types from default values are not sealed
+When a table type is inferred from a default value, it is not sealed. If the function body makes modifications, these are unified with the inferred type.
+
+In the following example, `point` has a final type of `{x:number,y:number,z:number}`.
+
+```lua
+function demo_unsealed(point = {x = 0, y = 0})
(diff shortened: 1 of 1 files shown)


docs/function-default-arguments.md
```diff
@@ -0,0 +1,200 @@
...
+end
+function demo_sealed_2(point = {x = 0, y = 0} :: PointType)
+    point.z = 0  -- Invalid
+end
+```
+
+### Language implementation
+Within the AST, it would likely be simpler to add a new `AstArray<AstExpr*> argsDefaults` to `AstExprFunction` alongside `args`, rather than modify `args` to be an `std::pair` due to the existing widespread usage of `args`.
+
+Rather than implement this as a feature of the `CALL` instruction within Luau's VM it is instead suggested by this RFC to implement this as part of the compiler. This both increases compatibility (as the VM remains unchanged) and makes it easier to allow *any* expression to be used.
```
Kampfkarren, line 173: So does this mean `function f(x: number = 3)` means that I call `f()` and it compiles as `f(3)`?

What happens then if I use that function as a first class value?

What happens then in this case?
```lua
local function g(callback: (number) -> ())
    callback()
end

g(f)
```

docs/function-default-arguments.md
```diff
@@ -0,0 +1,200 @@
...
+
+This allows any function argument to be provided with any expression as its default argument. To explain the proposed semantics, the following demonstration function will be used
+
+```lua
+local ARG2_CONSTANT = 2
+function demo(
+    arg1: number,
+    arg2: number = ARG2_CONSTANT,
+    arg3 = {x=0, y=0},
+    arg4 = print("arg4 default"),
```
Kampfkarren, line 73: Does this have access to prior arguments?

docs/function-default-arguments.md
Kampfkarren: I ask because surely something with default parameters cannot have a non-optional argument as part of its type and *also* be a first class value?

Bottersnike: The suggested implementation essentially injects a series of `if x == nil then x = default end` statements at the start of the function body, so passing `f` around as a first class value is totally fine. If we run
```lua
-- !strict
function f(x: number = 3)
    print(x)
end

local function g(callback: (number) -> ())
    callback()
end

g(f)
```
we get `3` on stdout as expected.

We do however get a type error of `TypeError: Argument count mismatch. Function 'callback' expects 1 argument, but none are specified` because the first class value `f` will have a type of `(number?) -> ...any` as defined there (to indicate that we need not provide that argument/could provide `nil`).

Amending g to `local function g(callback: (number?) -> ())` that type error goes away.

Bottersnike: I've [added a section](https://github.com/Bottersnike/luau-rfcs/blob/function-default-arguments/docs/function-default-arguments.md#type-semantics-arguments-with-a-default-value-are-implicitly-optional) that hopefully formalises that brief explanation into the RFC. lmk if there's still confusion (or another type interaction I've not formalised!).

stravant:
What is the behavior of the upvalue references in the default value expressions? Do they act like any other upvalue reference from within the body?

For example:

```luau
local x = "foo"
local function getX(arg = x)
    return arg
end
print(getX()) --> foo
x = "bar"
print(getX()) --> foo or bar?
```

Bottersnike:
It would be `bar` as the arguments are evaluated every time the function is called. Making this not be the case would require either dropping the behaviour where evaluations are performed at call time, or special-casing single upvalues, both of which I don't think would be ideal?

bjcscat:
If the intent is for this to replace redefining the argument with a default value in the body then it would make sense for it to be captured as an upvalue

jackdotink:
I don't like default values for function arguments; I think they cause more problems than they solve.

bjcscat:
I think any questions of the logic are resolved by viewing the default argument as a reassignment.

```
local function identity(a=1)
    return a
end
```

would be equivalent to

```
local function identity(a)
    local a = if a == nil then 1 else a
    return a
```

Bottersnike:
As described in the document at the moment, the semantics mean `call(nil)` would result in the default argument being used instead of the explicitly passed `nil`. This has the benefit of allowing `call(arg1, nil, arg3)`. There's been quite a bit of discussion about this. On one hand, this matches the existing patterns you'll come across when reading code (where people have implemented default values using `if`/`or`/etc. at the start of the function body).

On the other hand, this means passing `nil` explicitly isn't possible if there's a default argument present. For example we could consider `function foo(bar: number? = 1)` which no longer has any code path where `bar` could equal `nil`. The solution for this would be to disallow `call(arg1, nil, arg3)`, and instead **exclusively** apply default arguments to trailing omitted arguments.

I'm genuinely unsure which would be better (and maybe JavaScript was cooking with `null` and `undefined` :D), so after some thoughts--one might even say comments--regarding this.

Wunder-Wulfe:
I believe that the implementation could benefit from the use of named arguments like in other languages such as python, where
```py
def mix(alpha, low=0, high=1):
    return low + (high - low) * alpha

mix(0.5) # called with default arguments
mix(0.5, -1.0) # default argument for (high) used
mix(0.5, high=20.0) # called with named parameter

mix(0.5, None) # 'None' is used in place of low, as opposed to '0', results in error

# illegal function definition, results in an error as default arguments cannot be followed by normal arguments
def mix(alpha, low=0, high):
    pass
```

And, in accordance with the way most languages handle default arguments, I am under the assumption that the default value should be passed as an expression. This would resolve any issues that would arise if developers modify the value, and it would allow for operations to be computed within the default argument expression, such as the following:
```luau
local function heapInsert(heap, element, limit=MAX_SIZE-1)
    -- insert item into heap
end

-- later, when developer is adjusting their constants

Heap.MAX_SIZE = 2048
```

However, one caveat with default values is that they are not always a good idea depending on the implementation!
```luau
local function deep_search(tbl, item, visited={})
    --perform recursive deep search, keeping track of visited tables using a lookup
end
```
If the implementation evaluates `{}` once as opposed to re-using the expression, it will result in undefined behavior and memory leaks, as `visited` would function as a static variable, keeping track of the same list if no argument is provided.

InfraredGodYT:
This is a feature I'd love to see. It's practical, simple, and is well-specified (especially the semantics around call-time evaluation and `nil` handling). It also fits well with existing Luau patterns while improving readability and type clarity.

Jediweirdo:
I've also been wanting a feature like this, and at this point, I'm wondering why it hasn't been merged or closed yet. Thank you @nwinn-student for pointing me this way.

My biggest point of contention is the idea of using `nil` to skip arguments. Semantically, a Python-esque approach (named arguments + defaults) seems a lot cleaner: it preserves the ability to intentionally pass `nil` as an actual value, and it avoids pushing the “convenience cost” onto callers/readers (e.g. `doThing("a value", repeatCount = 10)` is much easier to reason about than `doThing("a value", nil, nil, nil, nil, 10)`). That said, I realize named args have been discussed before in #135 and got shot down; if that’s truly off the table, something like this is better than the nothing we have now. Though if there’s any world where #135 can be revisited and combined with this RFC, that would be awesome.

I have three more thoughts from this thread:

1. I agree with @Bottersnike that “snapshotting” variables in default expressions is a bad idea; call-time evaluation seems like the least foot-gun-y option.

2. I’m concerned about how “defaulted parameter => `T?`” interacts with first-class functions and callbacks. Using `T?` (union-with-`nil`) to represent “caller may omit the argument” introduces a conflicting double meaning to `?`:
   - omission (argument may be left out), vs
   - value domain (`nil` is a meaningful value)

   For instance, take a look at @Kampfkarren's example:

   ```luau
   local function g(callback: (number) -> ())
       callback()
   end

   g(f)
   ```

   Note that `f()` is something along the lines of `function f(x: number = 3) end`. The issue they are trying to raise (at least in my eyes) is that there isn't a good way to convey to people that `g()` specifically supports/accepts callback functions with default parameters. If the intended solution is to widen to `(number?) -> ()`, that reads to many people as “`nil` is a valid value to pass” rather than “this argument may be omitted”, which seems misleading—especially if the RFC also treats `nil` the same as omitted for defaulting.

3. I don’t think forwarder inference needs to be solved in this RFC. It’s okay if wrapper functions like `function g(...) return f(...) end` don’t automatically preserve `f()`’s full parameter/return types without explicit annotation (e.g. using type packs) and people must explicitly type wrappers alongside the functions that they are wrapping; the tech for that can be hashed out and added later when there are better ideas on how to handle this.

Bottersnike:
I think the best way to think about default arguments is as pure sugar over doing
```luau
local function foo(x: number?)
    if not x then x = 5 end
    -- use x
end
```
The semantics are entirely consistent with that logic. The point about `number?` not discriminating between "nil has meaning" and "nil is for not providing a value" is a problem already faced, and nothing new for this RFC.

The example provided there should have been

```luau
local function g(callback: (number?) -> ())
    callback()
end
g(f)
```
to make the `number` argument optional, and it doesn't care if it's a `local function foo(x: number?) if not x then ......` or a `local function foo(x = 5)` because both "look" identical from an outsider's perspective (beyond, eg, IDE hover text).

In general the proposed semantics are entirely identical to patterns already widely used, just with syntax sugar for ergonomics and devexp.

While it's now a _very_ outdated branch, the original demo implementation I linked in the OP can be useful for playing around and seeing that it should Just Work. I can't remember if I implemented types there for the new solver or the old.

I do agree named arguments would make this much nicer, but alas we work with what we have, and a big wall of `nil`s as function arguments before you get to the "useful" ones is that. One interesting point though is that often this is worked around with by doing

```luau
local function foo(args: {arg1: number?, arg2: string?, arg3: any?, ...})
```

This RFC as written doesn't provide any convenience helpers for that situation (defaulting individual items within a table argument). While I think that could be an interesting addition, I also think figuring out how to actually do that would be a job for future work.

MagmaBurnsV:
If that example desugared form is what users are expected to build their mental model from, then I think it's worth reconsidering the visibility of prior arguments in the default expression.

It seems intuitive that this:

```luau
local function f(a, b = a)
end
```

Should desugar into this:

```luau
local function f(a, b)
	if b == nil then b = a end
end
```

And this can't be reconsidered after implementation, since changing the visibility rules changes the result of the expression.

Bottersnike:
Honestly it could be cool to allow `(a, b = a)`. Can you think of any good motivating examples?

alexmccord:
I don't think this is the correct desugaring. This conflates the difference between an explicit `nil` vs implicit `nil`.

This should do it:

```luau
local function f(a, ...)
  local b = if select("#", ...) >= 1 then a else nil
  return (function(a, b)
    -- impl
  end)(a, b)
end
```

Basically, you just take out the arguments that has defaults, replace it with varargs, then measure the nargs.

Bottersnike:
The reason that's the desugaring is for the `foo(a=1, b=2, c=3)` `foo(10, nil, 30)` case. There's no way to get a default `b` without an explicit `nil`. It was a compromise designed in with intention from the start to work around the absence of named arguments in function calls.

alexmccord:
So that's the difference between two somewhat related ideas, but from _opposite_ sides! Default arguments (this RFC, I think) is the one that my desugaring solves for, and _named_ arguments is what you're thinking of. You'd write that one as `foo(a, c=30)`. Passing `nil` is genuinely different from implicit `nil`s. That's why VMs typically have some kind of `undefined` value as a sentinel that differentiates the two, and whenever the VM sees an `undefined` value, it replaces it with `nil` (in certain positions).

Bottersnike:
This RFC is intentionally making a trade-off here. Ideally yes, we'd be able to do `foo(a=10, c=30)` and, if we wanted explicit `nil` to remain distinct from omission, we'd write `foo(b=nil)`.

The problem is that this simply is not a thing in Luau today. It came up in #135, which is now closed. I fully agree named arguments would make this much nicer, but within the semantics of the language as it currently exists, that just isn't practical.

Given this pattern is primarily replacing the already-common:

```luau
local function foo(x: number?)
	if x == nil then
		x = 5
	end
end
```

I don't think treating explicit `nil` identically to omission is a particularly painful compromise.

Luau does not currently have an `undefined`-style sentinel value, only `nil`, and types like `number?` already do not distinguish between "unspecified" and "`nil` provided". This RFC is intentionally extending those existing semantics rather than introducing a second notion of absence into the runtime and type system.

Explicit `nil` being treated the same as omission is absolutely a trade-off, but I think it's a pragmatic one that fits the language as it exists today.

(With all the said, I would absolutely love to see #135 or a new RFC for the same job become a thing, and block this RFC on that, because then we can avoid this compromise entirely; I just don't know if that's something we can reasonably expect to happen)

PR #2, opened 2023-10-30
Loosening the recursive type restriction

asajeffrey:
Rendered: https://github.com/asajeffrey/rfcs/blob/recursive-type-unrestriction/docs/recursive-type-unrestriction.md

PR #92, opened 2025-01-15
Relax the recursive type restriction.

aatxe:
[Rendered](https://github.com/aatxe/luau-rfcs/blob/awe/relax-recursive-type-restriction/docs/relax-recursive-type-restriction.md).

### DIFF
--- a/docs/relax-recursive-type-restriction.md
+++ b/docs/relax-recursive-type-restriction.md
@@ -52,9 +52,29 @@ gained for users of the type system.
 
 ## Alternatives
 
-The only serious alternative besides doing nothing at all would be to implement equirecursive types.
+
+### Equirecursive Types
+
+The main alternative besides doing nothing at all would be to implement equirecursive types.
 Equirecursive types have considerably greater complexity, both for humans and for computers. They
 pose considerable algorithmic challenges for both type checking and type inference, and require
 an implementation of a canonicalization of recursive types that is O(n log n). Isorecursive types
 are simpler, map closely to what is already implemented in Luau today, and match the intuitions
 users likely have from interacting with recursive types in nominal object-oriented languages.
+
+### Strict type aliases with Caching
+
+[A previous RFC](https://github.com/asajeffrey/rfcs/blob/recursive-type-unrestriction/docs/recursive-type-unrestriction.md)
+proposed loosening the recursive type restriction, but still leaving it partially in tact, by implementing a strategy of
+mapping distinct recursive uses of a type alias to a free type during the definition portion of constructing the type alias
+and then tying the knot using that mapping as a cache. This feels in the spirit of Landin's Knot, and was a meaningful
+proposal to improve the status quo without us implementing a new type inference engine, but as mentioned in the drawbacks
+section, this work is already well underway for other reasons and so the appeal of taking this route today is lessened.
+
+### Lazy recursive type instantiations with Caching
+
+The previous RFC also mentioned briefly both the approach proposed by this RFC and an approach that uses a cache as well
+to limit the size of the graph. This may be useful, and may in fact be what we implement, but the author of this RFC does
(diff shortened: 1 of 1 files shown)


docs/relax-recursive-type-restriction.md
```diff
@@ -0,0 +1,60 @@
...
+
+Eliminating this restriction means committing to an intrinsically more complicated architecture for
+type inference in Luau in general. The Luau team, however, has already made the decision to do this,
+and so there seems to be little drawback beyond the obvious implementation work required in making
+type alias expansion completely lazy. This pays for itself in the considerable gain in expressivity
+gained for users of the type system.
+
+## Alternatives
+
+The only serious alternative besides doing nothing at all would be to implement equirecursive types.
```
hgoldstein, line 55: Maybe more of a nit: don't we already have equirecursive types? I would think that any class-ish looking table would be one.

vegorov-rbx:
Relation to https://github.com/luau-lang/rfcs/pull/2 could be noted.

aatxe:
> Relation to #2 could be noted.

Honestly, did not remember that it existed, but the relation between the two is that the old one was proposing a caching strategy to allow finite expansions that would work as a mild relaxation of the already in-place restriction. This is proposing what was listed as the top alternative (infinite expansion). I don't know whether we will or won't implement a cache for that expansion, but frankly, I think that caching or not is an implementation detail and not a part of the language design proposal. We probably do not want to implement caching in the type system that has an impact on the static semantics.

vegorov-rbx:
I view it as an implementation detail for the old solver similar to how your proposal mentions `by treating type aliases as proper type functions that are lazily expanded when their contents are needed` in the new solver.

I don't think it's that bad to explain it, but it could have been in the PR desc as opposed to the RFC body.
Didn't feel to me that described caching there was particularly arbitrary.

I'm also not saying that it has to be made that way, but it is prior art for the same problem and I would say that RFC was already accepted in spirit, but we switched gears on the implementation.
If new solver has a simpler implementation without exploding, I'm all for this RFC as well.

docs/relax-recursive-type-restriction.md
hgoldstein: Talked offline: you could consider type inference as constructing equirecursive types, but this PR is about relaxing the current restrictions on recursive types, and saying we will implement such via isorecursion.

aatxe: In some sense, type _inference_ will sometimes construct a recursive type that you could probably consider equirecursive, e.g. `t1 where t1 = { foo(self: t1) -> ... }` but this is one of the things that hampers the developer experience by leading to complex types, and is not supported by any of our syntax for actually annotating types. It's strictly a representation of a cyclic type that has occurred because of inference. #5 should also hopefully help us get rid of those patterns as well.

docs/relax-recursive-type-restriction.md
```diff
@@ -0,0 +1,80 @@
...
+```luau
+type Tree<T> = { data: T, children: {Tree<T>} }
+```
+but the [recursive type restriction](recursive-type-restriction.md) limits us to supporting only
+direct recursion as above, where the type aliases' parameters remain unchanged.
+
+At the time the restriction was implemented, Luau featured a greedy type inference engine that would
+eagerly expand every occurrence of a type alias, and, as such, would infinitely expand until
+ultimately crashing from overflowing the stack, or would hit a recursion limiter and produce a
+strange, very deep type.
```
WheretIB, line 25: This is not correct, old solver did not behave this way.

aatxe: What did it do then? As far as I could tell from reading old discussions from @asajeffrey, this was the case, but if you know something that I don't, happy to change it.

aatxe: Thanks, I'll revise the text then.

vegorov-rbx (approved):

PR #93, opened 2025-01-24
Fix `table.newtype` argument

Ketasaja:
[source](https://github.com/luau-lang/luau/blob/c13b5b7440a65987ee9dd592ee468d421d83f223/Analysis/src/TypeFunctionRuntime.cpp#L454)


aatxe (approved):
Thanks for the fix!

PR #94, opened 2025-01-26
Multiple indexing

dphblox:
An attempt at a comprehensive continuation of #24, addressing all concerns put forward in the thread.

[Rendered](https://github.com/dphblox/rfcs/blob/destructuring/docs/syntax-multiple-index.md)

### DIFF
--- a/docs/syntax-multiple-index.md
+++ b/docs/syntax-multiple-index.md
@@ -79,12 +79,12 @@ local numbers = {3, 5, 11}
 local three, five, eleven = numbers[1], numbers[2], numbers[3]
 ```
 
-Arrays can read out an inclusive range of values from consecutively numbered keys. This is specified with `[x : y]`, where `x` and `y` evaluate to a number.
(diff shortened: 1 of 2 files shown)


docs/syntax-multiple-index.md
```diff
@@ -0,0 +1,413 @@
...
+local numbers = {3, 5, 11}
+
+local three, five, eleven = numbers[1, 2, 3]
+```
+
+This will desugar to this behaviour exactly:
+
+```Lua
+local numbers = {3, 5, 11}
+local three, five, eleven = numbers[1], numbers[2], numbers[3]
```
vegorov-rbx, line 79: Really doesn't feel right to cause such a confusion with multi-dimensional array indexing for such a marginal feature.
(which might not exist today in Luau, but this will also closes the door on that)

Doesn't feel natural overall as well, but this part is my opinion.

dphfox: Got it, I'll look in other directions then :)

dphblox:
Closing to rework.

PR #95, opened 2025-01-29
Structure matching syntax

dphblox:
Gathering consensus on the destructuring syntax inside of the braces (i.e. punting on other issues explicitly), so we can align on this with the community.

[Rendered.](https://github.com/dphblox/rfcs/blob/destructuring/docs/syntax-structure-matching.md)


docs/syntax-structure-matching.md
```diff
@@ -0,0 +1,358 @@
...
+*Open question: are we OK with having no delimiter between key and name? Discuss in comments.*
+
+```Lua
+{ [1] foo, [#data] bar }
+```
+
+This desugars to:
+
+```Lua
+foo, bar = data["foo"], data[bar()]
```
jackdotink, line 102: I think you meant to do `data[#data]` rather than `data[bar()]`

docs/syntax-structure-matching.md
```diff
@@ -0,0 +1,358 @@
...
+
+#### Basic matching
+
+This is the most verbose, but compatible way of matching values.
+
+Keys are specified in square brackets, and are allowed to evaluate to any currently valid key (i.e. not `nil`, plus any other constraints in the current context).
+
+An identifier is specified on the right hand side, showing where the value will be saved to.
+
+*Open question: are we OK with having no delimiter between key and name? Discuss in comments.*
```
jackdotink, line 93: I'm okay with this but I think `name = index` is better:
```luau
{ foo = [1], bar = [#data] }
```
I have no good reason other than it makes me happy, overall pretty unimportant.

docs/syntax-structure-matching.md
```diff
@@ -0,0 +1,358 @@
...
+
+```
+foo, bar = data["foo"], data["bar"]
+```
+
+#### Consecutive keys
+
+Consecutive keys can be implicitly expressed by dropping the key.
+
+*Open question: are we OK with this in the context of dot keys without names? Discuss in comments.*
```
jackdotink, line 155: I am not okay with this *yet*. At the moment there is no such thing as a table tuple type, which means this syntax wouldn't see much use, and would probably be used on accident more often than not.

docs/syntax-structure-matching.md
```diff
@@ -0,0 +1,358 @@
...
+foo, bar = data[1], data[2]
+```
+
+#### Nested structure
+
+A structure matcher can be specified on the right hand side of a key, to match nested structure inside of that key.
+
+An identifier and a structure matcher cannot be used at the same time. Exclusively one or the other may be on the right hand side.
+
+*Open question: if we add a delimiter between key and identifier, do we need a delimiter here too? Discuss in comments.*
```
jackdotink, line 179: I would say yes:
```luau
{ myBar = [1]["foo"]["bar"] }
{ myBar = [1].foo.bar }
```
Maybe I'm diving off the deep end with this one.

docs/syntax-structure-matching.md
dphblox: good catch!

docs/syntax-structure-matching.md
dphblox: I considered this, I'd like to see what others think too. There's a few colours we could pick for this bikeshed

docs/syntax-structure-matching.md
dphblox: I'm not sure what you mean by table tuple type, could you expand on this?

docs/syntax-structure-matching.md
dphblox: That's an interesting alternate suggestion, but I wonder whether fully qualifying members like this is a good idea in all cases. For example, what if all the variables you're capturing exist multiple layers deep? Does this introduce too much repetition, and isn't the point of destructuring to reduce the prevalence of fully qualified paths in the first place?

docs/syntax-structure-matching.md
jackdotink: Yeah, equivalent to a tuple type in typescript that is actually just an array, it could look something like `{ number, string, boolean }` where `t[1]` is a number, etc.

Without this kind of type system support, this would only be used for indexing arrays multiple times at multiple indexes, which I don't really do.

docs/syntax-structure-matching.md
jackdotink: I see your point, maybe something more like this?
```luau
{ { { myBar = ["bar"] } = ["foo"] } = [1] }
```
This reverses the order that indexes are written in though, which I think sucks. I'm not really sure what to do here.

docs/syntax-structure-matching.md
jackdotink: Yeah; at the moment I don't think it will see any meaningful usage.

docs/syntax-structure-matching.md
dphblox: I don't think it's out of the question to simply keep the nesting syntax as-is while having delimiters for individual values.

docs/syntax-structure-matching.md
dphblox: Got it. I think I'll leave it in for now, but perhaps I can amend the section discussing this.

docs/syntax-structure-matching.md
jackdotink: Another option would be using `index in name`, which also nicely solves my little dilemma with nested matching:
```luau
{ [1] in foo, [#data] in bar }
{ [1] in { ["foo"] in { ["bar"] in myBar } } }
```

dphblox: That might make it sound like you're reading from the right hand side, e.g. `[1] in foo` == `foo[1]`.

Would a syntax like this be more suitable?

```Lua
{ [1] = foo, [#data] = bar }
{ [1] { ["foo"] { ["bar"] = myBar } } }
```

Which makes it clear where bindings are being made vs where nested structures are being explored.

dphblox: Amended to add `=` delimiter for bindings. Visiting nested structures does not use `=` for now as it's an open question whether we should use `=` for non-assigning operations.

I can see examples where it may read less clearly:

```Lua
local { .foo = { .bar } } = { foo = { bar = 2 } }

print(foo) -- nil or { bar = 2 }?
```

docs/syntax-structure-matching.md
dphblox: As mentioned in the above comment thread, I amended the proposal to include delimiters for binding individual values, but kept the nesting syntax as-is. There's a possibility that introducing a syntax distinction between binding and visiting could resolve this downside.

docs/syntax-structure-matching.md
hgoldstein: My bias would be for syntax that mirrors construction: this is what other languages with pattern matching tend to do.
```ocaml
type Foo = {
  bar: int;
  baz: string;
}

let my_foo : Foo = { bar = 42, baz = "baz!" }

(* x = y means "bind field x to variable y", 
 * z by itself means "bind field z to the variable z" 
 *)
let print_foo f = match f with
| { bar = bound_bar; baz } ->
  Printf.printf "bar = %s, baz = %d\n" bound_bar baz
```
Rust looks more or less the same.

I talked with @dphblox offline about whether this is confusing due to using `=` in an assignment position. To me the _lack_ of delimiter makes the structure look nonsensical in Luau. 

Just to highlight: this is a case where bikeshedding matters more than the average language feature. We're talking _about_ syntax: the discussion is "what color bike shed would be best for our users?"

docs/syntax-structure-matching.md
```diff
@@ -0,0 +1,355 @@
...
+
+```
+foo, bar = data["foo"], data["bar"]
+```
+
+#### Consecutive keys
+
+Consecutive keys can be implicitly expressed by dropping the key.
+
+*Open question: are we OK with this in the context of dot keys without names? Discuss in comments.*
```
hgoldstein, line 153: Yeah this one concerns me. If you come from another language (specifically: JavaScript), you might assume:
```lua
local t = { foo = 42 }
local { foo } = t
print(foo)
```
... prints `42`, which it wouldn't. I think if this proposal went through I'd ask that we _require_ the dot and disallow anonymous binding to numbered indexes. e.g. one _must_ write:
```lua
local { [1] = foo } = { 42 };
```

dphblox: I'd be OK with that, given that for tuples we could always fall back on multiple returns:

```Lua
local id, text, isNice = unpack({1, "foo", true})
```

dphblox: If we really wanted to, we could even just drop that into the matcher. I don't propose exactly this but you get the idea:

```Lua
local { unpack(id, text, isNice) } = {1, "foo", true}
```

dphblox: I amended the proposal with the unpack syntax for now to solve this ambiguity, we can decide separately whether we want to keep it in the proposal or yeet it.

dphblox:
From my POV, the thing that sunk the previous RFCs were exactly the questions that the `unpack` syntax and similar features answer. From Andy's prior comments, it's clear that the Luau team need to feel confident that destructuring can actually serve as a strong language feature before its accepted, so I disagree that less detail is needed here. In fact, we should be proving to satisfaction that we can avoid all of the pitfalls.

dphblox:
Dropping the `unpack` syntax from the spec in favour of basic indexing.

docs/syntax-structure-matching.md
dphblox: To make the nesting syntax more clear, would we prefer to use the `in` keyword to denote visiting? Some possibilities:

```Lua
{ .foo, in .bar { .baz } }
{ .foo, { .baz } in .bar }
```

This would desugar to:

```Lua
local foo, baz = data.foo, data.bar.baz
```

Notably, the nested structure would not be bound.

rihok:
Coming into this a bit late I guess, but wanted to just ask some questions. Is there some issue with copying the JavaScript syntax wholesale? It's quite well thought out and tried and tested. For example one question I have is, why is it necessary to have a dot prefix for the destructured keys instead of:
```
-- keyed table desctructuring
local { foo, bar } = t

-- indexed table destructuring
local a, b, ...rest = t
-- or
local [a, b, ...rest] = t
```
for indexed table destructuring?

dphblox:
I don't entirely disagree with that idea personally, but IIRC others shot this down already.

The main trouble comes from the fact that array literals would look the same as keyed table destructuring:

```Lua
-- this would not work as expected, despite looking visually similar
local { foo, bar } = { "foo", "bar" }
```

This was considered a showstopper for that syntax in particular.

Beyond that, the square brackets idea has been floated a few times, but each time there's generally some disapproval since the only other time Luau uses square brackets is for indexing; we dont use `[]` to declare arrays like JS does.

I think JS has a well designed syntax for JS, but in the context of Luau there's multiple arguments against some of its choices.

As for:

```Lua
local foo, bar, ...rest = t
```

This would be backwards incompatible, as Luau supports multiple returns. Today, this runs:

```
local foo, bar, baz = { "hello" }

print(foo, bar, baz) --> { [1] = "hello" } nil nil
```

surfbryce:
I very much prefer something along the lines of what @rihok suggested at the very bottom of the code sample, possibly something that outlines similarly to:
```luau
local KeyedTable = {
    Hello = true;
    [5] = false;
    [Instance] = 5;
}
local PackedTable = table.pack(1, 2)
local NestedTable = {
    Nested = {
        World = true;
    };
}

local [ ["Hello"] = A, [5] = B, [Instance] = C ] = KeyedTable
local [ [1] = D, [2] = E, ["n"] = F ] = PackedTable
local [ ["Nested"] = [ ["World"] = G ] ] = NestedTable
```

I assume square brackets here would be an understandable extension of the Luau language given that square brackets are **only** used for indexing (a point illustrated by @dphblox) and table destructuring is simply an abstraction of that.

A few things to touch on surrounding this:
1. Most of the time destructuring will be done on string literal indices, so potentially string literals can be written without the index syntax to appear as:
```luau
local [ Hello = A, [5] = B, [Instance] = C ] = KeyedTable
```
2. There is concern about whether or not the equals sign could indicate assignment. Another alternative is to use colons instead, this would align with the ideology that table types use where we understand it as mapping a property to something else (in the case of table types, another type): so - in this case - we would be mapping a property to a variable name.
```luau
local [ Nested: [ World: G ] ] = NestedTable
```
3. When used for global variable instantiation we run into the same issue we have with parentheses immediately following an expression irrelevant to it:
```luau
local A = 5
(else :: Type).Call()
```
In this instance we would get an error since this would be interpreted as calling the number 5 defined immediately before it. This would be solved by adding a semi-colon to the end of the 5 to indicate the expression is finished. The same would have to be done in the case of destructuring to the global scope, I don't see this as a drawback given this is an already understood behavior of the language.

I don't know what the implications are on parser complexity in regards to this or whether or not this fits within the purview/scope of this RFC. But I believe it's a fun and reasonable suggestion purely in terms of syntax.

docs/syntax-structure-matching.md
```diff
@@ -0,0 +1,344 @@
...
+To save the value at that key, an `=` is used, and an identifier is specified on the right hand side, showing where the value will be saved to.
+
+```Lua
+{ [1] = foo, [#data] = bar }
+```
+
+This desugars to:
+
+```Lua
+foo, bar = data["foo"], data[#data]
```
jkelaty-rbx, line 101:
```suggestion
foo, bar = data[1], data[#data]
```

jkelaty-rbx: Or maybe you meant for the example to be:
```luau
{ ["foo"] = foo, [#data] = bar }
```

docs/syntax-structure-matching.md
jkelaty-rbx: The identifier being on the right hand side is very confusing imo. Arguably something that can be learned, but certainly unintuitive and very likely to trip people up, especially given that with this RFC, a Luau program may now consist statements of both left-hand and right-hand identifiers (at least, I could see myself getting tripped up on it)

docs/syntax-structure-matching.md
dphblox: Amended the proposal with key chaining

dphblox:
> I assume square brackets here would be an understandable extension of the Luau language given that square brackets are only used for indexing (a point illustrated by @dphblox) and table destructuring is simply an abstraction of that.

I would be somewhat OK with this, but it's a little weird that we would use square brackets both to surround the keys and to surround the whole matcher.

Perhaps it would help with the "array problem":

```Lua
local [ these, are, keys ] = data
local { these, are, consecutive, values } = data
```

But this would still be inverse of JS, which could be a stumbling block. IIRC, this is part of what caused the original RFC to fail, hence the dot prefixes. The older RFC has more details on that whole thing.

> There is concern about whether or not the equals sign could indicate assignment. Another alternative is to use colons instead

Colons are used for type annotations, so it would be almost certainly inappropriate to use them here.

dphblox:
I think I'll try and write down my full chain of thoughts at the moment around the array and reassignment comments from the old RFC, as well as some syntax comments more generally.

Unlike JS, Luau allows you to shuttle around multiple values at once, making unpacking the natural paradigm for array destructuring:

```Lua
local these, are, consecutive, values = table.unpack(data, 1, data.n)
```

So if we were to suggest any syntax there, it really feels like it ought to be expression-side, not assignment-side:

```Lua
local these, are, consecutive, values = ...data
```

Naturally I would be inclined to extend that to keys (thus making this RFC redundant), but that introduces the problem of having to specify the keys twice in almost all situations:

```Lua
local these, are, keys = ...data["these", "are", "keys"]
```

That's the grounds upon which I think the original destructuring RFC was made, and probably what we should focus on solving. That also aligns with the discussions we've had above. Trying to make this work for arrays seems misguided and I don't reckon it's even relevant.

Now onto syntax. One of my earlier explorations was having a "destructuring assignment", though I was concerned about it looking too much like a compound assignment, or like it's unpacking values positionally:

```Lua
local these, are, keys ...= data
```

In my eyes, the problem there is that throughout the rest of Luau, the names of those identifiers are not significant - only their positions are. Therefore, if the names are going to be significant, it's worthwhile to alter how those names are presented to make that clear. That's what the original RFC did.

Dot prefixes seem like a sensible and low-syntax way to indicate those names are keys. Also, since the `=` is behaving differently in this case (and would likely have different requirements, such as not accepting multiple RHS values), it makes sense to differentiate that too:

```Lua
local .foo, .bar, .baz in data
```

Of course, without the `local` indicating this is a new variable declaration, this would be ambiguous syntax. I personally only believe there is one valid use case for that, though, and that would be forward declaration of uninitialised variables:

```
local thing

local function usesThing()
    task.delay(1, function()
        print(thing)
    end)
end

local otherThings, foo, bar
.thing, .otherThings, .foo, .bar in getThings()
```

Forward declaration is a little bit of a code smell IMO (though not always!), it feels unintuitive for people to destructure in this position, and JS doesn't support destructuring when assigning - only when declaring - so I would be personally OK dropping this requirement, though I'm not the person who brought it up in the first place.

So all of that is my bias coming into this. I'm not saying we should do any of the above, but that's the general direction I lean in philosophically.

docs/syntax-structure-matching.md
jkelaty-rbx: > I don't know of a language that has identifiers on the left-hand-side.

To be clear, my comment about "right handed" vs "left handed" identifiers was more in reference to the position of the variable in an assignment statement (or something that looks like an assignment statement). In this case, I'm thinking of statements that have the form `local var = <expr>` where `var` is the thing that is on the left hand side of the equals sign.

The original proposal (which has since been modified) was to allow for a similar syntax, but where the name of the variable that will store the expression is on the right-hand side of the equals sign. The fact that we had something that syntactically resembled an assignment statement but with different positioning of the resultant variable name is what I was calling out as problematic. The example you shared for JS is using a colon which I think is reasonable to use if you really really wanted it to be "right-handed" (although imo it is equally as confusing to a newcomer, but we can discuss that separately).

hgoldstein: >  The fact that we had something that syntactically resembled an assignment statement but with different positioning of the resultant variable name is what I was calling out as problematic.

In my experience: languages which implement destructuring, or pattern matching more generally,  make the choice of the syntax for construction mirroring the syntax for destruction. The reason JavaScript uses a colon is because that's how you construct objects in JavaScript. Rust also uses a colon because that's how you make a struct in Rust, OCaml uses an equals because that's how you make a record in OCaml. If table literals in Luau looked like:
```
local t = {
  x => 42
}
```
... then my suggestion for this syntax would be:
```
local { x => bound_x } = t
```

jkelaty-rbx: Okay, you've opened my eyes a bit on this. I think the idea of table construction and destructuring having identical syntax is a really elegant idea. I still feel this will be confusing due to the fact that for that grammar, a declaration and "field assignment" have very different semantics. I think that pushes me toward the feeling that we shouldn't add this feature into Luau at all though.

If we proceed with this feature anyway, should any RFC which proposes sugar for destructing syntax (e.g. `local { .namedField } = struct`), also include extending this sugar for table construction (e.g. `local struct = { .namedField }`)?

dphblox:
I cleaned up the proposal - I would consider this to be the apex of what an agreeable _100%-brace-based_ syntax would look like.

To summarise, here's what's left at this point:

- Identifiers on the left, fully qualified paths on the right: `{ foo = [1], bar = [2] }`
- Dot keys can be used instead of string indexes: `{ foo = .red, bar = .blue }`
- Both kinds of key can be chained: `{ foo = [1].bar.baz }`
- If a dot key is used as the final index, the name can be inferred: `{ .foo, .bar[2].baz }`

Things that have been explicitly rejected;

- Positional array/tuple unpacking. 
    - This was mentioned as a potential showstopper last RFC, but almost everyone I've talked to since then seems to have changed their mind. It's clear that nobody on the Luau team has the appetite to ever support `{ foo, bar }` even if we did want to do array unpacking, and that alternate syntax (`{ unpack foo, bar }` / `{ foo = [1], bar = [2] }`) or just standard assignment (`local foo, bar = table.unpack(data)`) more than suffices for both arrays and tuple-style tables.
- Nested matchers. 
    - As it stands in this proposal, the only way to match nested structure is to chain together keys (`.foo.bar.baz`). Matchers cannot be put inside of other matchers - this was deemed to be too confusing especially when considering questions around whether those nested matchers would introduce a member to the namespace. This can always be revisited later, but for now there's not enough motivation to include it.

There's some other suggestions in this thread for syntaxes based on brackets (e.g. `[ foo, bar ]`) that haven't really been discussed in as much depth, so I think that's the next topic to tackle here. If we're not interested in incorporating brackets, I think we're about as far as we'll get with substantial design changes - the rest will be bikeshedding.

dphblox:
All this being said, I'm personally less convinced that we even need everything that's in this RFC as it stands. Given the direction the discussions have drifted towards, combined with the practical examples that've arose, I struggle to see much of the use case beyond simple named destructuring. There's an argument to be made that we should elide much of the complex concerns and stick to something much simpler to retain the spirit of the Luau language.

Much of what's in here is my attempt at serving as many use cases as possible, but if we wanted to get most of the value with less complexity, we could just do something like

```Lua
local .foo, .bar in data
```

and accept that we'll never have a full destructuring system, primarily because we'll never need one.

hgoldstein:
> Identifiers on the left, fully qualified paths on the right ...

Check out the conversation about that (. We shouldn't do this because, AFACT, all other languages that implement some sort of pattern matching opt to mirror their construction / destruction syntax.

> ... the primary missing use case is for named destructuring, and I feel like we should be careful not to lose sight of that.

I chatted about this offline with some folks (mainly @aatxe): destructuring syntax / pattern matching in general is a power user feature. The _unfortunate_ part is that it's also the principled way to do named imports without adding a feature for it, such as [JavaScript imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import).

I wonder if we should support exactly one syntax to start:
```
local { a, b, c = foo } = { a = 3, b = 4, c = 5 }
print(a, b, foo) -- prints 3 4 5
```
No nested syntax, no support for custom indexers, no support for array-like members. As-is I think the `local {` prefix should be unambiguous, you just need one token of lookahead (which we already have, pretty sure): I don't think taking this as a first step would preclude us from expanding the syntax later. The thing we *would* lose would be committing to a potentially ambiguous syntax. I've been convinced that the existence of `table.pack` / `table.unpack` means that the syntax above _won't_ be confused for an array-like de-structuring, and type checking can _hopefully_ push folks in the right direction.

dphblox:
> Identifiers on the left, fully qualified paths on the right ...

> We shouldn't do this because, AFACT, all other languages that implement some sort of pattern matching opt to mirror their construction / destruction syntax.

Yeah this is part of the bikeshedding we need to figure out still. I just tried to match the mood of some other comments but don't personally have a strong position on which I'd prefer.

I will say that this feels unclear to me:

```Lua
local { foo = bar } = getData()
```

It isn't obvious which side is the identifier and which side is the key. Since this isn't an ambiguous position, a contextual keyword would help clarify this and give a natural "orientation" to the statement. I also think it's more "Luau-like" to use keywords over symbols.

```Lua
-- for demonstration purposes
local { foo as bar } = getData()
local { foo in bar } = getData()
```

But the ambiguity could get resolved by symbols too, just more noisily:

```Lua
-- for demonstration purposes
local { foo -> bar } = getData()
local { foo = .bar } = getData()
local { .foo = bar } = getData()
local { [foo] = bar } = getData()
-- etc...
```

I think I'll just defer to the better judgement of the team for this one, but my personal preference would be `as` or `in` since they naturally give that "orientation" without introducing line noise or potentially easy-to-miss symbols.

dphblox:
> I wonder if we should support exactly one syntax to start:

Yeah this is the way I'm leaning at this point too. I might pare down the RFC to literally just this.

I'm happy we spent the time to explore the other options if only so we could figure out why not to pursue them.

dphblox:
Alright, I've whittled down the RFC to just the basic feature as suggested. Happy to integrate further thoughts if desired :)

deviaze:
Nice whittle! Could we also note syntax like
 
```lua
local .foo, .bar in data
local .foo, .bar = data
local foo, bar in data
```
in the alternatives section?

In drawbacks, maybe note that `local { foo, bar, meow } = data` syntax might not be as intuitive to users of only/mainly Luau, because it doesn't convey Luau indexing as clearly as alternatives like `local .foo, .bar in data` do. 

As an another alternative, we could generalize `.foo, .bar in data` into a multiret expression like `data.foo, data.bar`,
with the syntactic sugar of `local .foo, .bar in data` being similar sugar to it as `local function foo() end` is to local function initialization and assignment.

Since `local function` sugars local assignment without requiring an `=`, I think a destructuring assignment syntax without `=` is perfectly reasonable. Any ambiguity in the expression context (nested `.x in y` exprs) should be resolvable with parens.

For example:

```luau
local x = .x in somevector
assert(x == somevector.x)
-- w/ syntax sugar:
local .x in somevector
local .x, .y in somevector

type Profile = {
    ip: string,
    hostname: string,
    last_seen: DateTimeUtc,
}

type DateTimeUtc = {
    unix_timestamp: number,
    format: (self: DateTimeUtc, fmtstring: string) -> string,
}

local profile: Profile

local ip, hostname, timestamp = .ip, .hostname, .last_seen.unix_timestamp in profile

-- w/ sugar
local .ip, .hostname, .timestamp.unix_timestamp as timestamp in profile -- for demonstration purposes; `as` might not be the right choice here `as` usual
```

PR #96, opened 2025-01-29
Update read-only and write-only syntax

alicesaidhi:
This updates the syntax mentioned on the read-only and write-only files to the new, current syntax.


aatxe (approved):
Thanks so much!

PR #97, opened 2025-02-01
Scoped Module Imports

jrmelsha:
[Rendered](https://github.com/jrmelsha/luau-rfcs/blob/master/docs/syntax-from-require.md).

### DIFF
--- a/docs/syntax-from-require.md
+++ b/docs/syntax-from-require.md
@@ -1,0 +1,91 @@
+# Scoped Module Imports for Luau  
+
+## Summary  
+
+Introduce a new syntax for scoped module imports in **Luau**, allowing selective imports of module members without requiring manual `local` assignments. This would streamline working with many dependencies, making code more readable and maintainable.  
+
+## Motivation  
+
+Currently, Luau requires modules to be imported using `require`, followed by manual assignments:  
+
+```lua
+local Pet = require(game.ReplicatedStorage.Library.Items.Pet)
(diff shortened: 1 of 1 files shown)


jackdotink:
The type of imports used throughout this RFC are roblox's instance paths, which do not exist in Luau. Luau has imports with string requires.

jrmelsha:
> The type of imports used throughout this RFC are roblox's instance paths, which do not exist in Luau. Luau has imports with string requires.

Thanks! - no problem, we can just substitute instance paths with string paths for the non-Roblox version.

quantix-dev:
Great RFC. But, I believe that turning require into an explicit keyword would bloat luau's currently small number of keywords and confuse the matter where it was once a function. Instead, I think adding destructuring to objects would preserve the simplicity of require while offering cleaner module imports.

ishtar112:
In my eyes, this problem is pretty eloquently solved with user-defined aliases. To use your example, though user-defined aliases are not yet available in ROBLOX, we can imagine:
```luau
local Pet = require("@items/Pet")
local Currency = require("@items/Currency")
```
which is comparatively more verbose than your proposal but does indeed do away with the majority of the boilerplate, and it does so without such convoluted and controversial changes to the language.

In fact, this problem has always been solved quite well by simply using more variables:
```lua
local Items = game.ReplicatedStorage.Library.Items
local Pet, Currency = require(Items.Pet), require(Items.Currency)
```

PR #98, opened 2025-02-03
Resolve exported type of same name

NightrainsRbx:
[Rendered](https://github.com/NightrainsRbx/rfcs/blob/resolve-exported-type-of-same-name/docs/resolve-exported-type-of-same-name.md)

### DIFF
--- a/docs/resolve-exported-type-of-same-name.md
+++ b/docs/resolve-exported-type-of-same-name.md
@@ -1,0 +1,53 @@
+# Resolve exported type of same name as the module implicitly.
+
+## Summary
+
+Remove redundancy when using an exported type that shares the same name as its module by implicitly resolving the type with the module's name.
+
+## Motivation
+
+It is common to export a type that represents what the module returns, especially in OOP. However, when using that type, you end up repeating the same name, ie: `Maid.Maid`, `Signal.Signal`, `Promise.Promise`.
+
+This undoubtedly feels redundant and like boilerplate. While defining type aliases locally can mitigate this, they are also boilerplate.
+
+## Design
+
+When the name of a variable containing a module is `Something`, the type `Something` alone resolves to `Something.Something`, provided that:
+
+- The type `Something` does not already exist locally or globally.
+- The module exports a type named `Something`; otherwise, a type error is raised.
+
+This is based on the name of the variable, not the name of the module's script or file.
+
+For example:
+```luau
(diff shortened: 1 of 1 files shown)


surfbryce:
Having something to simplify this would be extremely valuable. I used to name these primary types just "Type" so that externally I could do "Module.Type", but of course then you lose the naming context if you use this in a function argument for example.

I think another good option is if we include a new token for type exports, like if we have "export default type Name = ...." or anything similar. This makes it so that we don't rely on the file name matching or any other unexpected behavior to have it automatically assume a type.

NightrainsRbx:
> I think another good option is if we include a new token for type exports, like if we have "export default type Name = ...." or anything similar. This makes it so that we don't rely on the file name matching or any other unexpected behavior to have it automatically assume a type.

I think that relying on the variable's name (as the design specifies) would be less prone to unexpected behavior than using the module's name. Still, I don't think it's dangerous to infer the type from the variable's name or the module's name, in the end, anything you change in a module will affect other scripts that depend on it

Your idea is interesting but has the same problems as the alternative design, and my intention with this feature is more like a syntactic sugar.

Crazyblox:
I raise a few concerns regarding this.

- Luau may become less human-readable if accepted as common practice.
- There would be less incentive for module hierarchies to be well-structured.
- Mental burden when reading unfamiliar code increases, adding friction to familiarising oneself with new code.

If a type is assigned to something and the path doesn't seemingly-originate from a module, I would assume that the current script I am viewing is responsible for that type; I can see this leading to false debug sessions where all that was missing was the path to the module for said type.

If I have 10 modules required by the script that I am reading and they all implicitly refer to export types, I'd lose all surface-level context of what comes from where, and the structural benefit of how all the code is put together wouldn't be so apparent.
Seeing the exact type path is something I believe helps make Luau very readable.

For behaviour where this is 100% required, you have mentioned in your RFC that it can be locally defined, which is an explicit communication from within the code that this type is coming from a specific location, removing ambiguity and ensuring that this behaviour is exposed for others to read and follow, solving the human-readability issue.

As for the issue regarding types having the same name as the module itself, this seems like more of a fundamental design issue from said modules rather than how Luau operates; Unless there's a valid reason that types have to be the exact same name as the module which holds it, Something.Something could easily become Something.Standard, Something.Data or any other appropriate type name.

PR #99, opened 2025-02-11
RFC: Glob imports

ffrostfall:
[Rendered](https://github.com/ffrostfall/rfcs/blob/glob-imports/docs/require-by-string-glob-imports.md)

### DIFF
--- a/docs/metatable-type-functions.md
+++ b/docs/metatable-type-functions.md
@@ -74,4 +74,6 @@ The name of `setmetatable` has an interpretation of implying a side effect; tabl
 
 ## Alternatives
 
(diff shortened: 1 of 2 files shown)


deviaze:
I think a reasonable choice for the return would a struct-like table with keys being the filenames (sans extension) and values being the return types of the files.

But I don't think a glob require scheme is the right way to go about it; I'd rather have the type system be filesystem-aware so that we can write our own type functions for requiring paths, importing dependencies, and verifying that the user inputted a valid path. 

In seal, for example, I have a thread.spawn function that expects the user to provide the path of another luau file relative to the file being required from. I think I could probably type hack that in with typeof and require now since it follows Luau's require semantics, but that's not a very general solution. It also breaks apart if I'm dealing with other filetypes (if i need to handle json configs, etc.) or deviate from require semantics. If I could check if the file exists in the type system, I could provide accurate and dedicated errors for that function without having to write a separate vscode plugin for seal.

PR #101, opened 2025-02-21
Subtype Checking methods for User-Defined Type Functions

gaymeowing:
This RFC is probably poorly described in some places, but I hope the examples do a better job explaining than I have. If anyone wants to edit this so it makes better sense, feel free.

[Rendered](https://github.com/gaymeowing/rfcs/blob/user-side-subtype-checking-methods/docs/user-side-subtype-checking-methods.md)

### DIFF
--- a/user-side-subtype-checking-methods.md
+++ b/user-side-subtype-checking-methods.md
@@ -1,0 +1,137 @@
+# Subtype Checking methods for User-Defined Type Functions
+
+## Summary
+
+New methods on the `type` userdata in type functions for performing subtype checks.
+
+## Motivation
+
+Currently there is no built in way to do subtype checks in User-Defined Type Functions. This means the developer has to write the following if they wanted to do a subtype check. Subtype checks are useful for type functions, because it makes code like in the following 2 example cases unneeded.
+
+Case 1: Checking if a type is a string or a string singleton
+```luau
+local is_string = t:is("string") or (t:is("singleton") and type(t:value()) == "string")
+```
+
+Case 2: Checking if a type is an enum
+```luau
+type function isenum(t)
+    if t:is("union") then
+        local components = t:components()
+        local components_that_are_string = 0
+
+        for _, component in components do
+            if tv:is("string") or (tv:is("singleton") and type(tv:value()) == "string") then
+                components_that_are_string += 1
+            end
+        end
+        return types.singleton(
+            components_that_are_string == #components
+        )
+    else
+        return types.singleton(false)
+    end
+end
+```
+
+Additionally it would be a pain for users to write a type function to be able to do subtype checks for any type. As the following example type function is already very long, despite not covering everything.
+
+```luau
+type function hassubtypeof(a, b)
+    local false_type = types.singleton(false)
+    local true_type = types.singleton(true)
+    local is_subtype = false
+    local a_tag = a.tag
+    local b_tag = b.tag
+
+    if 
+        (a_tag == b_tag) or
+        (b_tag == "singleton" and type(b:value()) == a_tag) or
+        (b_tag == "negation" and type(b:inner()) == a_tag)
+    then
+        is_subtype = true
+    elseif b_tag == "union" or b_tag == "intersection" then
+        local components = b:components()
+        local conmponents_that_are_subtype = 0
+
+        for _, component in components do
+            if hassubtypeof(component):value() then
+                conmponents_that_are_subtype += 1
+            end
+        end
+
+        return types.singleton(#components == conmponents_that_are_subtype)
+    elseif b_tag == "table" and a_tag == "table" then
+        local b_mt = b:metatable()
+        local a_mt = a:metatable()
+
+        if b_mt == a_mt then
+            return true_type
+        elseif b_mt and a_mt then
+            local b_mts = { b_mt }
+            local next_b_mt
+            local next_a_mt = a_mt
+
+            while next_b_mt do
+                table.insert(b_mts, next_b_mt)
+                next_b_mt = next_b_mt:metatable()
+            end
+
+            while next_a_mt do
+                for _, mt in b_mts do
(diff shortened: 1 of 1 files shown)


aatxe:
Haven't read this yet, but just wanted to say that this is, in general, a reasonable idea and something we'd like to consider, for sure. The devil is in the details, but this was originally part of the plan and got cut to scope down the MVP for user-defined type functions.

gaymeowing (commented):
~~Not too sure if having only in the names for these more explicit checks properly explains it, but theyre intended for having string enums and other enum like things be a simple one liner. Although it falters at only allowing string and a singleton with its subtype being string, so something like the following wouldn't be fixed by this rfc currently I'm realizing in retrospect.~~ 
```luau
if t:is("string") or (t:is("singleton") and type(t:value()) == "string" then
    -- code here
end
```

Edit: this makes no sense and isnt an issue, i think i was dehydrated when i was writing this

docs/user-side-subtype-checking-methods.md
```diff
@@ -0,0 +1,163 @@
...
+
+This RFC proposes adding 4 new methods to the [`type`](./user-defined-type-functions.md#type-instance) userdata in type functions.
+
+### New Methods
+
+| Name | Return Type | Description |
+| ------------- | ------------- | ------------- |
+| `hassubtype(arg: string)` | `boolean` | returns true if self has a subtype of the same tag as the argument. List of available tags: "nil", "unknown", "never", "any", "boolean", "number", "string", "singleton", "negation", "union", "intersection", "table", "function", "class" |
+| `hassubtypeof(arg: type)` | `boolean` | returns true if self has a subtype of the argument. |
+| `musthavesubtype(arg: string)` | `boolean` | returns true if self has only a subtype of the same tag as the argument. List of available tags: "nil", "unknown", "never", "any", "boolean", "number", "string", "singleton", "negation", "union", "intersection", "table", "function", "class" |
```
gaymeowing, line 110: Don't have an idea on what the description for the `musthave` methods should be, so its just the same as the `has` methods.

gaymeowing:
Making this not a draft as its much more well described now, plus to hopefully have someone contribute a description for the `musthave` methods.

gaymeowing:
Would an exact method be a good idea to include aswell? Although It'd be basically rendered useless by generic constraints. But It'd basically not allow unions when the provided subtype isnt a union.
ie
```luau
type meow = {
     hi: string
}

type mrrp = {
     hi: "hello"
}

type meowers = meow | mrrp

print(meowers:exacthassubtypeof(meow)) -- false
print(mrrp:exacthassubtypeof(meow)) -- true
```

vegorov-rbx:
You seem to be too focused on hard-coding some behaviors exactly targeting unions, which leads to the problems of structural subtyping.
Luau actually has moved to use semantic subtyping (more at: https://corp.roblox.com/newsroom/2022/11/semantic-subtyping-luau)

So to me, having anything aside from `hassubtypeof` doesn't make sense (even `hassubtype` - there is no notion of a 'union' or 'intersection' as a standalone type).
I was waiting for the other team members to notice this and didn't comment initially. Would be surprised if they disagree. (they are probably just busy)

gaymeowing:
> You seem to be too focused on hard-coding some behaviors exactly targeting unions, which leads to the problems of structural subtyping. Luau actually has moved to use semantic subtyping (more at: https://corp.roblox.com/newsroom/2022/11/semantic-subtyping-luau)
> 
> So to me, having anything aside from `hassubtypeof` doesn't make sense (even `hassubtype` - there is no notion of a 'union' or 'intersection' as a standalone type). I was waiting for the other team members to notice this and didn't comment initially. Would be surprised if they disagree. (they are probably just busy)

Yeah, the rest would be better solved with generic constraints admittedly.

gaymeowing:
Removed everything else and just left `hassubtypeof`

docs/method-type-hassubtypeof.md
```diff
@@ -0,0 +1,114 @@
...
+end
+```
+
+## Design
+
+The [`type`](./user-defined-type-functions.md#type-instance) will gain a new method: `hassubtypeof`:
+
+```luau
+type:hassubtypeof(arg: type): boolean
+```
```
aatxe, line 106:
```suggestion
The [`type`](./user-defined-type-functions.md#type-instance) will gain a new method: `issubtypeof`:

```luau
type:issubtypeof(arg: type): boolean
```
```

Subtyping is a relation between types, not something that a type "possesses." So, this should be named `issubtypeof` (or something analogous to that).

**Review by aatxe — CHANGES_REQUESTED** [review]

File: docs/method-type-hassubtypeof.md [hunk 3]
```diff
@@ -0,0 +1,114 @@
...
+    return types.singleton(is_subtype)
+end
+```
+
+## Design
+
+The [`type`](./user-defined-type-functions.md#type-instance) will gain a new method: `hassubtypeof`:
+
+```luau
+type:hassubtypeof(arg: type): boolean
```
aatxe, line 105:
```suggestion
type:issubtypeof(arg: type): boolean
```

docs/method-type-issubtypeof.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Design
+
+The [`type`](./user-defined-type-functions.md#type-instance) will gain a new method: `issubtypeof`:
+
+```luau
+type:issubtypeof(arg: type): boolean
+```
+
+This method will return a boolean indicating if the type is a subtype of `arg`.
```
aatxe, line 108: According to what algorithm? Again, we should at least say `according to Luau's subtyping algorithm` so that it's clear that it should e.g. always be consistent with whatever Luau in general is using for subtyping. After all, as you noted above in motivation, it is already possible for you to handwrite your own subtyping algorithm with whatever semantics you want.

aatxe (approved):

PR #103, opened 2025-02-21
Import syntax

jackhexed:
This RFC proposes import syntax like the following:
```luau
import myname, type mytype, myname2, type mytype2 from "mymodule"
```

[Rendered](https://github.com/jackdotink/rfcs/blob/import-syntax/docs/import-syntax.md).

### DIFF
--- a/docs/import-syntax.md
+++ b/docs/import-syntax.md
@@ -63,7 +63,7 @@ This gives users a second way to import modules, which could be confusing to new
 
 The primary alternative is to continue using the existing `require` function.
 
-Some smaller alternatives are:
+Some other alternatives are:
 
 * Change the contextual keyword `from` to the existing reserved keyword `in`. This may reduce parsing complexity, but would be less intuitive.
 * Allow programmers to rebind imported values to other names, perhaps something like `import createElement as e from "@React"`. This would be more flexible, but would also be more complex. This pattern could also not be done with autocomplete, as the autocomplete would not know the new name. It is also a fairly uncommon pattern, existing in usage of only a few libraries. Those users could continue to rebind with a `local` statement.
--- a/docs/import-syntax.md
+++ b/docs/import-syntax.md
@@ -1,0 +1,70 @@
+# Import Syntax
+
+## Summary
+
+This proposal introduces new syntax to import exported values and types from modules. This syntax pairs well with the existing `export type name = type` syntax and the proposed `export name = value` syntax. In a simple example: `import name, name2, type type1, type type2 from "module"`.
+
+## Motivation
+
+The last RFC to propose import syntax was rejected with the primary reason being that destructuring solved all of the same problems, and was a more expansive feature. At this point in time destructuring is pretty much dead in the water, with a destructuring RFC being rejected, and current RFC(s) in limbo.
+
+The proposed import syntax enables programmers to control precisely what they bring into scope from a module. This is particularly useful with larger libraries, like react.
+
+```luau
+-- what was once this:
+local React = require("@React")
+local createElement, useState, useEffect, useContext, useRef = React.createElement, React.useState, React.useEffect, React.useContext, React.useRef
+-- can now be this:
+import createElement, useState, useEffect, useContext, useRef from "@React"
+```
+
+## Design
+
+In EBNF:
+
+```ebnf
+chunk = imports block
+
+imports = {import [';']}
+import = 'import' {importname ','} [importname] 'from' STRING
+importname = ['type'] NAME
+```
+
+Imports would only exist at the top of files, shebang and comments are ignored, and so would be allowed above and inside of imports. 
+
+This import:
+
+```luau
+import createElement, useState, useEffect, useContext, useRef from "@React"
+```
+
+Would be equivalent to:
+
+```luau
+local React = require("@React")
+local createElement, useState, useEffect, useContext, useRef = React.createElement, React.useState, React.useEffect, React.useContext, React.useRef
+```
+
+Types could also be imported, which was previously impossible:
+
+```luau
+import myvalue1, type mytype1, type mytype2 from "MyModule"
+```
+
+At runtime, type imports would be ignored.
+
+The string after the `from` contextual keyword would be a string literal, and would follow the same rules for module resolution as the `require` function.
+
+## Drawbacks
+
+This gives users a second way to import modules, which could be confusing to new users and could lead to inconsistent codebases. This would already be the case with the proposed export-value syntax.
+
+## Alternatives
+
+The primary alternative is to continue using the existing `require` function.
+
+Some smaller alternatives are:
+
+* Change the contextual keyword `from` to the existing reserved keyword `in`. This may reduce parsing complexity, but would be less intuitive.
+* Allow programmers to rebind imported values to other names, perhaps something like `import createElement as e from "@React"`. This would be more flexible, but would also be more complex. This pattern could also not be done with autocomplete, as the autocomplete would not know the new name. It is also a fairly uncommon pattern, existing in usage of only a few libraries. Those users could continue to rebind with a `local` statement.
+* Figure out destructuring syntax, and use that instead. This would be a more expansive feature, and would be more powerful, but would also be more complex. This would also be a more common pattern, and would be more intuitive to users coming from other languages.
--- a/docs/import-syntax.md
+++ b/docs/import-syntax.md
@@ -43,6 +43,7 @@ Would be equivalent to:
 ```luau
 local React = require("@React")
 local createElement, useState, useEffect, useContext, useRef = React.createElement, React.useState, React.useEffect, React.useContext, React.useRef
+React = nil
 ```
 
 Types could also be imported, which was previously impossible:
--- a/docs/import-syntax.md
+++ b/docs/import-syntax.md
@@ -42,17 +42,17 @@ Would be equivalent to:
 
 ```luau
 local React = require("@React")
-local createElement, useState, useEffect, useContext, useRef = React.createElement, React.useState, React.useEffect, React.useContext, React.useRef
+const createElement, useState, useEffect, useContext, useRef = React.createElement, React.useState, React.useEffect, React.useContext, React.useRef
 React = nil
 ```
 
-Types could also be imported, which was previously impossible:
+Imported bindings would be constant. Types could also be imported, which was previously impossible:
 
 ```luau
 import myvalue1, type mytype1, type mytype2 from "MyModule"
 ```
 
-At runtime, type imports would be ignored.
+At runtime, type imports would be ignored. If an import statement is only importing types then at runtime it would not load the specified module.
 
 The string after the `from` contextual keyword would be a string literal, and would follow the same rules for module resolution as the `require` function.
 


docs/import-syntax.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Alternatives
+
+The primary alternative is to continue using the existing `require` function.
+
+Some smaller alternatives are:
+
+* Change the contextual keyword `from` to the existing reserved keyword `in`. This may reduce parsing complexity, but would be less intuitive.
+* Allow programmers to rebind imported values to other names, perhaps something like `import createElement as e from "@React"`. This would be more flexible, but would also be more complex. This pattern could also not be done with autocomplete, as the autocomplete would not know the new name. It is also a fairly uncommon pattern, existing in usage of only a few libraries. Those users could continue to rebind with a `local` statement.
+* Figure out destructuring syntax, and use that instead. This would be a more expansive feature, and would be more powerful, but would also be more complex. This would also be a more common pattern, and would be more intuitive to users coming from other languages.
```
aatxe, line 70: I would not call this a "smaller" alternative. It's quite a bit larger than what you're proposing here.

Almost89:
The main problem I have with this syntax is that the items imported come *before* the path, making autocomplete almost impossible.

A simple solution to this is to switch the imported items and the path around. Obviously, this can't be done with the exact syntax proposed due to it being ambiguous with function calls but a good alternative could be something like the following:
```luau
in "./foo/bar/baz" import foo, type bar, type baz
```

Apart from that, it feels a bit weird that there's no mention of a "default" import anywhere. I understand wanting to match the proposed `export` syntax, but it's strange that the current way everyone is exporting their code (via `return`) is completely ignored. This doesn't matter in most cases (as I assume most usually export a table anyway) but there are definitely valid use cases for exporting a value other than a table.

To solve this problem we could add something like this:
```luau
in "./foo/bar/baz" import default foo, bar, type baz
-- this would be equivalent to:
local foo = require("./foo/bar/baz")
local bar = foo.bar
type baz = foo.baz
```

On this topic, the RFC mentions that:
```luau
-- this import statement:
import createElement, useState, useEffect, useContext, useRef from "@React"
-- is equivalent to this:
local React = require("@React")
local createElement, useState, useEffect, useContext, useRef = React.createElement, React.useState, React.useEffect, React.useContext, React.useRef
```

Is it implying that the module name is implicitly imported or is it simply a typo?

surfbryce:
@Almost89 There are plenty of languages that have auto-complete for the suggested syntax implementation which always ends up being better than the explicit "from this import these".

When using the latter approach you only start getting auto-complete suggestions once the path has been defined for it to draw these suggestions from. In the former approach we can get auto-complete suggestions based on the indices being typed and then can choose the desired path based off that.

The former approach will always be better because of this. It goes from having to remember the path your desired value is in to just simply having to remember the import index for that value (which is typically just the name of the Value itself) and then having intellisense/auto-complete present to you the import options for that value.

Also for the module name being implicitly imported, it most likely wouldn't be and is simply a misleading formatting issue. Realistically it would be equivalent to just the below indexing.

jackdotink:
@Almost89
Imported items come before the path because that is what is more important. Users will generally not write import statements, instead having them autocompleted in by their lsp. Instead, when reading code, especially in an environment without an lsp like github, when someone wishes to find where something comes from they want to scan "what is imported" before "where it's from". Overall this is just a style thing - I don't think it really matters either way.

I did not consider a default import, but I do think it is viable. I think the syntax `import default name from "module"` would work best, with the usage of this syntax disallowing anything else but type imports. Let me know what you think and then I'll put something in the RFC.

Wunder-Wulfe:
My complaint with such a system is it leads to ambiguity in the compiler and may result in old code being broken due to a lack of backwards compatibility.

Consider the following code:

```luau
import "Module Name";

import "Class1", "Class2" from "Module2 Name";

import "ClassA" "ClassB" from "Module3 Name";
```

Here is what this syntax is equivalent to:
```luau
import("Module Name");

import("Class1"), "Class2" from("Module2 Name");

import("ClassA")("ClassB"); from("Module3 Name");
```

The first case evaluates to a function call, which is fine if the implementation of `import` is a function call, similar to `require`.

However, the second case results in a syntax error and would create ambiguity in the code-base.

The third case is syntactically sound, however evaluates as a chain call followed by an additional call to `from`, which may not be what you intend.

Even if you reserve `import` and `from` as keywords, you will break existing code:
```luau
local function Vector4.Lerp(
    from: Vector4, 
    to: Vector4, 
    alpha: Vector4 | number
): Vector4
    -- function body
end

-- ...

local function Mesh.import(binary: buffer): Mesh
    -- function body
end
```

My suggestion would be to implement imports as comments, though that would clog comments and result in some confusion

```luau
--!import "Module Name"
--!import "Class1", "Class2" from "Module2 Name"
```

jackdotink:
@Wunder-Wulfe 
I specifically designed the syntax to not introduce any ambiguities. `import STRING` is not a valid import, and would be parsed as a function call statement. `import NAME` is the only valid start to an import.

dphfox:
I previously had a similar idea a while ago in the old repo; at the time I thought it would be a reasonable alternative to more general destructuring syntax like you do here. As I understand it, this syntax:

```
import useState, useMemo from "react"
```

Would basically mean you could never have anything that could be interpreted as an expression in that list of identifiers, which is unpalatable (but nonetheless, theoretically workable...)

However, over time I've thought about it a lot more and it really seems like whatever syntax we would choose for an import syntax should align with destructuring syntax anyway, in case the call is made to allow destructuring for local variable declarations too, thus not really sidestepping the problem at all.

This would rule out the use of a contextual keyword at the start of the statement.

```
local { .useState, .useMemo } from "react"
```

At that point, you may as well just use destructuring with the normal import system, unless you're particularly motivated to have a nice syntax for this.

```
local { .useState, .useMemo } = require("react")
```

So I don't think this is the right path to go down.

jackdotink:
I somewhat agree, though I am unsure if destructuring syntax will be added. I also feel that this syntax more closely aligns with Luau's export syntax.

Consider this an alternative for when destructuring fails to be accepted once again.

aatxe:
Worth noting the similarity with #97 too, fwiw.

gaymeowing:
Would be nice if this rfc allowed for 
```luau
import all from ./react
```
And this for all types
```luau
import types from ./react
```

Also perhaps worth considering having file paths as first class?

loukamb:
Could this be turned into the killer destructuring syntax we're all looking for, actually? The `from` keyword specified in the RFC would introduce a new keyword for a _single_ purpose, and it would be a shame if the usefulness of its introduction wouldn't be extended to, for example,

```lua
local t = { a = 1, b = 2 }
local function f()
    local a, b from t
    print(a, b)
end
```

Then you can have the module-like import syntax like this:
```lua
-- Import specific fields.
local useState, useEffect from require("react")

-- Import everything.
local react = require("react")
```

jackdotink:
@loukamb 
Anything that affects the global environment is a non-starter imo. This RFC doesn't really add a keyword in the traditional sense that it reserves it - it is fully contextual, only a bit of nice syntax. Those other examples could be added, but that should be done in a separate RFC.

loukamb:
Syntactical minimalism aside, I don't see an issue with adding a contextual keyword, but if one is to be added, it would be nice to give it additional functionality. A separate RFC can written but it would make this RFC redundant, code processors/bundlers notwithstanding. I do understand that a very predictable import statement form eases parsing module imports, and my suggestion would make that parsing a tiny bit less predictable.

The global import assignment is for parity with standard assignments. `local` can be learned and approached as a scope-modifying keyword that can be added to or removed from a variable assignment. My suggestion assigns variables, and therefore `local` should be addible and removable. The mutability of the global environment in Lua(u) is an awkward feature that very occasionally deserves to be remembered, the poor thing!

jackdotink:
It should not be remembered except by embedders.

cheesycod:
Ok, so I have a few implementation related questions just so I understand how this works:

1. Are the imports local (does the imported thing act like a local declaration or is it global)
2. Would the import rfc here conflict with luaus support for environment tables where each chunk can get its own environment
3. Would it be beneficial to also have a new export function/export table syntax be added like JS has?
4. Finally, do the imported values themselves reflect/get stored in the environment table (which is basically an extension of 1, but I was just curious)

bradsharp:
It looks like this syntax would not support legacy return types which will result in code using `import` and `require` at the same time. Likewise, there's no way to access the table returned by the module, which is useful when you know you want the entire module instead of just parts of it.

gaymeowing:
> Those are good suggestions. What do you think about `import * as name from "path"`? This allows you to import the entire exports table (or whatever value is returned, though the syntax is wonky for that).

Why not have an `all` 'keyword'? As using an asterisk feels sorta out of place given how keyword heavy Lua and Luau are.

JoshBashed:
I don't really see why we can't match javascript and just require destructuring in the first place.

```javascript
import something from "package"
import { item, item2 } from "package"
```

jackdotink:
@gaymeowing While I agree that a keyword would be better than a symbol, I don't see a way to integrate an `all` keyword into my proposed syntax without prohibiting using `all` as a name.

@JoshBashed As I've written before, this RFC is written under the assumption that destructuring syntax will never be added to Luau. Several destructuring syntax RFCs have been opened, and all have been closed for a wide variety of reasons that the Luau team seems to flip-flop on. I do not think destructuring will ever be added to Luau.

loukamb:
Not aiming for this to be a self-plug in any way, but I've solved this problem _and_ the problem of table destructuring at once by using a [`from` operator](https://lus.dev/manual/acquis-4/) in a Lua-like language. It's a revision of my previous idea.

```lua
local a, b, c from table_expr
```
is equivalent to:
```lua
local a, b, c = table_expr.a, table_expr.b, table_expr.c
```

Destructure directly from function calls:

```lua
local function getConfig()
    return { host = "localhost", port = 8080 }
end

local host, port from getConfig()
assert(host == "localhost" and port == 8080)
```

And, relevant to this RFC specifically, it would fit our need for an import-like syntax as well while retaining `require` semantics:

```lua
-- The repetitive way
local module = require("mymodule")
local func1 = module.func1
local func2 = module.func2
local func3 = module.func3
```

...turns into...

```lua
-- Definitely less repetitive I'd say
local func1, func2, func3 from require("mymodule")
```

Use it with types:
```lua
type str, num, other from require("mymodule")
```

Other examples of how it can improve overall readability and language ease of use:

```lua
local function configure(options)
    local host, port, timeout from options
    -- use host, port, timeout
end

local outer = { inner = { value = 42 } }
local value from outer.inner
assert(value == 42)
```

For picking from the default export while preserving a reference to the default export itself, you can just do this:
```lua
local t = require("mymodule")
local a, b, c from t
type d, e, f from t
```

This could be submitted as another RFC, but I'd suggest encompassing import functionality with `from`. I think the preservation of the `require` semantic is good.

deviaze:
> Not aiming for this to be a self-plug in any way, but I've solved this problem _and_ the problem of table destructuring at once by using a [`from` operator](https://lus.dev/manual/acquis-4/) in a Lua-like language. It's a revision of my previous idea.

Not backwards compatible because `from <expr>` can already be a functioncall (`from` isn't a keyword) if the RHS is a table or string literal.

PR #104, opened 2025-02-23
RFC: math.fma

Wunder-Wulfe:
[Rendered](https://github.com/Wunder-Wulfe/rfcs/blob/master/docs/function-math-fma.md)

The proposal in this RFC is to implement `math.fma`, as well as potentially its sister functions such as `math.fms` and `math.fma231`, in accordance with IEEE standard: [IEEE 754-2008](https://ieeexplore.ieee.org/document/4610935) *[(extension)](https://ieeexplore.ieee.org/document/5487502)*

Optimized code would appear as follows:

*before*
```luau
--- Calculate the dot product between two 4-dimensional vectors <code>a</code> and <code>b</code>
function Vector4.Dot( Vector4 a, Vector4 b ): number
    return a.X * b.X + a.Y * b.Y + a.Z * b.Z + a.W * b.W;
end
```
generated native arithmetic:
```asm
MULSD xmm0, xmm4             ; ax * bx
MULSD xmm1, xmm5             ; ay * by
MULSD xmm2, xmm6             ; az * bz
MULSD xmm3, xmm7             ; aw * bw

ADDSD xmm0, xmm1             ; (ax * bx) + (ay * by)
ADDSD xmm0, xmm2             ; + (az * bz)
ADDSD xmm0, xmm3             ; + (aw * bw)
```

*after*
```luau
--- Calculate the dot product between two 4-dimensional vectors <code>a</code> and <code>b</code>
function Vector4.Dot( Vector4 a, Vector4 b ): number
    return math.fma( a.X, b.X, math.fma( a.Y, b.Y, math.fma( a.Z, b.Z, a.W * b.W ) ) );
end
```
generated native arithmetic:
```asm
MULSD       xmm3, xmm7       ; aw * bw
VFMADD213SD xmm2, xmm6, xmm3 ; az * bz + (aw * bw)
VFMADD213SD xmm1, xmm5, xmm2 ; ay * by + (az * bz + aw * bw)
VFMADD213SD xmm0, xmm4, xmm1 ; ax * bx + (ay * by + az * bz + aw * bw)
```

Additional example:

*before*
```luau
--- Calculate the magnitude of a 4-dimensional vector
function Vector4.Magnitude( Vector4 vec ): number
    return math.sqrt( 
        vec.X * vec.X + vec.Y * vec.Y + vec.Z * vec.Z + vec.W * vec.W
    );
end
```
generated native arithmetic:
```asm
MULSD xmm0, xmm0             ; ax * ax
MULSD xmm1, xmm1             ; ay * ay
MULSD xmm2, xmm2             ; az * az
MULSD xmm3, xmm3             ; aw * aw

ADDSD xmm0, xmm1             ; (ax * ax) + (ay * ay)
ADDSD xmm0, xmm2             ; + (az * az)
ADDSD xmm0, xmm3             ; + (aw * aw)

; call sqrt()
```

*after*
```luau
--- Calculate the magnitude of a 4-dimensional vector
function Vector4.Magnitude( Vector4 vec ): number
    return math.sqrt( 
        math.fma( vec.X, vec.X, math.fma( vec.Y, vec.Y, math.fma( vec.Z, vec.Z, vec.W * vec.W ) ) ) 
    );
end
```
generated native arithmetic:
```asm
MULSD       xmm3, xmm3       ; aw * aw
VFMADD213SD xmm2, xmm2, xmm3 ; az * az + (aw * aw)
VFMADD213SD xmm1, xmm1, xmm2 ; ay * ay + (az * az + aw * aw)
VFMADD213SD xmm0, xmm0, xmm1 ; ax * ax + (ay * ay + az * az + aw * aw)

; call sqrt()
```

### DIFF
--- a/docs/function-math-fma.md
+++ b/docs/function-math-fma.md
@@ -1,0 +1,120 @@
+# math.fma
+
+## Summary
+
+Add `fma` as a function to the math library, computing the [Fused multiply–add](https://en.wikipedia.org/wiki/Multiply%E2%80%93accumulate_operation#Fused_multiply%E2%80%93add) operation, following the appropriate IEEE standard: [IEEE 754-2008](https://ieeexplore.ieee.org/document/4610935) *[(extension)](https://ieeexplore.ieee.org/document/5487502)*.
+
+## Motivation
+
+Fused multiply-add, also known as multiply-accumulate and often abbreviated as `fma` or `mad`, is a computation which condenses the operations `(a × b) + c`, which normally would require a `MULSD` and `ADDSD` instruction, into a single processor instruction.
+
+This operation is commonly used in calculations such as dot products, cross products, quaternion rotations, and matrix multiplications.
+
+The *two* advantages of this operation are as follows:
+1. Floating point rounding only occurs at the end of the instruction, allowing for enhanced precision over computing two separate instructions.
+2. Instruction count is reduced for math-intensive code, resulting in smaller code size and better performance.
+
+   The example below is a dot product between two 4-dimensional vectors:
+
+   ```cpp
+   double dot(
+     double ax, double ay, double az, double aw,
+     double bx, double by, double bz, double bw
+   )
+   {
+     return
+         (ax * bx)
+       + (ay * by)
+       + (az * bz)
+       + (aw * bw);
+   }
+   ```
+
+   This computation results in a total of `7` math instructions
+     ```asm
+     MULSD xmm0, xmm4             ; ax * bx
+     MULSD xmm1, xmm5             ; ay * by
+     MULSD xmm2, xmm6             ; az * bz
+     MULSD xmm3, xmm7             ; aw * bw
+  
+     ADDSD xmm0, xmm1             ; (ax * bx) + (ay * by)
+     ADDSD xmm0, xmm2             ; + (az * bz)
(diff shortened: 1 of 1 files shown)


vegorov-rbx:
Thank you for the interest in extending Luau!

Unfortunately, after talking with the team, we've agreed that this feature does not meet the bar for inclusion in Luau. This is for a few reasons, namely that: (1) it has fairly limited use cases, (2) poses portability issues with different architectures, and (3) would require more significant motivation.

In detail, some of the problems we're seeing here include:
* Existence of this function has no relation to the ability of Luau VM (vector library) and other embedders (Roblox CFrame) to use the fma instruction. It is actually already used by various VM and external methods.
* It seems to only focus on the native codegen side of things because VM implementation will still have dispatch and function call overhead. Philosophically, we like the idea of optimized math instructions, but we do not want the developer experience to be mired in constant micro-optimizing. We'd prefer to add better compiler support for optimizing math than to require every developer to optimize it themselves by hand.
* `vector.dot` is already implemented as a 'vdpps' on x64 so that's not a very motivating example.
* It is not correct to fallback in the described way as it will not fulfil the expectations of the function.
* If the operation is desirable, we will perform it automatically in native code generation. VM already operates in environments performing FP contractions allowed by C/C++. Luau has never provided numerical guarantees more precise than that.

PR #105, opened 2025-02-26
Correct `:readparent()` and `:writeparent()` for `:parent()` within the user defined type functions RFC

alicesaidhi:
The RFC mentions the existence of `:readparent()` and `:writeparent()`, yet the runtime only has `:parent()`. This updates the RFC so that it will properly reflect the deviation within the runtime.


aatxe:
Yeah, agreed with @vegorov-rbx. We should be fixing the doc site and implementation here, not changing the designed API.

PR #106, opened 2025-03-07
Add eager inference annotation for polymorphic types

Ukendio:
[rendered link](https://github.com/Ukendio/rfcs/blob/add-eager-inference-annotation-for-polymorphic-types/docs/eager-inference-annotations-for-polymorphic-types.md)


docs/eager-inference-annotations-for-polymorphic-types.md
```diff
@@ -0,0 +1,61 @@
...
+
+```luau
+function test<T>(first: T, second: T, third: T!): T
+    return first
+end
+
+test(1, "string", true) -- TypeError: Type `boolean` could not be converted into `number | string`
+```
+
+Notably, this behavior would be identical to other languages, such as typescript's `noinfer<T>`. This has the added drawback that the `!` syntax modifier would need to be barred from return types, as the return type holds no relevance to implicit instantiation.
```
Fireboltofdeath, line 57: I feel like a `noinfer<T>` type function should be listed more clearly as an alternative to the `!` syntax, since it doesn't introduce any new keywords or symbols

hardlyardi: I agree, I'll add an update shortly to add noinfer<T> clearly as an individual alternative.

docs/eager-inference-annotations-for-polymorphic-types.md
```diff
@@ -0,0 +1,61 @@
+# Eager Inference Annotations for Polymorphic Types
+
+## Summary  
+
+The RFC introduces a feature to annotate polymorphic function types to express that the first instantiation of a polymorphic type `T` is the one that sticks. 
+
+## Motivation  
+
+The purpose of this feature is to develop syntax to prevent polymorphic types from widening into (e.g., number | string) when a function is implicitly instantiated with different argument types. E.g., `test(1, "a")`. In the following code, Luau's current solver infers `T` to be of a union type:
```
aatxe, line 9: What is "Luau's current solver?"

We have two products right now, Luau's Type Solver and Luau's New Type Solver. "type solver" is really a brand name coined by @andyfriesen, not something external folks necessarily recognize. Could also reasonably describe it as Luau's V2 type inference engine or something of that sort, honestly not entirely sure how we want to deal with the language around the two type inference systems in RFCs.

docs/eager-inference-annotations-for-polymorphic-types.md
```diff
@@ -0,0 +1,61 @@
...
+
+```luau
+function test<T>(a: T, b: T): T
+    return a
+end
+
+local result = test(1, "string") -- inferred type `T`: number | string"
+```
+
+This behaviour can be useful in some cases but is undesirable when a polymorphic function is intended to constrain the input types to be consistent.
```
aatxe, line 19: > when a polymorphic function is intended to constrain the input types to be consistent.

This doesn't mean anything to me. The input type that was inferred in the example is already consistent. Every single instance of the generic `T` is indeed instantiated with the same consistent type, always, regardless of this RFC. That's just the semantics of polymorphism. I think the way you probably want to talk about your desire is something like... wanting to retain the exact type of the argument.

docs/eager-inference-annotations-for-polymorphic-types.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Design  
+
+We propose adding some symbol as a suffix (or prefix) that annotates the "eager" inference behaviour for a polymorphic type.
+Subsequent usages of type `T` where `T` is "eager" would be ignored during instantiation.
```
aatxe, line 24: This description is misleading. They're not ignored during instantiation, you'll still replace each occurrence with the inferred type. The only thing you're proposing changes here, afaik, is that an annotated generic will opt call sites into inferring the exact type of the first argument for automatic instantiation.

docs/eager-inference-annotations-for-polymorphic-types.md
```diff
@@ -0,0 +1,61 @@
...
+function test<T!>(a: T, b: T): T
+    return a
+end
+
+test(1, "string") -- TypeError: Expected `number`, got `string`
+```
+
+## Drawbacks  
+
+- Introduces a new syntax modifier `!`, which may lead to a symbol soup. However, `T!` doesn't seem too shabby next to `T?`.
```
aatxe, line 40: > `T!` doesn't seem too shabby next to `T?`

I think this is actually a reasonably strong argument against this syntax: people might confuse it as having something to do with optionals or `nil`, and it does not at all.

docs/eager-inference-annotations-for-polymorphic-types.md
```diff
@@ -0,0 +1,61 @@
...
+end
+
+test(1, "string", true) -- TypeError: Type `boolean` could not be converted into `number | string`
+```
+
+Notably, this behavior would be identical to other languages, such as typescript's `noinfer<T>`. This has the added drawback that the `!` syntax modifier would need to be barred from return types, as the return type holds no relevance to implicit instantiation.
+
+
+### Keywords
+Something like `<greedy T>` or `<strict T>` should also be considered if we want to reduce symbols. This idea has merit when considering the potential complexity of type aliases combined with `T!?`
```
aatxe, line 71: > `T!?`

This is also a very strong argument against this syntax.

aatxe: Though, is this actually possible under your proposal? You made `!` a property of the definition at the generic, and you can't write optional in that context.

docs/eager-inference-annotations-for-polymorphic-types.md
```diff
@@ -0,0 +1,61 @@
+# Eager Inference Annotations for Polymorphic Types
+
+## Summary  
+
+The RFC introduces a feature to annotate polymorphic function types to express that the first instantiation of a polymorphic type `T` is the one that sticks.
```
aatxe, line 5: This description is misleading. Instantiation is the process of replacing the generics at a call site with a specific type, so the first instantiation, to me, would mean "the first call to this polymorphic function" which is not the behavior you're proposing at all (fortunately, because that would be awful).

gaymeowing:
I think this is something better solved with type functions, because this seems really niche and isn't very explicit with its syntax.

docs/eager-inference-annotations-for-polymorphic-types.md
hardlyardi: Oops. I'll change it to "constrain the subsequent input types to be identical to the first usage." or something, couldn't think of anything better at the time.

docs/eager-inference-annotations-for-polymorphic-types.md
hardlyardi: You're absolutely right, "ignored" is the wrong term. I'll update it to "Subsequent uses of a polymorphic type `T` where `T` is "eager" will be inferred as the precise type of the first occurrence."

docs/eager-inference-annotations-for-polymorphic-types.md
hardlyardi: cc @Ukendio

docs/eager-inference-annotations-for-polymorphic-types.md
hardlyardi: You're right, `T!?` wouldn't be possible under the main proposal. It is however possible under the per-usage-binding alternative with unique syntax, I'll move it to a drawback of that alternative where it should be.

docs/eager-inference-annotations-for-polymorphic-types.md
hardlyardi: cc @Ukendio

docs/eager-inference-annotations-for-polymorphic-types.md
Ukendio: Yeah might lean onto using keywords or putting the annotation as a prefix instead `!T`.

docs/eager-inference-annotations-for-polymorphic-types.md
```diff
@@ -0,0 +1,72 @@
+# Eager Inference Annotations for Polymorphic Types
+
+## Summary  
+
+The RFC introduces a feature to annotate polymorphic function types to express that the first bind to `T` will be the one that sticks.
```
Fireboltofdeath, line 5: What would be the behavior when a type references `T` more than once, such as an object?

```luau
function F<T!>(value: { a: T, b: T })
```

Additionally, what about when the object type is aliased (i.e there's no specific "first occurence")

```luau
type MyObject<T> = { a: T, b: T };

function F<T!>(value: MyObject<T>)
```

docs/eager-inference-annotations-for-polymorphic-types.md
Ukendio: > You're right, `T!?` wouldn't be possible under the main proposal. It is however possible under the per-usage-binding alternative with unique syntax, I'll move it to a drawback of that alternative where it should be.
The problem is with inference not syntax as the annotation is defined at the definition, not the argument so there is no conflict in syntax. The problem would be that if a binding is required to be inferred by the first argument that binds to `T` annotated that it is an optional then the inference simply wouldn't work unless we changed the behaviour to infer the first non optional argument that binds to T to dictate the type.

I am curious to know how it worked before with eager inference by default, as that is likely what we want to mirror.

docs/eager-inference-annotations-for-polymorphic-types.md
Ukendio: > What would be the behavior when a type references `T` more than once, such as an object?

It would behave in the same way as the old type solver which is taking the first definition that binds to `T`. 

> Additionally, what about when the object type is aliased (i.e there's no specific "first occurence")

Similar answer to the last point which is mirroring the old behaviour but in this case would be to go in the order of the fields. I don't know the exact heuristics for that is but I feel most of this could be answered with just saying it should do whatever the type solver did. 

I will note that I do think something like `noinfer<T>` is a pretty good solution in these cases but I haven't personally prepared to make a case for that proposal.

alexmccord:
This really looks like `<a, b | a ~ b>(a, b) -> ()` which has an equality constraint, equivalently expressed as `<a: b, b: a>(a, b) -> ()`. The problem with this syntactic sugar is that it conflates the two type variables with a single generic, which I don't think is accurate within Luau's set theoretic type system.

alexmccord:
Also, suppose I write this:

```luau
function f(): string | number
  return if math.random() > 0.5 then "hello" else 7
end

the_rfc_fn(f(), 5) -- is this ok?
the_rfc_fn(5, f()) -- what about this one?
```

This means the type system will be too finicky if this was defined as "greedy" rather than finding some supremum and then verifying that all other arguments are indeed a subtype of some supremum. _That_ would relax the constraint of being able to apply the function.

alexmccord:
There's also this one, which I think requires higher order unification which is undecidable:

```luau
function mk_foo(): { tag: "foo" }
  return { tag = "foo" }
end

function mk_bar(): { tag: "bar" }
  return { tag = "bar" }
end

the_rfc_fn(mk_foo(), mk_bar()) -- is this ok?
```

For this to typecheck, you need to introduce a fresh type variable _as you intersect_ all types that flows in, arriving at the type `{ tag: "foo" | "bar" <: 'a <: "foo" & "bar" }`. The reason for this is because the caller has a typing context stronger than what the constraints of the function would allow! But now here's the higher order unification: what if the return type of the application was bounded by `{ tag: "foo" } | { tag: "bar" }`? Resolving this conflict depends on the constraints in the _return type_ of the caller and now we're in square one. This means there exists some type that makes the function application legal and the type that the function accepts actually have no subtyping relations between each other, which turns this into a search problem.

And finally, this one:

```luau
function f(x: string, y: string)
  if x == "x" and y == "y" then
    the_rfc_fn(x, y) -- is this ok?
  end
end
```

This is already really hard! There exists a supremum, but it isn't done by intersecting, which directly conflicts with the above typechecking logic. Instead we now have to decide if this is valid by going directly to their supertypes: `string`. This then draws a new problem: where do you stop in a way that preserves the idea of this RFC? Because you can also say the supertype of `string` and `number` is `unknown` and that would still cause the function application to pass typechecking.

If your reaction is that the last one above is illegal, you could just do a simple coordinate transformation such that the refinements forces them to agree on their primitive types.
```luau
function f(x: string | number, y: string | number)
  if typeof(x) == "string" and typeof(y) == "string" then
    the_rfc_fn(x, y)
  else
    the_rfc_fn(x, y)
  end
end
```

The point I'm making here is this: the RFC proposing a greedy instantiation would very quickly run into problems with type inference, and any attempt to relax it a little is self-defeating and renders the feature useless. Welcome to undecidability hell.

alicesaidhi:
This might be more interesting to implement with generic constraints? Following existing generic constraint syntax proposals, we can write something like `function foo<T, U where T>(a: T, b: U)` so that U is constrained to whatever type T is.

```luau
foo(arg1", 2) -- Type 'number' could not be converted into 'string'
```

Sure, this is more explicit, but I feel like it makes much more sense and would be more realistic than greedy instantiation. This should also satisfy the use-cases for many libraries that provide their own setter functions or something similar.

PR #107, opened 2025-03-08
Frozen table syntax

dphfox:
Add syntax for freezing a table upon construction to:
- Improve ergonomics of immutable code
- Reduce friction incurred by fully qualified `table.freeze`
- Trivially detect optimisation opportunities for deeply frozen tables at construction time
- Leverage the technical benefits of immutability in more places via easy adoption

[Rendered](https://github.com/dphfox/rfcs/blob/freeze-syntax/docs/syntax-frozen-tables.md)

### DIFF
--- a/docs/syntax-frozen-tables.md
+++ b/docs/syntax-frozen-tables.md
@@ -79,8 +79,9 @@ TableUtils.join(
 )
 ```
 
-This leaves the feature in an awkward place, where it provides great theoretical benefits, but is impractical to express without using undesirable language features.
-Ideally, frozen tables should have only marginal cost to adopt, to reduce friction and encourage wider usage. This inexpressibility is a serious cost.
+This leaves the feature in an awkward place, where it provides great theoretical benefits, but is much more verbose to express.
+
+Ideally, frozen tables should have only marginal cost to adopt, to reduce friction and encourage wider usage.
 
 ## Design
 
--- a/docs/syntax-frozen-tables.md
+++ b/docs/syntax-frozen-tables.md
@@ -79,7 +79,7 @@ TableUtils.join(
 )
 ```
 
-This leaves the feature in an awkward place, where it provides great theoretical benefits, but is much more verbose to express.
+This leaves the feature in an awkward place, where it provides great theoretical benefits, especially as a "sensible default", but is much more verbose to express and so can't properly serve that purpose.
 
 Ideally, frozen tables should have only marginal cost to adopt, to reduce friction and encourage wider usage.
 
--- a/docs/syntax-frozen-tables.md
+++ b/docs/syntax-frozen-tables.md
@@ -2,7 +2,11 @@
 
 ## Summary
 
-Add a syntax to table literals to allow them to be frozen from construction.
+Add syntax for freezing a table upon construction to:
+- Improve ergonomics of immutable code
+- Reduce friction incurred by fully qualified `table.freeze`
+- Trivially detect optimisation opportunities for deeply frozen tables at construction time
+- Leverage the technical benefits of immutability in more places via easy adoption
 
 ```Lua
 local struct = {:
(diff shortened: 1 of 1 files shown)


dphfox (commented):
Update summary to match PR

bradsharp:
While the syntactic sugar for calling a function with a table may not be favored by all users it does still exist so I don't feel like we should be trying to replace it with more sugar without a greater motivation for doing so

I personally have never seen an issue with this style and I think it reads more nicely than a symbolic representation that does the same thing.

dphfox:
I would beg to differ as someone who almost exclusively uses immutable tables these days. I find myself really wishing for a built-in syntax for this.

This snippet from the RFC is based on actual code I've been writing recently:

```Lua
TableUtils.join(
    table.freeze({
        within = container,
        alignment = table.freeze({ x = 0.5, y = 0.5 })
    }),
    table.freeze({
        group_bounds = table.freeze({ min = vector.create(0, 0, 0), max = vector.create(1, 1, 1) })
    })
)
```

When compared to the "easier" mutable version:

```Lua
TableUtils.join(
    {
        within = container,
        alignment = { x = 0.5, y = 0.5 }
    },
    {
        group_bounds = { min = vector.create(0, 0, 0), max = vector.create(1, 1, 1) }
    }
)
```

The amount of added noise is significant and inhibits my review of the code.

Furthermore, as the maintainer of Fusion, I've been in favour of the use of the function call syntax sugar for years, even going as far as to have promoted it in official documentation. However, from my lived experience, very few people actually understand what it's doing at all. People treat this syntax like some kind of magical code preprocessing or macro system, and I've been mired in many long Discord threads having to explain it or defend it. 

And that's just the UX side of things - that's before you get into how it complicates Fusion's relationship with tools like StyLua by requiring extra configuration and potentially letting through some things we'd rather not enable as a second order effect of ignoring that formatting.

It's gotten to the point that the Fusion project is probably going to drop the use of the syntax just to keep it simple. So suffice it to say I may not be enthusiastic about it.

Dekkonot:
I will be honest, I don't see the motivation as being very strong. You've listed your own project as the motivation, which is fair because it's a large project, but I dont really know of very many places that encourage it's use outside of Fusion. I'd be welcome to being corrected on this front but without more motivation this reads as "I use this feature a lot so I want sugar for it" to me.

deviaze:
I'm not fully convinced of the usecase of dedicated frozen table syntax. If I needed a bunch of frozen tables right now, I'd probably do something like 

```luau
local freeze = table.freeze -- or something even shorter when idc about readability
TableUtils.join(
    freeze {
        within = container,
        alignment = { x = x, y = y },
    },
    ...
)
```

I'm honestly perfectly fine with table call syntax. It's pretty useful in increasing readability and allowing for expressive DSL-like syntax in some circumstances; I'm not sure it needs to be maligned or obviously discouraged in the RFC. 

If people haven't seen table (or string literal) calls before, it's more of a language onboarding issue than a "stop using bad Lua feature in Luau" imho.

From a visual syntactic standpoint, I prefer `{* *}` over `{: :}` as not only is it more visually distinct, the star icon `*` kinda looks like a snowflake & therefore better implies `table.freeze`. I also disagree that `{:` is more ergonomic than `{*`; the key combination `{` + `:` forms a more complicated one-hand bigram on US keyboards (harder/more annoying to type) than `{` + `*`, which you'd type with separate hands.

Kampfkarren:
TBH I don't even think table call syntax is relevant to this at all given that it reduces just two characters while this reduces far more. I just don't think these sigils are the type of thing I want Luau to be full of

dphfox:
Fair enough. I can accept that my reasoning here is niche

andyfriesen:
For what it's worth, we're either building or planning to build a bunch of smaller, independent features that should add up to most of what you want.

First, the new solver already supports the concept of a read-only table property.

Secondly, we eventually want to introduce separate table types for "precise" and "imprecise" tables.  tldr: If the type of a particular table is precise, that means that we statically know every single property in the table.  It has nothing else.  By contrast, an inexact table type indicates that we statically know some but not all of the properties of the table.

A precise table whose properties are all read-only should be very close to what you're asking for here.

PR #108, opened 2025-03-08
Precision parameter in math.round

T045TN1NJ4:
Proposes introducing a second parameter to `math.round` to allow users to specify a number of decimals round numbers to.
```luau
print( math.round(3.1415, 2) ) -- Output: 3.14
```
In doing so, we:
- Improve user convenience by providing an implementation of the behavior for them
- Provide a marginally more performant alternative to existing implementations

[Rendered](https://github.com/T045TN1NJ4/luau-rfcs/blob/master/docs/precision-parameter-in-math-round.md)


deviaze:
A very long time ago, I tried passing a second argument into `math.random` hoping it'd round to that decimal place. It didn't work and I was sad. That's why I support this RFC.

PR #109, opened 2025-03-24
Adjust the meaning of relative `require()` calls within `init.luau`

andyfriesen:
[Rendered](https://github.com/luau-lang/rfcs/blob/adjust-relative-init-requires/docs/abstract-module-paths-and-init-dot-luau.md)

### DIFF
--- a/docs/abstract-module-paths-and-init-dot-luau.md
+++ b/docs/abstract-module-paths-and-init-dot-luau.md
@@ -1,4 +1,4 @@
-# Adjust the meaning of `init.luau`
+# Abstract module paths and `init.luau`
 
 ## Summary
 
@@ -65,9 +65,27 @@ project, we would like to rectify this.
 
 ## Design
 
-Luau will no longer consider relative requires from a package `init.luau` file
-to resolve relative to the script itself.  It will instead resolve relative to the
-script's parent, i.e. the folder containing the module.
+First, we introduce a new abstraction: a _module path_.  Module paths refer
+either to _modules_ or _directories_.
+
+For the purposes of navigation, both modules and directories are functionally
+identical: modules and directories can both have children, which could
+themselves be modules or directories, and both types can have at most one
+parent, which could also be either a module or a directory.
+
+The thing that separates a module from a directory is precisely that modules
+represent source code that can be imported via the `require()` function.
+Directories, by contrast, are merely organizational units.
+
+The central feature of this RFC is about how module paths correspond to
+filesystem paths: A module path refers to a module if it corresponds either to a
+`.luau` file or to a directory that contains a file named `init.luau`.  A module
+path refers to a directory if it refers to a filesystem directory that lacks an
+`init.luau` file.
+
+More concretely, Luau will no longer consider relative requires from a package
+`init.luau` file to resolve relative to the script itself.  It will instead
+resolve relative to the script's parent, i.e. the folder containing the module.
 
 Secondly, we recognize an unfortunate side effect of this change: code within
 `package/init.luau` is forced to write `require('./package/dependency')` when it
@@ -75,7 +93,7 @@ specifically wants to carry out the ordinary task of importing a subordinate
 module.
 
 We propose to alleviate this with a special import alias `@self` that resolves
-to the current module.
+to the path to the current module.
 
 ```lua
 -- package/init.luau
--- a/docs/abstract-module-paths-and-init-dot-luau.md
+++ b/docs/abstract-module-paths-and-init-dot-luau.md
@@ -85,7 +85,8 @@ path refers to a directory if it refers to a filesystem directory that lacks an
 
 More concretely, Luau will no longer consider relative requires from a package
 `init.luau` file to resolve relative to the script itself.  It will instead
-resolve relative to the script's parent, i.e. the folder containing the module.
+resolve relative to the location of the abstract module it represents, i.e. the
+location of its parent folder in a filesystem context.
 
 Secondly, we recognize an unfortunate side effect of this change: code within
 `package/init.luau` is forced to write `require('./package/dependency')` when it
--- a/docs/abstract-module-paths-and-init-dot-luau.md
+++ b/docs/abstract-module-paths-and-init-dot-luau.md
@@ -110,7 +110,7 @@ name of the current source file in order to know what `require('./foo')` will
 do, and if a developer renames a file to or from `init.luau`, they will be
 forced to rewrite every `require()`.
 
-We can mitigate this with a new lint: If a module is the parent of other
+We could someday mitigate this with a new lint: If a module is the parent of other
 modules, it is poor style for it to directly import sibling or parent modules.
 Files named `init.luau` should never issue a require that resolves to a sibling
 module.  In our example, we would warn if `package/init.luau` were to
--- a/docs/adjust-meaning-of-init-dot-luau.md
+++ b/docs/adjust-meaning-of-init-dot-luau.md
@@ -1,0 +1,123 @@
+# Adjust the meaning of `init.luau`
+
+## Summary
+
+We will redefine `foo/init.luau` to be the contents of the module named foo for
+the purpose of resolving relative `require()` calls within its body.
+
+We will do this because a significant portion of our users today rely on the
+idiom that `init.luau` represents the module's contents, especially when using
+Rojo to build for Roblox.
+
+## Motivation
+
+First, consider an example filesystem:
+
+```
+./foo.luau
+./package
+./package/init.luau
+./package/foo.luau
+./package/dependency.luau
+```
+
+Luau maps this filesystem into a module tree.  Module paths largely correspond
+to filesystem paths, but they are distinct ideas.  Given the filesystem above,
+Luau considers the following module paths to be valid:
+
+```
+foo
+package
+package/init
+package/foo
+package/dependency
+```
+
+When Luau is running on a conventional filesystem, we presently offer some
+special behaviour to afford the ability for modules to both contain other
+modules, and for them to themselves export values and types: The module path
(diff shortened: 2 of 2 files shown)


vegorov-rbx:
> This feels complicated to implement from a runtime standpoint (especially as Luau offers no abstraction that also virtualizes the filesystem bits out)

We are working to deliver a C module to initialize `require` function with support for custom VFS navigation.

Dekkonot:
Wrote more than I expected to, and most of it is getting stuck on semantics and how the problem is presented. Just want to get out of the way that I support this solution. I'm not thrilled with it because it'll require me to adjust my mental model for how require-by-string works but I think this is finally a real solution for cross-runtime compatibility.

Just to clarify, this would not change the semantics of _other_ modules requiring outside of `init.luau`, correct? So (in the example tree you gave) `require("./package/dependency")` and `require("./package")` would still work the same?

--- 

> Because we consider `package/init` to be an ordinary Luau module, it is considered to be rooted at `package`. Therefore, a relative `require('./foo')` call will resolve to `./package/foo.luau` rather than `./foo.luau`.
> 
> Rojo, by contrast, considers `./package/init.luau` to belong to the containing directory. A developer using Rojo must instead write `require('./package/foo')`. This is only necessary when the script is named `init.luau`.

I feel as though this doesn't really represent the problem very well. The issue isn't what Rojo "considers" a require to be, since we don't do any of that. The issue is Rojo interprets a directory with `init.luau` in it as being a `ModuleScript`, so when it enters Roblox it "changes location" so to speak. Here's a visual example for anyone unfamiliar:

image

It's also we can't fix it on Rojo's end. A toggle for Rojo to make it not do that would solve the problem for any _new_ requires, but the vast majority of Rojo projects expect this behavior so it could not be the default. And since it isn't the default, any non-Roblox projects that want to be compatible with Roblox would need a Rojo project file to set the flag. I feel as though requiring a compatibility layer harms the entire point of trying to maintain compatibility.

> On a filesystem, Rojo's behaviour is frankly very weird! 

I'd like to defend this design choice a bit. It is an _extremely_ common pattern in Roblox to have scripts that contain children. This is outright incompatible with the file system though: you cannot have files which contain other files. The only way to represent that is with directories. This is something Roblox's own file syncing project has ran into, and it's one of the fun and exciting parts of this ecosystem.

While Rojo could have just not supported nesting scripts and forced people to use directories and structure their code around it, there wasn't really a compelling reason to do that. Rojo long predates Luau as a project, and it wasn't a consideration when making the design. The reason Rojo uses `init.lua(u)` for stuff is because that was just a convention: Lua 5.1 uses `init.lua`, and precursors to Rojo also used `init.lua` to solve this problem.

> We can mitigate this with a new lint: If a module is the parent of other modules, it is poor style for it to directly import sibling or parent modules. Files named `init.luau` should never issue a require that resolves to a sibling module. In our example, we would warn if `package/init.luau` were to `require('./foo')` because that reaches outside of its folder.

I'm not sure if this is reasonable long-term. While I don't have any examples of wanting to require parent modules, I can see wanting to require siblings. An obvious use case would be the infamous "Util.luau" file that so many projects end up having. I don't feel strongly enough to oppose a lint for this but I'm just raising it as a thought.

> Adjusting Rojo to match what Luau does today without breaking any existing application code is tricky and most likely requires something far weirder than the change outlined in this RFC.

I'll do you one better: I'm not convinced it's actually possible. I hope I do not need to describe the difficulty of parsing user code and analyzing what modules it requires and how it requires them (i.e. whether the require needs to be fixed, and where their path has to be fixed to

andyfriesen:
> Just to clarify, this would not change the semantics of other modules requiring outside of `init.luau`, correct? So (in the example tree you gave) `require("./package/dependency")` and `require("./package")` would still work the same?

Assuming that those requires were issued from `./foo.luau`, those two calls would resolve to the files `./package/dependency.luau` and `./package/init.luau`.

> I feel as though this doesn't really represent the problem very well. The issue isn't what Rojo "considers" a require to be, since we don't do any of that. The issue is Rojo interprets a directory with init.luau in it as being a ModuleScript, so when it enters Roblox it "changes location" so to speak. Here's a visual example for anyone unfamiliar:

That's also an accurate characterization of what happens.  I think we are in complete agreement.

> I'd like to defend this design choice a bit. It is an extremely common pattern in Roblox to have scripts that contain children. This is outright incompatible with the file system though: you cannot have files which contain other files. The only way to represent that is with directories. This is something Roblox's own file syncing project has ran into, and it's one of the fun and exciting parts of this ecosystem.

I don't mean to pass judgment.  I mean only to point out that there is a significant drawback to Rojo's encoding.  We acknowledge and accept that drawback as part of the tradeoff being proposed in this RFC.

> I'll do you one better: I'm not convinced it's actually possible. I hope I do not need to describe the difficulty of parsing user code and analyzing what modules it requires and how it requires them (i.e. whether the require needs to be fixed, and where their path has to be fixed to

I have some ideas and some of them might work, (some do not require parsing any code) but I don't think the extra complexity would be worth it.

dphblox:
For my understanding, does this mean that any existing use of `@self` in any Luau environment will now be erroneous? Or is this just for the Roblox environment, and other environments won't have this?

dphblox:
Yeah that makes sense. I agree with the idea to reserve `@self` instead of `@`.

vrn-sn:
Could we make `@self` overridable by `.luaurc` configs (i.e. treat it as if it's "defined" in a top-level configuration file)?

In Roblox, for example, `@self` is a replacement for the `script` global, which can technically be shadowed/overidden by scripts.

dphblox:
Would we want to encourage the practice of overwriting the definition of `@self`? And would we want those changes to be localised to a file, or to apply to all descendants in the module hierarchy?

I'm not sure it'd be a good idea to allow overwriting `@self`; even in Roblox, I'd consider assigning to `script` to be an antipattern unless you're explicitly deleting it for some technical reason.

Dekkonot:
I don't think there's really any harm in it, because it helps make this less breaking than it already is.

I'd agree that it's an anti-pattern to override it, but at the same time is that really our problem? I'm inclined to let people do things if they want since it won't impact most users either way.

vrn-sn:
This feels well-suited for static analysis tools to call out as an anti-pattern, but I'm not sure the "good pattern" has to be enforced at runtime.

deviaze:
I like `@self`! 

Just to confirm, will `@self` will be visible in all Luau modules, but only be particularly useful if used in a directory with an `init.luau` (making the parent directory the module root)? Or will `@self` only be available in files named `init.luau`?

So:
```
-src/
--discordluau/
---init.luau -- @self refers to ./src/discordluau/
---src/
----init.luau -- @self refers to ./src/discordluau/src/
----channels.luau -- @self refers to ./src/discordluau/src/channels or invalid alias?
```

Also, I think Micah briefly touched on this but I want to reiterate; in my opinion, I disagree with an `init.luau` requires sibling lint. 

I feel it's perfectly okay to require `./foo` from an `init.luau` because not all multi-file modules are standalone libraries. Lots of the time you want to abstract part of your codebase functionality into separate files and putting them in a directory with an `init.luau` is the most natural way to do that in Luau. In fact, you can argue that `init.luau` should be the only place you require sibling modules in that structure.

vrn-sn:
This is correct, @deviaze:
> @self will be visible in all Luau modules, but only be particularly useful if used in a directory with an init.luau (making the parent directory the module root)

For the purposes of require-by-string, `foo.luau` is just sugar for `foo/init.luau` that someone can use if they know the module `foo` is a leaf node in the module hierarchy. In other words, a folder `foo` containing only an `init.luau` file is exactly equivalent to defining the module as `foo.luau`.

A generalized implementation of the navigation for require-by-string won't even be aware of `init.luau` files—they exist only as an implementation detail of representing nestable modules on a filesystem. Because `foo/init.luau` and `foo.luau` represent identical modules, there is no difference in the semantics for `@self`.

docs/abstract-module-paths-and-init-dot-luau.md
```diff
@@ -0,0 +1,141 @@
...
+
+The central feature of this RFC is about how module paths correspond to
+filesystem paths: A module path refers to a module if it corresponds either to a
+`.luau` file or to a directory that contains a file named `init.luau`.  A module
+path refers to a directory if it refers to a filesystem directory that lacks an
+`init.luau` file.
+
+More concretely, Luau will no longer consider relative requires from a package
+`init.luau` file to resolve relative to the script itself.  It will instead
+resolve relative to the script's parent, i.e. the folder containing the module.
```
vrn-sn, line 88:
```suggestion
resolve relative to the location of the abstract module it represents, i.e. the
location of its parent folder in a filesystem context.
```
For clarity: "module" means the abstract module visible to require-by-string; terms like "file" and "folder" specifically refer to the filesystem representation of modules.

docs/abstract-module-paths-and-init-dot-luau.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+## Drawbacks
+
+On a filesystem, Rojo's behaviour is frankly very weird!  A reader must know the
+name of the current source file in order to know what `require('./foo')` will
+do, and if a developer renames a file to or from `init.luau`, they will be
+forced to rewrite every `require()`.
+
+We can mitigate this with a new lint: If a module is the parent of other
```
vrn-sn, line 113: We're at a good place with this RFC, and we've hit the 2 week minimum time now as well. I think this section about the lint is probably the only remaining piece to figure out. I'd propose we just leave this out—it's a bit weird on the filesystem but makes more sense in the context of Roblox, so I'm not sure we should prescribe anything at the language level here.

vrn-sn (approved):

aatxe (approved):

PR #110, opened 2025-03-27
RFC: `if` and `while` statement initializers (`if local` and `while local` statements)

deviaze:
This RFC is an update/continuation/alternative to *if statement initializers* (#23), with updated fallthrough semantics, the addition of `while local` statements, and pancakes (stacked `local` initializations in `if local` statements).

[Rendered](https://github.com/deviaze/luau-rfcs/blob/syntax-if-while-locals/docs/syntax-if-while-local-initializers.md).

Basically, 

```luau
if local index = table.find(list, element) then
    table.remove(list, index)
end

while local line = nextline() in line ~= "" do
    -- do something with line
end
```

in Roblox examples:

```luau
if local character = player.Character then
    -- character is not nil and is bound here
    if local humanoid = character:FindFirstChildOfClass("Humanoid")
        in humanoid.Health < 20
    then
        -- humanoid is not nil, is bound here, and humanoid's health < 20
    else
        -- humanoid is not bound here
    end
end
-- character not bound here
```

and as a better way of handling nested `if local`s,

```luau
if
    local character = player.Character
    local head = character:FindFirstChild("Head") in notTooFarAway(head.Position)
    local humanoid = character:FindFirstChildOfClass("Humanoid")
    local humanoidRootPart = character:FindFirstChild("HumanoidRootPart")
then
    -- condition executes if all the above are guaranteed to be not-nil here and `head` is notTooFarAway
else
    -- handle your fallback cases
end
```

This RFC turned out quite long, apologies, but the discussion on *if statement initializers* was pretty long and there's a lot to keep track of and address, especially in terms of alternatives. Please enjoy!

### DIFF
--- a/docs/syntax-if-while-local-initializers.md
+++ b/docs/syntax-if-while-local-initializers.md
@@ -1,0 +1,228 @@
+# `if local` and `while local` statements
+
+This RFC is an update and continuation to [if statement initializers](https://github.com/luau-lang/rfcs/pull/23), featuring improved semantics as discussed and agreed-upon in that RFC's thread and in ROSS.
+
+## Summary
+
+`if local statements`: Allow `local` identifiers to be bound in `if` statements to improve the ergonomics of extremely common control flow idioms, improve code clarity, reduce scope pollution, and improve the developer experience.
+
+`while local statements`: Allow `local` identifiers to be bound in `while` statements to improve code clarity, improve sentinel value handling semantics, reduce scope pollution, and provide parity with `if local` statements.
+
+## Motivation
+
+In Luau, an extremely common idiom is for fallible functions to return an optional value: an intended result if the operation succeeds or `nil` if it fails. Users are expected to *nilcheck* this result to handle success and failure/empty cases, and this constitutes a major aspect of control flow. An extremely common example of such is Roblox's `Instance:FindFirstChild`, which returns an `Instance` if one was found or nil otherwise:
+
+```luau
+local model = workspace:FindFirstChild("MyModel")
+if model then
+    -- model is bound and not nil
+end
+-- model is still bound here
+```
+
+With `if local` statements, this code may be rewritten as:
+
+```luau
+if local model = workspace:FindFirstChild("MyModel") then
+    -- model is bound and is not nil
+end
+-- model is not bound here
+```
+
+In many cases, developers use an expression in an if statement's condition and then immediately use it again in its body:
+
+```luau
+if folders[folder][file_name].last_updated < now - TWO_DAYS then
+    last_updated = folders[folder][file_name].last_updated
+end
+```
+
+In this case, an `if local` statement with an `in` clause can reduce repetition and greatly improve readability:
+
+```luau
+if local update_time = folders[folder][file_name].last_updated 
+in update_time < now - TWO_DAYS then
+    last_updated = update_time
+end
+```
+
+The primary motivation for `if local` statements isn't in small examples like those above, however, it's how it fits into whole codebases. `if local`s drastically improve code shape, readability, and the general conciseness and expressiveness of the Luau language.
+
+<!-- TODO: add evidence -->
+
+`while local` statements allow you to set a value that persists throughout the execution of the loop as well as an `in` condition that re-evaluates every iteration of the loop.
+
+In this case, `current_path` persists throughout all iterations, and without an `in` clause, the `while local` defaults to `while true do` behavior:
+
+```luau
+local init_luau_path = ""
+while local current_path = provided_path do
+    if local init_luau = fs.find(path.join(current_path, "init.luau")).file then
+        init_luau_path = init_luau
+        break
+    elseif local parent = path.parent(current_path) then
+        current_path = parent
+    else
+        error("ran out of parents")
+    end
+end
+```
+
+In another case, this `while local` loop allows for easy request retries:
+
+```luau
+local result
+local retries = 0
+
+while local response = http.get("https://my.unreliable.dev/api/") 
+in 
+    response.status_code ~= 200
+    and retries < 3 
+do
+    if response.status_code == 429 then
+        task.wait(response.headers["Retry-After"] or 3)
+    elseif response.status_code == 404 then
+        break
+    end
+    
+    local new_response = http.get("https://my.unreliable.dev/api/")
+    if new_response.status_code == 200 then
+        result = new_response.body
+    else
+        retries += 1
+        response = new_response
+    end
+end
+```
+
+## Design
+
+This proposal introduces `if local` and `while local` statements, or more precisely, allows `local` bindings to be initialized within `if` and `while` statement declarations.
+
+### `if local` statements
+
+An `if local` statement is any `if` statement with one or more `local` bindings. `local` bindings may be declared after the `if` or `elseif` keywords of an `if` statement, may be followed by one `in` clause expression, and must be followed by the `then` keyword:
+
+```luau
+if local identifier = expression() then
+end
+-- or
+if local identifier = expression() in condition() then
+end
+```
+
+If `local` bindings are provided, then one optional `in` clause may be provided per branch to partially determine the evaluation condition of the `if/elseif` branch.
+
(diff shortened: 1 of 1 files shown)


dphfox:
The `in` syntax doesn't sit right with me, though I like this idea in concept.

Perhaps a semicolon is a more natural statement separator? And perhaps that leads to the notion of those `in` expressions not actually being special at all, and just acting like `and` except it can terminate a statement fully (i.e. evaluate a `local`) and thus let you access previous identifiers.

```lua
if
    name ~= nil;
    local player = Players:FindFirstChild(name);
    local character = player.Character;
    local humanoid = character:FindFirstChildOfClass("Humanoid");
    humanoid.Health > 0
then
    humanoid.Health -= 5;
end
```

A simpler example:

```lua
if local match = s:match(REGEX); #match > 2 then
    print("found it!")
end
```

I think that, while I appreciate the attempt to use a keyword in line with the rest of Luau, that it's better to use a symbol that people can assign meaning to themselves, rather than constructing an explicit phrase that reads wonkily.

deviaze:
> The in syntax doesn't sit right with me, though I like this idea in concept.
> Perhaps a semicolon is a more natural statement separator? And perhaps that leads to the notion of those in expressions not actually being special at all, and just acting like `and` except it can terminate a statement fully (i.e. evaluate a local) and thus let you access previous identifiers.

Hmm, `in` was chosen in *if statement initializers* last year after a lot of bikeshedding between `where` vs `in` vs `;`, so I basically adopted it towards this RFC. I can see why it sounds a bit wonky ~~without an OCaml background~~ because you have to explain "it's `in` because you're putting the identifier *into* the expression"; I think @aaxte was the person who was most pro-`in` last time, so maybe she could share her thoughts on the keyword?

Otherwise, I kinda like the idea that you can mix regular `if` expressions and `if local`s, especially because it better implies that the `local` can't be nil, but I also like `in`s because they're more visually distinct. I'm not a huge fan of putting semicolons everywhere, maybe we could require semicolons only to separate `local` assignments from standard expressions?

For example:

```luau

if
    name ~= nil; -- we don't even need this one tbh
    local player = Players:FindFirstChild(name)
    local character = player.Character
    local humanoid = character:FindFirstChildOfClass("Humanoid");
    humanoid.Health > 0
then
    humanoid.Health -= 5
end

**Comment by dphblox** [comment]
We could adopt the same tack as semicolons in normal Luau; can be used to disambiguate statements, but when there's no ambiguity perhaps there's no need for them.

**Comment by dphblox** [comment]
I suppose another keyword-based alternative would be `and if`, which would only require one token of lookahead to disambiguate;

```
if name ~= nil
    and if local player = Players:FindFirstChild(name)
    and if local character = player.Character
    and if local humanoid = character:FindFirstChildOfClass("Humanoid");
    and if humanoid.Health > 0
then
    humanoid.Health -= 5
end
```

**Comment by deviaze** [comment]
I'm reminded of a humorous idea from RFC's comments.. `in` ->`if` as `if` guards. It's kinda wack when you're talking about single `if local` but actually might just make sense with `local` pancakes...

```luau

if
    local character = hit.Parent if character:IsA("Model")
    local hit_player = Players:GetPlayerFromCharacter(model) if hit_player.Team ~= player.Team
    local humanoid = character:FindFirstChildOfClass("Humanoid") if humanoid.Health > 0
then
    humanoid.Health = math.max(0, humanoid.Health - 5)
end
```

it's definitely parseable in the face of `if` expressions because you can't chain 2 `if` expressions without an operator.

```luau
if local saying = if obj.animal == "Cat" then "meow" elseif obj.animal == "Dog" then "arf" else getsaying(obj.animal) if #saying > 0 then
    print(saying)
end
```

will be fun but I don't think is *too* bad anyway.

dphblox:
I was going to suggest `and if` before but it's ambiguous for the same reason.

deviaze:
That's a really good point on the information flow direction; I was somewhat worried about the same kind of thing when I first saw `in` clauses and I didn't \*get\* the reasoning why `in` would make sense there, but I didn't know how to express it at the time. I'd say that putting variables `in` an expression is mainly intuitive to OCaml programmers, who use `in` for left -> right information flow all the time:

```ocaml
let () = 
  let a = 1 in 
  let b = a + 1 in
  print_endline(string_of_int b)
```

And I'd say Luau programmers would be a lot more familiar with Python, Lua, JavaScript, (and for compiled languages, probably Rust, C/++, etc.) over OCaml, and so `in` could be confusing when it's an operator in Python and used in `for` loops in Luau (and other languages).

For what it's worth, we aren't actually locked into existing keywords here; in fact, the previous rfc originally suggested a new conditional keyword `where` to demark the conditional clause.

Although I like the idea of making the evaluation condition non-special, I'm wondering how that could affect its implementation. From what I know from last time I talked to Luau engineers about this, it was reasonable to special case `local`s into a control structure, but I'm not sure if it's possible to do that if subsequent locals refer to previous ones, and if non-`nil` check has to be evaluated one at a time. That's why I suggested that

```luau
-- using ; instead of in for ya
if
    local character = hit.Parent; character:IsA("Model")
    local hit_player = Players:GetPlayerFromCharacter(model); hit_player.Team ~= player.Team
    local humanoid = character:FindFirstChildOfClass("Humanoid"); humanoid.Health > 0
then
    humanoid.Health = math.max(0, humanoid.Health - 5)
end
```
Could expand to

```luau
if local character = hit.Parent; character:IsA("Model") then
    if local hit_player = Players:GetPlayerFromCharacter(model); hit_player.Team ~= player.Team then
         if local humanoid = character:FindFirstChildOfClass("Humanoid"); humanoid.Health > 0 then
             humanoid.Health = math.max(0, humanoid.Health - 5)
         end
    end
end
```
if it's not possible to get a more efficient/bespoke pancake implementation.

Also just looking at the syntax highlighting for the above two.. I kinda want my red `in`s back :sob:

Bottersnike:
I really don't like that it does a nilness check not a truth check as the primary condition. Semantics for `if` are pretty well established that `if false then ... end` won't execute that block of code and this feels like it would just be confusing. In fact, the given
```luau
if local success, result = pcall(foo) then
    if success then
```
example demonstrates confusion as one might expect the `if local` to be checking `success` for us already. Similarly this example is a tautology as pcall will never return `nil`.

+1 on the dislike of `in` too for pretty much all the reasons stated above. Along with the confusion with other languages where `x in y` is a membership check, the existing use of `in` in luau is as part of iteration over members which is a very different use to this.

LoganDark:
I also dislike the use of `in`. Why *not* `where`? `if local humanoid = character:FindFirstChildOfClass("Humanoid") where humanoid.Health < 20 then`

`in` makes no sense to me, and semicolons are kind of ew.

deviaze:
Thank y'all for the feedback! Changed the nilcheck default to a truthiness check, and after using Rust's new `let` chains for a bit I was inspired to add it as another currently-considered option. If we're okay with adding a new operator or figuring out a new keyword for it, I think it'd be a much cleaner option than the current mental model around `if local`s.

I also empathize with the why `in` comments! The keyword was debated ad-nauseum last RFC and was eventually changed from `where` to `in`, but imho it's not an intuitive word choice for most Luau users atm.

xHeptc:
@jackdotink elaborate? Just blatantly saying "zero complexity to the parser" and "would not require complex compilation or bytecode" without providing any supporting reasoning or concrete examples.

xHeptc:
@jackdotink for the record, I never associated complexity with parser, I just mentioned it will need new rules, I've myself written parsers before, and I know all it needs is just to dupe VariableDeclaration statement and put it in primary, following some type of precedence.

But question is, how would the scope work exactly? How would a hidden scope be created?

xHeptc:
@itsmath my mistake, I've never coded in vanilla lua, didn't know there was a `in` keyword up until now.

felixsidzed:
> @itsmath my mistake, I've never coded in vanilla lua, didn't know there was a `in` keyword up until now.

clearly you did
image

annieetang:
hello! we discussed this RFC as a team a bit more and liked the general idea, so we wanted to put out a tighter scoped RFC for just the `if local` part for now, until we have more consensus on chaining and extending the feature (noted in the RFC's future work section). Would love any thoughts on this! https://github.com/luau-lang/rfcs/pull/238

PR #111, opened 2025-04-03
Propose SYNTAX; Walrus-Operator

ssynical:
:= 

Thanks!

### DIFF
--- a/docs/syntax_walrus-operator.md
+++ b/docs/syntax_walrus-operator.md
@@ -1,0 +1,67 @@
+# Syntax-Assignment-Expression: Walrus Operator (:=)
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


dphblox:
While I think it's interesting to toy with the idea of this, I'm not sure if it's good style to support this outside of very limited applications in e.g. the if-local/while-local RFC. I have reservations about whether the downsides to complexity and code clarity would be worth the upside of being able to define locals at any point in the program, especially where this starts to interact with more complex expressions or conditionals.

andyfriesen:
I agree with @dphblox.

I would consider the current restriction on assignments to be a feature.

andyfriesen:
Thanks for the submission, but this feature is not consistent with the direction we'd like to take for Luau.

PR #113, opened 2025-04-07
Implicitly convert strings in type userdata methods that accept a `key` arg and allow strings as keys in the `props` table of `types.newtable`

gaymeowing:
RFC name is a mouthful, but couldn't think of a better name.

[Rendered](https://github.com/gaymeowing/rfcs/blob/type-functions-strings-as-keys/docs/user-defined-type-function-strings-as-keys.md)

### DIFF
--- a/docs/user-defined-type-function-strings-as-keys.md
+++ b/docs/user-defined-type-function-strings-as-keys.md
@@ -6,9 +6,7 @@ This RFC proposes allowing strings as keys for methods on the `table` type userd
 
 ## Motivation
 
-Currently having to always wrap string literal keys in `types.singleton` is annoying, and becomes especially annoying when using `types.newtable` due to having to write `[types.singleton("key")]` for every single key in the props table.
-
-This leads to developers having to write the following:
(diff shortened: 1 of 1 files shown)


gaymeowing:
Realized it'd be good for `types.unionof()` to accept string args aswell, but debating on if it should be included or if a new rfc for a new method should be made (currently thinking of having `types.enum("meow", "mrrp")` with it returning `"meow" | "mrrp"`).

gaymeowing:
Had some thoughts on this recently, and I think adding a new global `totype` to the type function environment would be better than what this RFC is proposing currently.

It'd be basically how `typeof` works in types, except for type functions.

gaymeowing:
> Had some thoughts on this recently, and I think adding a new global `totype` to the type function environment would be better than what this RFC is proposing currently.
> 
> It'd be basically how `typeof` works in types, except for type functions.

This would also make typing out singletons for keys much more pleasant, as I've found it annoying to have to type out `table_type:readproperty(types.singleton("key"))` just to index a tables property in a type function. Whereas with `totype` this would be much simpler as I could just write `table_type:readproperty(totype("key"))`.

PR #114, opened 2025-04-11
types.optional

gaymeowing:
[Rendered](https://github.com/gaymeowing/rfcs/blob/types.optional/docs/types-library-optional.md)

### DIFF
--- a/docs/types-library-optional.md
+++ b/docs/types-library-optional.md
@@ -6,7 +6,7 @@ This RFC proposes adding a new method to the `types` library in type functions, 
 
 ## Motivation
 
-Currently its annoying to have to always type `types.unionof(meow, types.singleton(nil))` to make a optional type, this leads to having the following appear in codebases that make use of optional types a lot in type functions:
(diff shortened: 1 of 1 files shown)


docs/types-library-optional.md
```diff
@@ -0,0 +1,19 @@
+# types.optional
+
+## Summary
+
+This RFC proposes adding a new method to the `types` library in type functions, that provides a shorthand for `types.unionof(type, types.singleton(nil))`.
+
+## Motivation
+
+Currently its annoying to have to always type `types.unionof(meow, types.singleton(nil))` to make a optional type, this leads to having the following appear in codebases that make use of optional types a lot in type functions:
```
aatxe, line 9:
```suggestion
Optional types occur very commonly, and it is annoying to have to always type `types.unionof(meow, types.singleton(nil))` to make a optional type, this leads to having the following appear in codebases that make use of optional types a lot in type functions:
```

aatxe (approved):

PR #116, opened 2025-04-12
 RFC: Support for Type Packs in User-Defined Type Functions 

TenebrisNoctua:
[Rendered](https://github.com/TenebrisNoctua/rfcs/blob/support-for-type-packs-in-usdf-type-functions/docs/support-for-type-packs-in-user-defined-type-functions.md)

### DIFF
--- a/docs/support-for-type-packs-in-user-defined-type-functions.md
+++ b/docs/support-for-type-packs-in-user-defined-type-functions.md
@@ -13,16 +13,16 @@ This limits the usage for packs as they are used often in the language, and thei
 
 ## Design
 
-We introduce a new runtime type called `pack` to extend the support of type packs within the type runtime.
+We introduce a new userdata called `pack` alongside `type` to extend the support of type packs within the type runtime.
 
-Packs will have a `kind` property that specifies the kind of the pack. A pack can either be a `list` kind or a `variadic` kind.
+`pack` instances will have a `kind` property that specifies the kind of the pack. A `pack` can either be of a `literal` kind or of a `variadic` kind.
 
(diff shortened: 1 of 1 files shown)


docs/support-for-type-packs-in-user-defined-type-functions.md
```diff
@@ -0,0 +1,127 @@
...
+type A = f<(string, number)>
+```
+
+Type-packs can now be placed at the `tail` argument of function types:
+
+```luau
+type function f() 
+    local newPack = types.pack({types.string, types.number, types.pack({any}, true)}) -- (string, number, ...any)
+    local newfunc = types.newfunction()
+    newfunc:setparameters({}, newPack)
```
PhoenixWhitefire, line 48: It feels quite strange to me to have two different kinds of "tails". Why can't they be represented with a `{ head: {type}?, tail: type }` table, like function parameters and returns are today? It also seems inelegant to me that, despite function parameters and returns themselves being packs, to set them as these new "pack" types they have to have an empty `{}` as the "head" even though packs themselves have heads.

TenebrisNoctua: Because I believe that type packs should not be represented as two table properties, and a type pack can contain multiple packs inside of itself. A "head" and "tail" representation does work today, but personally I thought that a representation that treats a pack as an array of types or packs in the API is better.

The "head" and "tail" fields also create the problem you've noticed. These fields describe two portions of the same type pack, so it would be quite inaccurate and weird to place a `pack` object in one of them. 

I'm not sure how to solve this problem, though. Perhaps we can overload `:setparameters()` to accept one `pack` argument alongside a `head` and a `tail`, if we wish to preserve backwards compatibility, or we create a whole new method for the function `type` object where it accepts a `pack` object to set the parameters and returns type packs.

PR #117, opened 2025-04-12
RFC: Support Variadic User-Defined Type Functions

TenebrisNoctua:
[Rendered](https://github.com/TenebrisNoctua/rfcs/blob/variadic-type-functions/docs/variadic-user-defined-type-functions.md)

### DIFF
--- a/docs/variadic-user-defined-type-functions.md
+++ b/docs/variadic-user-defined-type-functions.md
@@ -1,0 +1,54 @@
+# Variadic User-Defined Type Functions
+
+## Summary
+
+Add support for variadic arguments for user-defined type functions.
+
+## Motivation
+
+Today, user-defined type functions can only take a pre-defined set of parameters. This limits the usage in a way that makes it not possible to make the function variadic, where the user can supply
+the function as many arguments as provided.
+
+By adding support for variadic functions, we allow users to call user-defined type functions with as many arguments as they wish, providing a more dynamic system.
+
+## Design
+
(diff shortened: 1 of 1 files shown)


cheesycod:
Personally, I fully disagree with the rfc on the basis that it’ll probably lead to postponing the new type solvers release by several months if not years due to all the bugs that’ll probably arise from variadic type functions.

Other than that, I personally fully agree with the idea.

TenebrisNoctua:
> Personally, I fully disagree with the rfc on the basis that it’ll probably lead to postponing the new type solvers release by several months if not years due to all the bugs that’ll probably arise from variadic type functions etc. and I think it’s more important to get all the bugs fixed first over adding new features (and this one, from a cursory read, feels pretty complicated implementation wise)
> 
> Other than that, I personally fully agree with the idea.

I understand the concern, but I believe it won't cause any important delays for the new type-solver's release.
I also don't see the potential for many bugs either, it behaves the same as normal variadic functions, except now it's in the type-runtime.

znotfireman:
This RFC doesn't account for multiple variadics/type packs, eg. `type a = f<("a", "b"), ("c", "d")>

Maybe type packs can be represented as an array of types?

```luau
type function f(first, second)
    assert(typeof(first) == "table")
    assert(typeof(second) == "table")
    for _, type in first do
    end
    for _, type in second do
    end
end

type a = f<("a", "b"), ("c", "d")>
```

PR #118, opened 2025-04-21
RFC: Declaration syntax 

lodinukal:
[Rendered](https://github.com/lodinukal/rfcs/blob/declaration-syntax/docs/syntax-declaration.md)

### DIFF
--- a/docs/syntax-declaration.md
+++ b/docs/syntax-declaration.md
@@ -2,17 +2,17 @@
 
 ## Summary
 
-Formalizing the luau declaration syntax used in the luau parser.
+Formalizing the Luau declaration syntax used in the Luau parser.
 
 ## Motivation
 
-RIght now the declaration syntax used by the luau parser is undocumented and an unstable implementation detail. By defining the syntax, it allows for embedders of luau to have typechecking for their apis without having to resort to empty modules. This should allow tools like luau-lsp to be able to provide easier ways to use these declarations.
+RIght now the declaration syntax used by the Luau parser is undocumented and an unstable implementation detail. By defining the syntax, it allows for embedders of Luau to have typechecking for their apis without having to resort to empty modules. This should allow tools like luau-lsp to be able to provide easier ways to use these declarations.
 
 ## Design
 
 Declaration syntax should only be usable in definition files. Definition files have the extension of `.d.luau` and are used to provide type information for modules. They should not be allowed to do anything beyond declaring type aliases and declarations.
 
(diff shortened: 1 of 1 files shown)


gaymeowing:
I think class as a keyword should be removed from typedefs, it exists currently because theres no way to express nominal types with the current type system outside of typedef files, and a rfc should be made for some sort of nominal type syntax. 
As there is no such thing as a "class" in luau like there is in object oriented languages like java.

docs/syntax-declaration.md
```diff
@@ -0,0 +1,105 @@
...
+
+More examples can found in [luau-lsp's globalTypes.d.luau](https://github.com/JohnnyMorganz/luau-lsp/blob/main/scripts/globalTypes.d.luau) and [Luau's parser tests](https://github.com/luau-lang/luau/blob/master/tests/Parser.test.cpp#L1908)
+
+## Drawbacks
+
+A drawback would be that this is an unstable implementation detail, which allows for the parser to freely change the syntax without worry for backwards compatibility. By formalizing the syntax, it would mean more consideration would be needed when changing how declarations are parsed in the future.
+
+## Alternatives
+
+The alternative to this is to leave this as an implementation detail. However this makes it inconvenient for embedders of Luau to use the declaration syntax.
```
aatxe, line 105: How does it make it inconvenient to use? luau-lsp is using it right now.

docs/syntax-declaration.md
```diff
@@ -0,0 +1,105 @@
+# Declaration syntax
+
+## Summary
+
+Formalizing the Luau declaration syntax used in the Luau parser.
+
+## Motivation
+
+RIght now the declaration syntax used by the Luau parser is undocumented and an unstable implementation detail. By defining the syntax, it allows for embedders of Luau to have typechecking for their apis without having to resort to empty modules. This should allow tools like luau-lsp to be able to provide easier ways to use these declarations.
```
aatxe, line 9: luau-lsp already uses declaration files as they are, so it's not really clear what we hope to actually gain here. the syntax of declaration files has been essentially unchanged since they were introduced, and it seems like the only thing this RFC has actually done is _inspire_ conversation to make changes to it.

aatxe (changes requested):
Overall, my feeling about this RFC is that it doesn't actually _do_ anything. The syntax for declaration files has been just as stable as the rest of Luau's syntax has been, and it's not clear what problem we're aiming to solve by having an RFC that just says it "formalizes" the syntax without actually formalizing the syntax at all. There's no formal characterization of the syntax present in this RFC in any form, and the motivation for it as-stated is completely imagined since it talks about luau-lsp using declaration files when it already does use them.

PR #119, opened 2025-04-23
Unordered `table.remove()`

VegetationBush:
# Unordered `table.remove()`

## Summary

Add an unordered remove method to the table library.

## Motivation

For arrays that are known to be unsorted, using `table.remove()` to remove indicies is unperformant, as `table.remove()` takes O(n) time to complete. This is a gross use of resources, especially if this behavior is not required by the user. Thus, it's logical to conclude that an analogous method that performs an *unordered remove* would be the clear next step.

## Design

The naming of the new method is proposed as `uremove`. However, for clarity reasons, this may not be the best name choice as `uremove` types very similar to `remove`. The implementation of the method could be as such:
```lua
function table.uremove(t: {any}, index: number)
	t[index], t[#t] = t[#t], nil
end
```
Valid argument checking is optional.

## Drawbacks

Possibly none, apart from an extra method in the table library.

## Alternatives

Users can define the function themselves.


aatxe:
The PR must add the file for an RFC, not just contain it in its description.

aatxe:
This RFC does not meet the standards for inclusion as a builtin API. It does not provide any justification for why this should be built in, and it even provides the simple one line of Luau code that already implements this behavior today. In order to be worth adding as a builtin function, there must be a substantial benefit in terms of performance, ease of implementation, expressivity of the language, etc. supported by its inclusion, and this falls very short of that.

VegetationBush:
Hi, apologies I couldn't meet the standards. I am extremely new to GitHub, and I have no idea what I'm doing. This was my best shot at trying to follow the guidelines. Continuing:

The prevelance of `table.remove()` in code is fairly high. Going off the previous observation that most instances of this method is used on arrays that are unsorted/non-stable, this method can replace table.remove() to significantly reduce computational overhead.

For the argument of being a simple one-liner. Yes, it can be done in one line. However, just because it is trivial to write does not mean that it clearly communicates the idea. A dedicated method would drastically improve relevant code clarity.
Consider the recent addition of `math.lerp()`. The initial lerp calculation, `a + (b - a) * t`, is trivial to write, and the format is arguably more well known than that of `table.uremove()`. But again, it was still implemented (obviously, for other smaller reasons, such as precision). The implementation not only increased the clarity, but also came with micro performance boosts as well. Similarly, because Lua tables are funky, there is definitely grounds for backend optimizations for read/write access.

PR #121, opened 2025-04-27
RFC: Definition files specified in luaurc

ffrostfall:
[Rendered](https://github.com/ffrostfall/rfcs/blob/luaurc-definition-files/docs/luaurc-definition-files.md)

### DIFF
--- a/docs/luaurc-definition-files.md
+++ b/docs/luaurc-definition-files.md
@@ -1,0 +1,21 @@
+# Definition Files in .luaurc Configs
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


aatxe:
I suspect that the actually good solution from this, that doesn't lead to people doing very silly things with their projects (like defining tons of declaration files for luau code they've written), is to handle runtime registration at a build tooling level and have that inform the LSP and so forth. Not necessarily hard blocking this on that, and I recognize that it may be worth introducing something before we get to that point, but the long view IMO is that projects will declare the runtimes they support, registries will know what runtimes are, the package manager will pull type definitions for the appropriate runtimes the project is declared to support, and you'll be able to systematically check locally (and make available to the LSP) those definition files.

PR #122, opened 2025-05-06
RFC: 'noinfer' Type Operator

hardlyardi:
[Rendered](https://github.com/hardlyardi/rfcs/blob/rfc-noinfer-type-operator/docs/noinfer-type-operator.md)

### DIFF
--- a/docs/abstract-module-paths-and-init-dot-luau.md
+++ b/docs/abstract-module-paths-and-init-dot-luau.md
@@ -1,0 +1,142 @@
+# Abstract module paths and `init.luau`
+
+## Summary
+
+We will redefine `foo/init.luau` to be the contents of the module named foo for
+the purpose of resolving relative `require()` calls within its body.
+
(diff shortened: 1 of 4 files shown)


deviaze:
We def need this functionality in the new solver, but I think I'd prefer it being a `@noinfer` attribute instead of an annotation that looks like a type function.

For example (using ariel's example from discord):

```luau
local function greedy_insert<V>(tbl: { V }, value: @noinfer V)
  -- ...
end
local a: "a"
local b: "b"
local some_table = { a, b }
local c: "c"
greedy_insert(some_table, c) -- TypeError: Type '"c"' could not be converted into '"a" | "b"'
```

daimond113:
I think using generic constraints once they are implemented instead of adding a separate feature for this will be better. For example, a simple setter function in TypeScript looks like this: 

```typescript
function set<const S, V extends S>(source: S, value: V)
```

This leverages generic constraints to emulate greediness without requiring a separate feature.

PR #125, opened 2025-05-30
RFC: Support Luau-syntax configuration files

vrn-sn:
[Rendered](https://github.com/vrn-sn/rfcs/blob/vrn-sn/rfc-luauconfig/docs/config-luauconfig.md)

### DIFF
--- a/docs/config-luauconfig.md
+++ b/docs/config-luauconfig.md
@@ -1,0 +1,129 @@
+# Support Luau-syntax configuration files
+
+**Status**: Not implemented
+
+## Summary
+
+Introduce `.luauconfig`: a Luau-syntax file used to configure analysis and runtime behavior.
+
+## Motivation
+
+We currently support analysis and runtime configuration in `.luaurc` files, which use a JSON-like syntax ([RFC](https://rfcs.luau.org/config-luaurc)).
+
+Under the current design, the burden of providing JSON-editing support is placed on anyone embedding Luau, even if their environment is not naturally suited for JSON configuration.
+Ideally, embedding Luau should not require implementers to build or maintain specialized tooling for editing JSON-based configuration files.
+
+## Design
+
+We introduce a new configuration file called `.luauconfig`.
+
+### Syntax
+
+An example `.luauconfig` is given below.
+
+```luau
+return {
+    languageMode = "nonstrict",
+    lint = {
+        ["*"] = true,
+        LocalUnused = false
+    },
+    lintErrors = true,
+    typeErrors = true,
+    globals = {"expect"},
+    aliases = {
+        src = "./src"
+    }
+}
+```
+
+The file must return a Luau table of the following `LuauConfig` type:
+```luau
+type LanguageMode = "strict" | "nonstrict" | "nocheck"
+
+type LintWarning =
+    "*"
+    | "UnknownGlobal"
+    | "DeprecatedGlobal"
+    | "GlobalUsedAsLocal"
+    | "LocalShadow"
+    | "SameLineStatement"
+    | "MultiLineStatement"
+    | "LocalUnused"
+    | "FunctionUnused"
+    | "ImportUnused"
+    | "BuiltinGlobalWrite"
+    | "PlaceholderRead"
+    | "UnreachableCode"
+    | "UnknownType"
+    | "ForRange"
+    | "UnbalancedAssignment"
+    | "ImplicitReturn"
+    | "DuplicateLocal"
+    | "FormatString"
+    | "TableLiteral"
+    | "UninitializedLocal"
+    | "DuplicateFunction"
+    | "DeprecatedApi"
+    | "TableOperations"
+    | "DuplicateCondition"
+    | "MisleadingAndOr"
+    | "CommentDirective"
+    | "IntegerParsing"
+    | "ComparisonPrecedence"
+    | "RedundantNativeAttribute"
+
+type LuauConfig = {
+    languageMode: LanguageMode?,
+    lint: { [LintWarning]: boolean }?,
+    lintErrors: boolean?,
+    typeErrors: boolean?,
+    globals: { string }?,
+    aliases: { [string]: string }?,
+}
+```
+
+If configuration fields are not provided, `.luauconfig` uses the same default behavior as `.luaurc`.
+
+### Search semantics
+
+The search semantics of `.luauconfig` files are the same as `.luaurc`'s:
+
+> For a given .lua file, Luau will search for .luaurc files starting from the folder that the .lua file is in; all files in the ancestry chain will be parsed and their configuration applied.
+> When multiple files are used, the file closer to the .lua file overrides the settings.
+
+However, if both `.luaurc` and `.luauconfig` files are present in a directory, only the `.luauconfig` file is used.
+Configuration files are not merged; at most one configuration file is recognized per directory.
+
+### Configuration extraction
+
+Unlike `.luaurc` files, `.luauconfig` files can contain runtime constructs like variables, functions, and loops.
+To extract configuration from a `.luauconfig` file, its contents are executed in an isolated Luau VM with only the standard libraries enabled (`luaL_openlibs`).
+This approach ensures that configuration files can leverage Luau features while maintaining a sandboxed execution environment.
+
+As with the [user-defined type functions RFC](https://rfcs.luau.org/user-defined-type-functions.html), a time limit is imposed on the execution of `.luauconfig` files to guard against excessive computation.
+For example, executing the following configuration terminates after exceeding the allowed execution time, and an error is thrown during configuration parsing:
+```luau
+-- .luauconfig with an infinite loop
+return (function()
+    while true do
+    end
+end)()
+```
+
+This time limit is set to two seconds and can be revisited based on real-world usage and feedback.
+This duration is generally sufficient for typical configuration logic, while being short enough to prevent accidental or malicious long-running scripts from impacting performance.
+
+## Drawbacks
+
+**Complexity**: Allowing arbitrary Luau code in configuration files increases the complexity of parsing and validating configurations, which might make debugging more difficult.
+
+**Multiple formats**: There might be some confusion surrounding the two different configuration files.
+Most of this can be mitigated by (1) editor tooling, and (2) embedded contexts will likely choose one to support, not both.
+
+## Alternatives
+
+**Restrict syntax**: Limit `.luauconfig` files to only contain table literals and forbid functions, loops, and other runtime constructs to reduce complexity.
+Instead of spawning a new Luau VM to "parse" the file, we could simply spin up a Luau parser.
+
+**Improve support for `.luaurc` files**: We already provide an autocomplete API for Luau-syntax files and could expose an API that facilitates editing `.luaurc` files.


docs/config-luauconfig.md
```diff
@@ -0,0 +1,129 @@
...
+However, if both `.luaurc` and `.luauconfig` files are present in a directory, only the `.luauconfig` file is used.
+Configuration files are not merged; at most one configuration file is recognized per directory.
+
+### Configuration extraction
+
+Unlike `.luaurc` files, `.luauconfig` files can contain runtime constructs like variables, functions, and loops.
+To extract configuration from a `.luauconfig` file, its contents are executed in an isolated Luau VM with only the standard libraries enabled (`luaL_openlibs`).
+This approach ensures that configuration files can leverage Luau features while maintaining a sandboxed execution environment.
+
+As with the [user-defined type functions RFC](https://rfcs.luau.org/user-defined-type-functions.html), a time limit is imposed on the execution of `.luauconfig` files to guard against excessive computation.
```
aatxe, line 104: User-defined type functions do not have a timeout. They will run indefinitely unless the embedding environment asks them to stop. This works for analysis-time-only things, but the config affects runtime semantics as well.

docs/config-luauconfig.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+If configuration fields are not provided, `.luauconfig` uses the same default behavior as `.luaurc`.
+
+### Search semantics
+
+The search semantics of `.luauconfig` files are the same as `.luaurc`'s:
+
+> For a given .lua file, Luau will search for .luaurc files starting from the folder that the .lua file is in; all files in the ancestry chain will be parsed and their configuration applied.
+> When multiple files are used, the file closer to the .lua file overrides the settings.
```
vegorov-rbx: .luau or both

docs/config-luauconfig.md
vrn-sn: I forgot about the distinction between analysis-time vs. runtime here, but I will clean this section up after we iron out these details.

On the analysis side, we can just use our timeout logic like UDTFs do. On the runtime side, I currently have a section about the two-second default timeout, but I wonder if this should be more configurable. Maybe someone linking against `Luau.Config` defines a `CONFIG_RUNTIME_TIMEOUT` macro to set their own default (similar to the vector library's `LUA_VECTOR_SIZE`)?

Another option is a new hot comment at the top of `.luauconfig` files like `--!runtime-timeout=2` since that can be resolved at parse-time.

docs/config-luauconfig.md
vrn-sn: This is just quoted from the old RFC, which is slightly outdated—cleaning it up now.

docs/config-luauconfig.md
vrn-sn: Actually, the hot comment is a bad idea. I think we'd only want the embedding environment to be able to configure the timeout in this case, not the `.luauconfig` itself. That could be set at configure-time as a macro but also open to hearing alternatives.

vrn-sn:
I've renamed `.luauconfig` to `.config.luau` in this RFC, and I've now split the section about the timeout into analysis-time and runtime sub-sections.

For runtime configuration, I've introduced a new `LUAUCONFIG_RUNTIME_TIMEOUT_SECONDS` macro to override the default runtime timeout of two seconds.

docs/config-luauconfig.md
```diff
@@ -0,0 +1,138 @@
...
+-- .config.luau with an infinite loop
+return (function()
+    while true do
+    end
+end)()
+```
+
+For analysis-time configuration extraction, we adopt the same behavior described in the [user-defined type functions RFC](https://rfcs.luau.org/user-defined-type-functions.html): we simply respect the time limit that the embedding context provides through the `Frontend` API.
+
+At runtime, the configuration extraction timeout is set to two seconds by default.
```
JohnnyMorganz: Is it 2 seconds per config file (my current understanding), or 2 seconds overall for config resolution? Given that the directory hierarchy matters and we'll potentially be extracting a config from every ancestor directory.

vrn-sn: I'm kind of going back and forth. My original intention was per-file, but now am unsure about a design that causes a linearly scaling hang time wrt. virtual filesystem depth.

At runtime, there's only one use case, which looks like this with a cumulative timeout: every time we see an alias in `require`, we attempt to resolve it by executing each VFS node's `.config.luau` (going up the ancestry chain as needed). If we reach the timeout limit while executing a `.config.luau` in the chain, we throw a timeout error.

That seems okay to me. Ultimately, the timeout is a protective feature—I'd expect that extracting information from `.config.luau` would be trivial in almost all cases.

docs/config-luauconfig.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Search semantics
+
+The search semantics of `.config.luau` files are the same as `.luaurc`'s (adapted from old RFC):
+
+> For a given `.luau` or `.lua` script, Luau will search for `.luaurc` files starting from the folder that the script is in; all files in the ancestry chain will be parsed and their configuration applied.
+> When multiple configuration files exist throughout the ancestry chain, configurations closer to the script override those in higher-level directories.
+
+However, if both `.luaurc` and `.config.luau` are present in a directory, only the `.config.luau` file is used.
+Configuration files are not merged; at most one configuration file is recognized per directory.
```
jackhexed: I think that to remain consistent with other similar cases with `require`, this should be ambiguous and error.

docs/config-luauconfig.md
```diff
@@ -0,0 +1,138 @@
...
+## Design
+
+We introduce a new configuration file called `.config.luau`.
+
+### Syntax
+
+An example `.config.luau` is given below.
+
+```luau
+return {
```
alexmccord, line 25: You could add a special case in Frontend to add a constraint on the module's return type to be `LuauConfig` like below. This enables autocomplete for free.

docs/config-luauconfig.md
```diff
@@ -0,0 +1,138 @@
...
+
+**Multiple formats**: There might be some confusion surrounding the two different configuration files.
+Most of this can be mitigated by (1) editor tooling, and (2) embedded contexts will likely choose one to support, not both.
+
+## Alternatives
+
+**Restrict syntax**: Limit `.config.luau` files to only contain table literals and forbid functions, loops, and other runtime constructs to reduce complexity.
+Instead of spawning a new Luau VM to "parse" the file, we could simply spin up a Luau parser.
+
+**Improve support for `.luaurc` files**: We already provide an autocomplete API for Luau-syntax files and could expose an API that facilitates editing `.luaurc` files.
```
alexmccord, line 146: Ditto.

vegorov-rbx:
> * Set the timeout to be _really_ low.

I think this is a bad idea. Small interrupts on a machine will cause builds to be flaky.

> * Allow for these files to be required at runtime.

This is best left to a separate proposal explaining the exact runtime semantics.
There's no need to overcomplicate the simple move from existing configuration files in this RFC.

docs/config-luauconfig.md
aatxe: Seems fine.

docs/config-luauconfig.md
```diff
@@ -0,0 +1,138 @@
...
+For more fine-grained control, however, embedders are free to override this at build-time by defining the `LUAUCONFIG_RUNTIME_TIMEOUT_SECONDS` macro.
+
+### Interaction with `require`
+
+Despite `.config.luau` being a `.luau` file, it is not recognized as a module by `require` at runtime.
+Concretely, the source file `foo.luau` cannot extract the contents of its sibling `.config.luau` file with the expression `require("./.config")`.
+
+## Drawbacks
+
+**Complexity**: Allowing arbitrary Luau code in configuration files increases the complexity of parsing and validating configurations, which might make debugging more difficult.
```
JohnnyMorganz, line 135: I think, somewhat adjacent to the "Complexity" point, another drawback is that this makes it more difficult for third-party tooling to process Luau configuration.

Now, any external tool that wants to read Luau configuration needs to support executing Luau code, essentially introducing a dependency on Luau itself when it might not've originally been needed.

If a tool wants to make automated changes / transformations to such a file, or attempt to process the contents statically in a half-baked way, at a minimum they need to depend on a Luau parser. And this assumes that users don't write anything too fancy in their configuration (e.g., assume that they return a simple static table)

Contrast that to configuration stored in JSON where the config is static and has a consistent structure, and JSON parsers are commonplace, making processing a lot simpler.

docs/config-luauconfig.md
gaymeowing, line 25: Could returns for config.luau be made implicit? Allowing for simple config.luau files to just be
```luau
{
     languagemode = "nonstrict"
}
```

vrn-sn: We had been in two camps for this design when we were discussing internally:
1. A static table (no runtime constructs allowed) that depends only on the Luau parser.
2. A dynamically created table that depends on the Luau VM to extract the result.

Option 2 is the state of the current proposal, so we wanted to keep the syntax the same as Luau modules that explicitly return a table. Option 1 is a subset of Option 2 though, so I do get that syntactic sugar for that case might be nice regardless.

vrn-sn: Sounds good. I haven't explicitly added this to the RFC, but I did add a note about type checking and autocomplete facilitating writing these files.

alexmccord: Option 2 allows for nondeterminism, which is something I categorically hate the most in the context of configurations.

aatxe:
Overall, I think there are two big interests in using a Luau script as configuration for Luau, rather than some alternative Luau-derived syntax like a JSON-style equivalent.

The first is that we want to be able to inherit support for editing and interacting with these files from all of the already-existing work folks have done on building a good editor experience for Luau (including our work in Roblox, and @JohnnyMorganz's work in luau-lsp). One of the design principles for Luau is that the language should be small, but expressive. In this case, we'd be keeping the language small, despite supporting very powerful configuration, by just reusing the whole language. Since we operate with a small team, being able to leverage all of the work we've already done for the language to accomplish this is an incredible boon.

The second is that we are familiar with the large history of configuration languages in developer tooling broadly, and the impact that having non-programming languages for configuration has had. It is a very common occurrence that people have written [blog posts](https://solutionspace.blog/2021/12/04/every-simple-language-will-eventually-end-up-turing-complete/) about, and that most folks have probably seen in their workplace: if you go to write a configuration language of some kind, you'll suddenly find a need for bindings and abstractions, and if you have bindings and abstractions, you'll find reasons to want to have them depend on one another recursively. Many software systems I've worked on have had this problem (and likewise for others on the team), evolving extremely janky pseudo-programming languages that have all the features of a language, with none of the nice parts. By using Luau for the configuration, we skip this journey and settle at a language that is already the central focus of our design, and the outcome should be a much, much better configuration experience.

It definitely comes at a cost, since it means that tooling that wants to interact with require resolution will need to evaluate Luau as well. This is not free, but I think that it's worth it since it saves us from having to build out extensive tooling for some configuration format in order to support a good quality developer experience in every environment we care about as a team.

docs/config-luauconfig.md
vrn-sn: Going to go with 2 seconds per config. Should be pretty rare to hit this, but that way, whether or not parsing an individual config fails doesn't depend on other configs in the virtual filesystem.

docs/config-luauconfig.md
vrn-sn: Resolving this, I added a note to drawbacks, but I think we've agreed that it's an acceptable tradeoff.

docs/config-luauconfig.md
vrn-sn: Resolving this comment. There are definitely costs associated with Option 2, but @aatxe's [comment](https://github.com/luau-lang/rfcs/pull/125#issuecomment-3354050807) does a good job of explaining why we believe this is acceptable.

I'm going to leave the shortened syntax out of this proposal, but it could be proposed in a separate RFC if desired since this is forwards-compatible with it.

docs/config-luauconfig.md
```diff
@@ -0,0 +1,144 @@
...
+    while true do
+    end
+end)()
+```
+
+For analysis-time configuration extraction, we adopt the same behavior described in the [user-defined type functions RFC](https://rfcs.luau.org/user-defined-type-functions.html): we simply respect the time limit that the embedding context provides through the `Frontend` API.
+
+At runtime, the configuration extraction timeout for each `.config.luau` file is set to two seconds by default.
+This duration is generally sufficient for typical configuration logic while being short enough to prevent accidental or malicious long-running scripts from impacting performance.
+For more fine-grained control, however, embedders will be free to override this.
```
aatxe, line 125:
```suggestion
For more fine-grained control, however, embedders will be free to override this. That is to say, we do not wish to prescribe that this limit in general, but note that a limit being present in general is important.
```

Maybe a bit too legalistic, but I sort of want to explicitly say that we don't consider it wrong to have a different limit, and we're really just saying a limit must be present at runtime.

vrn-sn: Added.

andyfriesen:
FYI, it is looking fairly likely that we will be implementing some form of the `export` feature described here: https://github.com/luau-lang/rfcs/pull/42

Unwrapping the outermost `luau` keyword might unlock a nice opportunity to use the two together.

```luau
export languagemode = "nonstrict"
export lint = { ... }
export linterrors = true
...
```

vrn-sn:
Before we finalize this RFC, I want to get people's thoughts about @andyfriesen's comment above. In a world where we have an `export` keyword, do we still want to namespace configuration fields?

`.config.luau` with namespacing:
```luau
export luau = {
    languagemode = "nonstrict",
    lint = {
        ["*"] = true,
        LocalUnused = false
    },
    linterrors = true,
    typeerrors = true,
    globals = {"expect"},
    aliases = {
        src = "./src"
    }
}

export othernamespace = {
    foo = "bar"
}
```

`.config.luau` with flattened configuration:
```luau
-- "Luau" configuration
export languagemode = "nonstrict",
export lint = {
    ["*"] = true,
    LocalUnused = false
},
export linterrors = true,
export typeerrors = true,
export globals = {"expect"},
export aliases = {
    src = "./src"
}

-- Other configuration
export othernamespace_foo = "bar"
```

vrn-sn:
After some discussion in the OSS Discord, let's keep making a decision about whether `export` should be valid syntax in `.config.luau` files out of scope for this RFC. We can decide on that either in the `export` RFC or in a follow-up RFC. I think the main takeaway for us is that we want to keep Luau config under a `luau` field.

PR #126, opened 2025-06-17
RFC: `RedundantCast` lint rule

Shiawaseu:
[Rendered](https://github.com/Shiawaseu/rfcs/blob/master/docs/lint-redundant-cast.md)


SeloTheGreat:
as seen in https://github.com/luau-lang/luau/pull/2184#issuecomment-3746861695

> > this should be documented on a rfc and approved first before creating a new linter rule i think
>
> The linter is not part of the language itself, and frankly will probably be sunset in favor of https://github.com/luau-lang/lute/pull/597 we're writing [in lute](https://github.com/luau-lang/lute) since that allows programmers to author their own lints in luau instead of writing C++ to lint luau. Even if it continues to exist as well, it's not something we've treated as requiring an RFC.

*Apparently*, lint rules do not require RFCs, because lints aren't apart of the language itself. However, i'm not super sure if the team have changed their minds on this since that comment was made, but i'd assume not...?

PR #127, opened 2025-06-29
RFC: Indexing syntax when using long strings

Bottersnike:
[Rendered](https://github.com/Bottersnike/luau-rfcs/blob/indexing-syntax-long-strings/docs/indexing-syntax-long-strings.md)

Defines the logic for how to parse `foo[[[a]]]`, opting to parse this as `foo["a"]` rather than the current `foo"[a"]`.

This logic can be implemented in the official Luau lexer via the following modification to https://github.com/luau-lang/luau/blob/master/Ast/src/Lexer.cpp:

```diff
index c2743640..d311758c 100644
--- a/Ast/src/Lexer.cpp
+++ b/Ast/src/Lexer.cpp
@@ -744,6 +744,11 @@ Lexeme Lexer::readNext()

     case '[':
     {
+        if (peekch(1) == '[' && (peekch(2) == '[' || peekch(2) == '=')) {
+            consume();
+            return Lexeme(Location(start, 1), '[');
+        }
+
         int sep = skipLongSeparator();
```

While not mentioned in the document, as it is focused on Luau, it is interesting to note that most Lua minifiers stumble when trying to minify statements of the form discussed, removing this semantic whitespace and producing code that then parses differently. Lua currently still uses the semantic-whitespace behaviour that Luau is also currently using. (I tested `mathiasbynens/luamin`, `stravant/lua-minify` and `stravant/LuaMinify`)

### DIFF
--- a/docs/indexing-syntax-long-strings.md
+++ b/docs/indexing-syntax-long-strings.md
@@ -40,6 +40,8 @@ All current Luau parser implementations do not follow this behaviour. They inste
 
 Under the current behaviour `foo[bar[[[a]]]` parses as `foo[bar"[a"]`, however this proposed behaviour would parse that as `foo[bar["a"]` with an unterminated bracket pair. This is not backwards compatible, however the author of this RFC believes this to be a more expected behaviour, and could not find evidence of this problematic pattern being used in the wild.
 
+Under the current behaviour `print([[[[a]])` parses as `print("[[a")`. The change proposed here would cause this to parse as `print(["[a")` which is now invalid code. A single use of this was found on GitHub in a Lua file used for a custom nvim init.
+
 ## Alternatives
 There are two alternatives:
 
--- a/docs/indexing-syntax-long-strings.md
+++ b/docs/indexing-syntax-long-strings.md
@@ -43,9 +43,12 @@ Under the current behaviour `foo[bar[[[a]]]` parses as `foo[bar"[a"]`, however t
 Under the current behaviour `print([[[[a]])` parses as `print("[[a")`. The change proposed here would cause this to parse as `print(["[a")` which is now invalid code. A single use of this was found on GitHub in a Lua file used for a custom nvim init.
 
 ## Alternatives
-There are two alternatives:
+There are three alternatives:
 
 1. Do absolutely nothing. This is undesirable, as this behaviour remains ambiguous
 2. Retain the current behaviour, defining it explicitly with another RFC
+3. Define `[[[` and `[[[=...=[` as syntax errors.
 
-Of these, (2.) would be preferable if the proposed behaviour in this RFC is rejected.
+(3.) may be the "cleanest" of all the solutions here as it resolves the question of how this should be based, however it still retains the semantic whitespace behaviour.
+
+(1.) is heavily undesirable, and (2.) would be preferable instead.
--- a/docs/indexing-syntax-long-strings.md
+++ b/docs/indexing-syntax-long-strings.md
@@ -1,0 +1,49 @@
+# Indexing syntax when using long strings
+
+## Summary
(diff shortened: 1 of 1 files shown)


Bottersnike:
As a side note, this is totally ambiguous in Lua[u], even if deciding between long brackets vs indexing was performed during the parse stage using an advanced parser. The expression `foo[[[[a]]]` could be parsed as `'foo' '[' '[[' [a ']]' ']'`->`foo["[a"]` or also as `'foo' '[[' '[[a]' ']]'`->`foo("[[a]")`. The latter case is the existing behaviour as a side-effect of greedy parsing.

`lua-l` has some discussion of this issue at http://lua-users.org/lists/lua-l/2015-10/msg00171.html, though there doesn't seem to be any real conclusion drawn in that thread beyond some joking around.

There's a third alternative I forgot, that I'm adding into the RFC now.

jackdotink:
I don't think this passes the -200 points test.

deviaze:
I support this RFC because, as of right now

```luau
local meowkey = {
    [[[meow]]] = 2, -- syntax error
}

local meowkey2 = {
    [ [[meow]] = 2, -- completely fine
}
```

Luau shouldn't be relying on load-bearing whitespace here, so fixing this would make Luau a more consistent language and thereby (in my view) meets the -whatever point requirement.

Nit, but I thought `[[text \n \n]]` strings were called raw strings (even though they're often used as multiline strings)

jackdotink:
I think RFC is very similar to the trailing commas in function calls RFC. That RFC was rejected because while it made users think they were happier, it didn't actually make the language better or actually make users happier. This RFC is the same.

alexmccord:
There's a little bit of an ambiguity here, isn't there? It makes sense to want `{ [[[a]]] = 5 }` to be a table with some key `"[a]" = 5`, but what about when you use it outside of table keys? `local a = [[[a]]]`? That argues for lexing `[[[a]]]` as `"[a]"`, which is mutually incompatible with the table parsing as proposed.

docs/indexing-syntax-long-strings.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+This RFC does not address this second case, as the solution to this case is far less trivial.
+
+### Existing behaviour across tooling
+The predominant tool used for minification of Luau code at the time of writing, [darklua](https://github.com/seaofvoices/darklua), does not fall into the trap of minifying `foo[ [[a]]]` into `foo[[[a]]]`. It instead minifies to `foo[ [[a]] ]` or `foo['a']` depending on the configuration. Darklua would be able to remove those spaces after this change.
+
+The first party Luau parser follows the semantic-whitespace behaviour.
+
+The predominant Rust parser for Luau, [full-moon](https://github.com/Kampfkarren/full-moon), follows the semantic-whitespace behaviour.
```
alexmccord, line 31: These aren't really semantic, since it only affects the parser step and it's a parse error or not which makes this entirely syntactic. I'd use "significant whitespace" instead (although that term is traditionally used to decide which block to place a statement in, I think it still applies)

Bottersnike:
That's a really good point. I think the "logical" parse of `a = [[[a]]]` would be `a = "[a]"` meanwhile `{[[[a]]] = ...}` as `{["a"] = ...}`.

Is there a good way to solve this? Quite likely not. The lexer could do some pretty nasty checks to decide when to perform which behaviour (ie if the previous lexeme was a name, `]`, `)` or `}`, let `[` take precedence over strings) though this still has utter ambiguity with `foo [[[a]]]` which could either be a call or an index.

The simplest solution here might be to consider this just a general poor design decision of lua, unfixable, and then codify the current behaviour in writing rather than seeking to amend it at all?

PR #128, opened 2025-07-06
noinline attribute

jackhexed:
This RFC adds the `@[inline("never")]` attribute to replace the `local func; function func()` pattern to prevent inlining.

[Rendered](https://github.com/jackdotink/rfcs/blob/noinline-attribute/docs/syntax-attribute-functions-inline-never.md)

### DIFF
--- a/docs/syntax-attribute-functions-inline-never.md
+++ b/docs/syntax-attribute-functions-inline-never.md
@@ -1,0 +1,43 @@
+# Inline Attribute for Functions
+
+## Summary
+
+This RFC proposes a `@[inline("never")]` function attribute. This attribute's presence will restrict the compiler from inlining a function.
+
+## Motivation
+
+When optimizing Luau code it is common to place cold-path code into a function that cannot be inlined to reduce both bytecode size and cpu time usage.
+
+```luau
+local function cold_path()
+	clean_up_operation()
+	error(some_error_value())
+end
+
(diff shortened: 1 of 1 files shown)


Bottersnike:
Given #49 strikes down the idea of a `@inline` is the motivation against `@noinline` valid? Personally I think it would make far more sense over a mandatory argument that's only ever a single value (especially given the syntax for arguments to attributes isn't the prettiest too).

Dekkonot:
I'm inclined to boost `@noinline` over `@inline(never)` for the reasons Bottersnike mentioned above. If we ever get an `@inline` attribute, we can add `never` as an option without too much trouble.

jackdotink:
I am happy with either.

tommyscholly:
I was actually planning on revisiting this RFC, since we shipped a `@debugnoinline` (for debugging) and people at RDC expressed interest.

jackhexed:
I'll reopen in that case.

tommyscholly:
I would want this as `@noinline`, as we will never be able to guarantee anything besides *not* inlining, so having an inline attribute where its only parameter is `never` seems redundant.

jackhexed:
I can imagine a scenario with `@[inline("please")]` with a gentler cost function or no cost function. With Luau's compilation model a cost function will never be able to universally make better decisions than the programmer can.

PR #129, opened 2025-07-14
RFC: type externname - Method to get an ``ExternType``'s name

karl-police:
[Rendered](https://github.com/karl-police/rfcs/blob/patch-3/docs/udtf-type-extern-name.md)

I also made an implementation test of it, see a draft https://github.com/luau-lang/luau/pull/1916

### DIFF
--- a/docs/udtf-classname.md
+++ b/docs/udtf-classname.md
@@ -110,9 +110,10 @@ type a = onlyClassName<CustomClass>
 
 ## Drawbacks
 
-The implementation for ``type:name()`` seems to copy the Generic Name.
+I don't know if a generic can be directly passed into a type function and then use ``type:name()``.
+But Generic Names are being collected.
 
-If ``type:classname()`` would do this as well, maybe it would be catastrophic to memory.
+If ``type:classname()`` would do this as well, maybe it would be inefficient for memory, because the name already exists.
 In my first implementation, it doesn't do that though within the "serialize" process, instead it's just "read on request", but without caching.
 
 It is currently confusing that there's ``:value()`` but the function name itself, doesn't speak out _"Hey, I am only for singletons"_ or similar.
--- a/docs/udtf-classname.md
+++ b/docs/udtf-classname.md
@@ -1,0 +1,124 @@
+# type classname method utility
+
+## Summary
+
+Add a new method to ``type`` for User Defined Type Functions, ``type.classname``. Which will retrieve a ``ExternType``'s name.
+
+Or any other way to get the classname.
+
+## Motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+**Use cases:**
+- A lazy (but therefore also quick way?) to filter out class based types (e.g. from ``declare class``)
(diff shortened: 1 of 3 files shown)


karl-police:
> We will not accept any additions to the language that refer to extern types as classes. The choice of changing that language was very conscious to avoid confusion with the idea that language supported some kind of object-oriented programming. You'll need to adjust the RFC to use the right language, and not reference classes.

I changed it.

I didn't even name the C++ function classname, I actually named it ``getExternTypeName``. I sticked with "classes" probably because the I only saw ``declare class`` being a Extern Type.

deviaze:
Old references to "class" might just be fallout from the rename and aren't intentional to thn language design. At some point `declare class` syntax is supposed to be replaced with `declare extern type` syntax, but it doesn't seem to have been yet. Well, at least my `globals.d.luau` hasn't broken with luau-lsp yet.

karl-police:
> Old references to "class" might just be fallout from the rename and aren't intentional to the language's design. At some point `declare class` syntax is supposed to be replaced with `declare extern type` syntax, but it doesn't seem to have been yet. Well, at least my `globals.d.luau` hasn't broken with luau-lsp yet.

Ah. That would be better if it's renamed. I was getting confused about ``declare class`` thinking about _"Are we getting classes, huh?"_

But since it's not at the moment, I kept the RFC like so, but I guess I could tweak ``CustomClass`` and name it ``CustomExternType``

aatxe:
Yes, `declare extern type` has worked for a while now (several months?), and we're giving people (probably mostly @JohnnyMorganz if I'm being honest) time to switch things, but no RFCs were ever or will ever be accepted referring to extern types as classes.

karl-police:
> Yes, `declare extern type` has worked for a while now (several months?), and we're giving people (probably mostly @JohnnyMorganz if I'm being honest) time to switch things, but no RFCs were ever or will ever be accepted referring to extern types as classes.

I think ``is("class")`` is still unchanged though

karl-police:
> > I think `is("class")` is still unchanged though
> 
> Yeah, this one is unfortunate, and we should change it before it's more codified as well because, again, these are not "classes" and setting the expectation for Luau programmers that the language has a notion of "class types" right now is misleading at best, and actively harmful at worst.

Yes https://github.com/luau-lang/rfcs/pull/131

aatxe:
I'll clean up the flag now, so that you can do so in this next release. :)

karl-police:
Now, all ``"class"`` is gone.

SeloTheGreat:
Has there been any other drawbacks that have held this RFC from being merged? This seems like a rather useful feature.

...or has this feature been put into some sort of icebox?

aatxe:
> Has there been any other drawbacks that have held this RFC from being merged? This seems like a rather useful feature.

No other drawbacks, but the RFC text itself could do with a lot of polishing. We're also currently trying to sort through changes to our processes to make all of the open source work move along more smoothly than it has been historically.

karl-police:
I polished a bit

PR #130, opened 2025-07-17
RFC: Table Destructuring Syntax

cameronpcampbell:
[Rendered.](https://github.com/cameronpcampbell/rfcs/blob/master/docs/syntax-table-destructuring.md)

### DIFF
--- a/docs/syntax-table-destructuring.md
+++ b/docs/syntax-table-destructuring.md
@@ -26,7 +26,7 @@ local useState = React.useState
 
 With destructuring syntax this could be shortened to the following:
 ```lua
-local .{ useEffect, useMemo, useState } = React
+local .{ .useEffect, .useMemo, .useState } = React
 ```
 
 ## Design
--- a/docs/syntax-table-destructuring.md
+++ b/docs/syntax-table-destructuring.md
@@ -65,7 +65,7 @@ We propose nested key destructuring via the following syntax (This also works wi
 ## Array destructuring
 We propose the following syntax for array destructuring:
 ```lua
-.{ one, two, three } == thing
+.{ one, two, three } = thing
 -- one == thing[1]
 -- two == thing[2]
 -- three == thing[3]
@@ -73,13 +73,23 @@ We propose the following syntax for array destructuring:
 
 We propose the following syntax for nested array destructuring:
 ```lua
-.{ one, .{ apple }, three } == thing
+.{ one, .{ apple }, three } = thing
 -- one == thing[1]
 -- apple == thing[2][1]
 -- three == thing[3]
 ```
 
-## Rest destructuring
+## Mixed Destructuring
+Key and array destructuring can be used together:
+```lua
+.{ .foo, one, .bar as .{ .baz }, .{ apple } } = thing
+-- foo == thing.foo
+-- one == thing[1]
+-- baz == thing.bar.baz
+-- apple = thing[2][1]
+```
+
+## Rest Destructuring
 
 We propose destructuring the rest of the properties via the following syntax (`rest` can be any identifier). Only one rest parameter can be defined.
 ```lua
--- a/docs/syntax-table-destructuring.md
+++ b/docs/syntax-table-destructuring.md
@@ -111,7 +111,7 @@ type Props = {
     pear: any
 }
 
-function test(.{ apple, pear }: Props)
+function test(.{ .apple, .pear }: Props)
     ...
 end
 ```
--- a/docs/syntax-table-destructuring.md
+++ b/docs/syntax-table-destructuring.md
@@ -1,0 +1,156 @@
+# Table Destructuring Syntax
+
+
+
+## Summary
+Introduce new syntax for unpacking values from tables.
+
+```lua
+local &{ foo, ["@bar"] = bar } = thing
+-- foo == thing.foo
+-- bar == thing["@bar"]
+```
+
+This RFC addresses issues brought up from previous table destructuring RFC's.
+
+
+
+## Motivation
+Aliasing table items through variable assignments is a common pattern in luau, especially when working with large libraries. However doing this with a lot of items becomes repetitve:
+```lua
+local useEffect = React.useEffect
+local useMemo = React.useMemo
+local useState = React.useState
+```
+
+With destructuring syntax this could be shortened to the following:
+```lua
+local &{ useEffect, useMemo, useState } = React
+```
+
+## Design
+
+We propose that the table destructuring syntax is prefixed with an ampersand (`&`) symbol. Without a prefix destructuring becomes ambiguous in certain scenarios, such as the one below.
+```lua
+baz = foo
+{ bar } = test
+```
+*(This is ambiguous as it can be interpretted as `baz = foo { bar }; = test` or `baz = foo; { bar } = test`).*
+
+### Key destructuring
+We propose the following syntax for destructuring by keys:
+```lua
+&{ ["foo"] = foo } = thing
+-- foo == thing.foo
+```
+
+If the key can be expressed as a valid variable name then a shorthand can be used instead:
+```lua
+&{ foo } = thing
+```
+
+Shorthands can be destructured with a different name:
+```lua
+&{ foo = bar } = thing
+-- bar == thing.foo
+```
+
+We propose the following syntax for nested key destructuring:
+```lua
+&{ ["@foo"] = &{ bar } } = thing
+-- bar == thing["@foo"].bar
+```
+This also works with shorthands.
+
+### Array Destructuring
+We propose the following syntax for destructuring arrays:
+```lua
+&{{ one, two, three }} = thing
+-- one == thing[1]
+-- two == thing[2]
+-- three == thing[3]
+```
+
+You can omit a certain index via the `nil` keyword:
+```lua
+&{{ one, nil, three }} = thing
+-- one == thing[1]
+-- three == thing[3]
+```
+
+We propose the following syntax for nested array destructuring:
+```lua
+&{{ one, &{{ apple }}, three }} = thing
+-- one == thing[1]
+-- apple == thing[2][1]
+-- three == thing[3]
+```
+
+### Mixed Destructuring
+Key and array destructuring can be used together.
+
+```lua
+&{
+    ["@foo"] = foo,
+    { one, nil, three },
+} = thing
+-- foo == thing["@foo"]
+-- one == thing[1]
+-- three == thing[3]
+```
+
+### Local Destructuring
+We propose that if a destructure assignment is preceeded with `local` then all of the destructured variables will be local.
+```lua
+local &{ hello, world } = thing
+```
+
+### Function Argument Destructuring
+We propose that function arguments can be destructured:
+
+```lua
+type Props = {
+    apple: any,
+    pear: any
+}
+
+function test(&{ apple, pear }: Props)
+    ...
+end
+```
+*(The type annotation is not neccesary, its purely there for demonstration purposes).*
+
+### Type Destructuring.
+We propose special behaviour for destructing types from requires:
+```lua
+&{ type Foo } = require(...)
+```
+
+You can also precede a destructure assignment with `type`. This is particularly helpful for destructuring multiple types without having to repeat the `type` keyword, although it does have the caveat of only allowing type destructuring.
+```lua
+type &{ Foo, Bar, Baz } = require(...)
+```
+
+Types can also be destructured with a different name:
+```lua
+&{ type Foo = Biz } = require(...)
+-- type Biz == type require(...).Foo
+```
+
+
+## Drawbacks
+Prefixing table destructuring with a symbol may not fit into the language as historically luau has generally not used symbol prefixes. However there are some examples of symbol prefixes (especially in recent years):
+- length operator (`#thing`).
+- function attributes (`@native`).
+- negation operator (`~SomeType`).
+
+
+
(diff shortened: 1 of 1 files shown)


docs/syntax-table-destructuring.md
```diff
@@ -0,0 +1,156 @@
...
+-- two == thing[2]
+-- three == thing[3]
+```
+
+You can omit a certain index via the `nil` keyword:
+```lua
+&{{ one, nil, three }} = thing
+-- one == thing[1]
+-- three == thing[3]
+```
```
jackhexed: Instead of a `nil` keyword which is used in this way nowhere else in the language, users should use `_`.

docs/syntax-table-destructuring.md
```diff
@@ -0,0 +1,156 @@
...
+You can also precede a destructure assignment with `type`. This is particularly helpful for destructuring multiple types without having to repeat the `type` keyword, although it does have the caveat of only allowing type destructuring.
+```lua
+type &{ Foo, Bar, Baz } = require(...)
+```
+
+Types can also be destructured with a different name:
+```lua
+&{ type Foo = Biz } = require(...)
+-- type Biz == type require(...).Foo
+```
```
jackhexed: IMO leave this out for now and let another RFC cover this when associated table types are in the language

docs/syntax-table-destructuring.md
```diff
@@ -0,0 +1,156 @@
...
+{ bar } = test
+```
+*(This is ambiguous as it can be interpretted as `baz = foo { bar }; = test` or `baz = foo; { bar } = test`).*
+
+### Key destructuring
+We propose the following syntax for destructuring by keys:
+```lua
+&{ ["foo"] = foo } = thing
+-- foo == thing.foo
+```
```
jackhexed: I don't like that whats being assigned into is on the right side of the `=` when in every other case in the language it is on the left. Consider reversing this (`foo = ["foo"]`) or using the `in` keyword (`["foo"] in foo`).

docs/syntax-table-destructuring.md
cameronpcampbell: I initially wanted to propose that but `_` is a valid variable name.

docs/syntax-table-destructuring.md
cameronpcampbell: This is fair criticism, I have changed the proposal to use a `as` keyword instead of an `=` symbol for destructuring as a different name.

docs/syntax-table-destructuring.md
jackhexed: Yes, this is the point. When users wish to ignore a value, say when unpacking an array-like table, they use `_`:
```luau
local foo, _, bar = table.unpack(arr)
```
There is no reason for this to be a special case.

docs/syntax-table-destructuring.md
alexmccord: Renaming is orthogonal to the feature itself if it was a contentious point, but fwiw `as` is unambiguous both syntactically and directionally. I never remember which side of the colon renames the binding in `let Point { x: x1, y: y1 } = p1;`. +1 on `as`.

docs/syntax-table-destructuring.md
```diff
@@ -0,0 +1,137 @@
+# Table Destructuring Syntax
+
+
+
+## Summary
+Introduce new syntax for unpacking values from tables.
+
+```lua
+local &{ foo, ["@bar"] = bar } = thing
```
alexmccord: Not happy about the `&` tbh. If we _really_ had to pick a symbol, `.` would be better: `local .{ foo, ["@bar"] as bar } = thing` would be my preference. You could also envision `.{ x, y }` (table literal expression) as field-punning syntactic sugar for `{ foo = foo }` (orthogonal to the RFC, just mentioning an idea)

docs/syntax-table-destructuring.md
```diff
@@ -0,0 +1,137 @@
...
+
+Shorthands can also be destructured with a different name:
+```lua
+&{ foo as bar } = thing
+-- bar == thing.foo
+```
+
+We propose the following syntax for nested key destructuring:
+```lua
+&{ ["@foo"] = &{ bar } } = thing
```
alexmccord: Pattern matching is recursively defined, you can still use `as` here. `local &{ ["@foo"] as &{ bar } } = thing` to make `bar` be a projection of `thing["@foo"].bar`.

docs/syntax-table-destructuring.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+With destructuring syntax this could be shortened to the following:
+```lua
+local &{ useEffect, useMemo, useState } = React
+```
+
+## Design
+
+We propose that the table destructuring syntax is prefixed with an ampersand (`&`) symbol. Without a prefix destructuring becomes ambiguous in certain scenarios, such as the one below.
```
alexmccord: I do not think table destructuring should be allowed in assignments. Allowing this in assignments would introduce a footgun where you want to introduce a binding and assign a new value to some other binding.

Also, what would it even mean if you had this?
```luau
local foo

-- is `set` a no-op function?
-- or a function that captures `foo` and updates it?
function set(.{ foo }) end
```

It also isn't flexible enough to support assignments to fields (`.{ x, y as self.value } = t`??).

The way to think about it is that pattern matching is generally a readonly operation, and assignments certainly isn't one. I say generally, but there are counterexamples, like [mutable matching](https://counterexamples.org/mutable-matching.html) and in the case of Luau, metatables that does side effects, but alas. Even Rust has `DerefPure` trait because of interior mutability (`Rc<RefCell<T>>` or `Arc<Mutex<T>>`, etc).

docs/syntax-table-destructuring.md
```diff
@@ -0,0 +1,137 @@
...
+
+We propose the following syntax for nested key destructuring:
+```lua
+&{ ["@foo"] = &{ bar } } = thing
+-- bar == thing["@foo"].bar
+```
+This also works with shorthands.
+
+### Array Destructuring
+We propose the following syntax for destructuring arrays:
```
alexmccord: In this specific version of array destructuring, you could drop the `.` prefix in the suggestion I mentioned earlier. Let `local .{ x, y } = point` mean `local x, y = point.x, point.y` and `local { pt1, pt2 } = points` mean `local pt1, pt2 = points[1], points[2]`. 

Either mixed destructuring is seldom needed, or it is done in two lines (`local .{ x, y } = t` and `local { a, b } = t`, or spelling the field names with `as` (`local { a, b, x as x, y as y } = t`), or the same design like `local { .x, .y } = point` like the RFC #24 to discern the two.

docs/syntax-table-destructuring.md
```diff
@@ -0,0 +1,137 @@
...
+- function attributes (`@native`).
+- negation operator (`~SomeType`).
+
+## Alternatives
+Alternative ideas for table destructuring have been proposed before, most of which have been rejected:
+- https://github.com/luau-lang/rfcs/pull/24 (`local {.a, .b} = t`)
+- https://github.com/luau-lang/luau/pull/629 (`local { a, b } = t`)
+- https://github.com/luau-lang/rfcs/pull/95 (`local { a, b } = t`)
+
+The main reasons for their rejection include ambiguity issues, not handling arrays, and not handling non-local assignments.
```
alexmccord: Not sure the non-local assignments were used as a reason to reject the RFCs. In fact, I would have rejected the RFC that asked for it.

deviaze:
> This doesn't mention anything about whether table destructuring requires `typeof(value) == "table"` or the more slightly flexible `typeof(value) == "table" or typeof(value) == "userdata"`. If left unchecked, it technically allows `local .{ lower } = "gross"; print(lower("REALLY GROSS CODE"))`.
> 
> Constraining it to only work for tables and userdatas makes this simpler for type inference to do the correct thing too, so I'm in the camp of enforcing the constraint.

Not sure why we'd want to disallow that; you can already write
```luau
local gross = "Gross"
local lower = gross.lower
print(lower) --> gross
```
which works at runtime, typechecks fine, and doesn't throw any lints.

This would basically just be sugar for that with one fewer `local` binding.

Syntax wise, let's please not do the `&` prefix and `{{ binding }}` for array-likes; those look completely out of place and would only confuse people. In terms of minimizing confusion, I actually prefer `local .{ x, y } = point` over js' `const { x, y } = p` because it more clearly indicates indexing/property access.

cameronpcampbell:
I really appreciate everyone who has contributed to the conversation thus far. I've significantly overhauled this RFC based on the criticism.

docs/syntax-table-destructuring.md
```diff
@@ -0,0 +1,143 @@
...
+### Function Argument Destructuring
+We propose that function arguments can be destructured:
+
+```lua
+type Props = {
+    apple: any,
+    pear: any
+}
+
+function test(.{ apple, pear }: Props)
```
jkelaty-rbx:
```suggestion
function test(.{ .apple, .pear }: Props)
```

docs/syntax-table-destructuring.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Mixed Destructuring
+Key and array destructuring can be used together:
+```lua
+.{ .foo, one, .bar as .{ .baz }, .{ apple } } = thing
+-- foo == thing.foo
+-- one == thing[1]
+-- baz == thing.bar.baz
+-- apple = thing[2][1]
+```
```
vegorov-rbx, line 90: Proposal seems to lack the detail on exact operation sequencing, especially with relation to the metamethods.

Does the implementation has flexibility and/or unspecified ordering of the side-effects? Each index operation on `thing` can change the contents of `thing`.
Or will there be a strict set of steps for follow (will this prevent ability to do common sub-expression elimination for multiple nested unpacks?).

If dynamic keys are used with [expr], when are they evaluated?
And what happens if one of the expressions changes the value of `thing`?

Consider in part how https://www.lua.org/manual/5.4/manual.html#3.3.3 goes into the ordering and hidden side-effect interaction.

docs/syntax-table-destructuring.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Rest Destructuring
+
+We propose destructuring the rest of the properties via the following syntax (`rest` can be any identifier). Only one rest parameter can be defined.
+```lua
+.{ ["@foo"] as .{ .bar }, ...rest } = thing
+```
```
vegorov-rbx, line 97: Please explain the effect of this statement and how it will be implemented.

docs/syntax-table-destructuring.md
alexmccord: In my opinion, it should be in the same order as listed syntactically, depth first. Evaluate the side effects of `.foo`, then `[1]`, then `.bar.baz`, then `[2][1]`. But I _strongly_ dislike this specific construct where it'll perform strong updates on lvalues.

For example, `local .{ x as .{ y }, x as .{ z } } = t` is `local y, z = t.x.y, t.x.z` evaluates the side effects of `t.x` twice even though it could be coalesced as `local .{ x as .{ y, z } } = t` which would require a temporary `_x` which would evaluate the side effects of `t.x` only once.

docs/syntax-table-destructuring.md
```diff
@@ -0,0 +1,143 @@
+# Table Destructuring Syntax
+
+
+
+## Summary
+Introduce new syntax for unpacking values from tables.
+
+```lua
+local .{ .foo, one, ["@bar"] as bar } = thing
```
alexmccord, line 9: Wanted to note that if we're going to have `.foo`, the outer `.` is actually superfluous. It's like the `://` in URIs, which only needed `:` and makes `//` superfluous. Once removed, this ends up being the same exact design as https://github.com/luau-lang/rfcs/pull/24.

alexmccord: I guess with the exception of being able to perform assignments to locals/globals (not lvalues in general), but I've already said I am against that.

docs/syntax-table-destructuring.md
alexmccord: Doesn't look implementable to me. How can this work with `userdata`? You can't iterate over keys in `__index`, or remove a key from `__index`, not to mention that `table.clone` doesn't work if the table's metatable has `__metatable`, so it follows that `local { ...clone } = t` must follow the same rules as `table.clone`. And should the `rest` table be frozen if `thing` is also frozen? And what happens if there is an `__iter` metamethod?

docs/syntax-table-destructuring.md
vegorov-rbx: A fine option as long as it's specified that way.
Following strong program order might sound intuitive, but it's not how every statement is defined to be evaluated in Luau.

Following intuitive program order can still cause surprises though.

```luau
t = { x = { a = 4 }, a = 10 }
.{ .x as t, .a } = t
```

does this evaluate to
```luau
t = t.x -- { a = 4 }
a = t.a -- 4
```
in order to satisfy 'evaluate the side effects .x then .a'.

or is it actually still a multi-assignment, evaluating to:
```luau
t, a = t.x, t.a -- t == { a = 4 }, a == 10
```

Maybe by side-effects you only meant the 'rhs' parts without performing the assignment in which case the latter example will be correct. Would be nice to say that all reads complete before assignment is performed in that case.

cameronpcampbell:
I will be closing this PR to reflect on the feedback that was given and hopefully reopen it once the RFC is more robust. Thanks for everyone who helped contribute.

PR #131, opened 2025-07-22
RFC: Rename tag "class" to "extern" for ExternType

karl-police:
[Rendered](https://github.com/karl-police/rfcs/blob/patch-4/docs/udtf-is-extern.md)

~~``is("class")``~~
``ìs("extern")``

### DIFF
--- a/docs/udtf-is-extern.md
+++ b/docs/udtf-is-extern.md
@@ -1,19 +1,21 @@
-# tag name for extern types
+# `extern` tag in User-Defined Type Functions.
 
 ## Summary
 
-Currently, the ``tag`` for an ExternType is ``class``. This RFC suggests to change it to ``extern``.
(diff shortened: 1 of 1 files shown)


karl-police:
> Sounds good to me!

I made https://github.com/luau-lang/luau/pull/1926 and https://github.com/luau-lang/luau/pull/1927

two variations

karl-police:
It turns out that ``ìs("class")`` is actually not used anywhere in the Unit Tests in the first place as well. I could remove ``Debug`` infront of the flag and let it turn on through ``--fflags=true``, and allow that all current ``is("class")`` will just error.
Unless it's worth to wait and implement a warning that points out to use ``"extern"``

aatxe:
It's still a feature in the beta, and not usable elsewhere, and it will never break running code, so I think it's fine for us to not try to manage an awkward "warn to migrate" period (which is, intrinsically, the same as producing the error _anyway_).

PR #133, opened 2025-08-04
RFC: Octal integer literals

man-of-eel:
[Rendered](https://github.com/man-of-eel/rfcs/blob/master/docs/syntax-octal-literals.md)

[Implementation PR](https://github.com/luau-lang/luau/pull/1943)

### DIFF
--- a/docs/syntax-octal-literals.md
+++ b/docs/syntax-octal-literals.md
@@ -1,0 +1,26 @@
+# Python-style octal literals
+
+## Summary
+
+This RFC proposes allowing octal literals to be defined in Luau in the form `0oNNN...` - for example, `0o777` evaluates to 511 in decimal.
+
+## Motivation
+
+Many other languages also support octal literals, so it allows for easier porting of code. 
+Some legacy systems use octal extensively, and converting from octal to other bases to interface with them can be a chore.
+
+## Design
+
+Octal literals can be specified in Luau code with an `0o` prefix, just like how binary and hexadecimal literals use `0b` and `0x` respectively.
+
+Identifiers beginning with `0o` are already illegal, so this should not impact any existing code.
+
+## Drawbacks
+
+Since they use only decimal digits, octal literals may appear confusing at first to other people reading code using them.
+
+## Alternatives
+
+Simply using other bases will of course work, however it can be time-consuming and in some cases there are meaningful advantages to using octal (e.g. PDP-11 instruction encodings).
+
+Some languages use a leading zero to denote octal literals (most notably C), but this would break lots of existing code and is generally disliked by most programmers.
--- a/docs/syntax-octal-literals.md
+++ b/docs/syntax-octal-literals.md
@@ -1,26 +1,33 @@
-# Python-style octal literals
+# Octal integer literals
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


Bottersnike:
The main motivation for octals, imo, is when doing any interaction with a Linux filesystem, as they use octals for file permissions. Could be worth mentioning in the RFC?

man-of-eel:
Python 3 was the first language to introduce the `0o` syntax AFAICT and putting it there makes it clear it's not C-style octal literals. And not sure what you mean by turn people away... who exactly?

wrt. Unix file permissions I don't think there's much convincing anyone that this isn't a very niche feature. I was considering just opening a PR to the Luau repo since the actual change is just 3 lines (everything else is added warnings/tests) but since it says to open an RFC for feature additions I thought I should listen.  Maybe I'll just make it anyway and see

aatxe:
I dunno about turning people away, but I do not think the title of an RFC should ever say "PopularLanguage-style." It feels weird. You're proposing the feature for Luau, it has nothing to do with Python in any way beyond maybe you having been inspired by it in Python. We're not going to accept e.g. the idea that whatever Python changes with this feature in the future, we'll also change in precisely the same way.

This is indeed a language feature addition though, and it does indeed require an RFC. I would just frame it, especially in title, as just the feature you're proposing (Octal integer literals), and discuss any inspiration for the design today in the body of the RFC.

aatxe:
Yeah, no worries, the actual feature is fine afaik, just need the doc describing it to be more polished since it exists in perpetuity.

docs/syntax-octal-literals.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Motivation
+
+Some legacy systems use octal extensively, and converting from octal to other bases to interface with them can be a chore. For example, the PDP-11 uses octal extensively - its instruction encodings are a natural fit for octal as they are arranged in groups of 3 bits, and octal is used to represent addresses & words in nearly all materials.
+
+Unix also uses octal to represent file permissions (e.g. in commands like `chmod 755`), and this is perhaps the most recognisable use of it today.
+
+Many other languages (Python, JavaScript, Rust...) also support octal literals in this style, so it allows for more straightforward porting of code.
+
+The audience for this feature is admittedly small, but the parser provisions for supporting hexadecimal and binary literals extend easily onto adding octal literals.
```
Dekkonot, line 15: I don't know if this is actually a motivation, it's just a description of why you'd want to use octals in general. It doesn't answer why Luau should have them beyond "it makes porting code easier".

man-of-eel: I suppose you're right, it is pretty much just a list of octal facts.. I'm not sure what would be better to put there, though. The real story is just I was writing a PDP-11 emulator and got annoyed at having to constantly have an octal converter open lol. Went to see how easy it would be to implement (it's basically 3 lines) but wasn't really ready for all the red tape

andyfriesen:
Would it be a bridge to far to expressly disallow a capital `O` to indicate an octal literal?  `0O777` should be a syntax error.

vegorov-rbx:
0Bddd is not accepted today for binary literals, so I would expect the same from octal.
While RFC doesn't include a mention of 0Oddd, a note could be added that it's not valid explicitly.

Edit: I have rechecked and this is actually incorrect, 0B and 0X are accepted today, which will make 0O an exception.

PR #134, opened 2025-08-14
RFC: Table Dictionary Length

krakow10:
[Rendered](https://github.com/krakow10/rfcs/blob/dictlen/docs/table-dictionary-length.md)

Add a function `table.dictlen(t)` to the standard library which returns the size of the dictionary part of the given table.

Some alternative function name suggestions:
- `table.dictsize(t)`
- `table.nkeys(t)`

Maybe there's an even better name for the function.

### DIFF
--- a/docs/table-dictionary-length.md
+++ b/docs/table-dictionary-length.md
@@ -1,0 +1,33 @@
+# Table Dictionary Length
+
+## Summary
+
+A built-in function to get the length of the dictionary part of a table.
+
+## Motivation
+
+An O(1) method to count the number of elements in the dictionary part of the table by lifting the information directly from the underlying data structure.  Without this, it's only possible to get the count by premeditated management of a count variable or by outright counting on the spot.
+
+## Design
+
+Add a function `table.dictlen(t)` to the standard library which returns the size of the dictionary part of the given table.
+
+## Drawbacks
+
+If Luau's C++ Lua table implementation has no concept of its current length and the algorithmic complexity of the operation is no better than outright counting, then there's no reason to add this.
+
+## Alternatives
+
+Maintaining a count variable manually via proxy functions.  The implementation below has some caveats: it only counts correctly when inserting an element that hasn't already been inserted or when removing an element that hasn't already been removed.
+```Lua
+local t={}
+local t_count=0
+local function t_insert(i,v)
+	t[i]=v
+	t_count+=1
+end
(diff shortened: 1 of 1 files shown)


aatxe:
`table.count` is a much nicer name.

docs/table-dictionary-length.md
```diff
@@ -0,0 +1,33 @@
+# Table Dictionary Length
+
+## Summary
+
+A built-in function to get the length of the dictionary part of a table.
+
+## Motivation
+
+An O(1) method to count the number of elements in the dictionary part of the table by lifting the information directly from the underlying data structure.  Without this, it's only possible to get the count by premeditated management of a count variable or by outright counting on the spot.
```
alexmccord, line 9: I would like to see a motivation that isn't abstract. Right now, I could just tell you to write something with the type `{{k: string, v: unknown}}` and it would be O(1) and works with `nil`. Sometimes you have to accept that your data structure is inherently richer than a primitive. Or if you need O(1) access, `{keys: {string}, data: {[string]: unknown}}` which enables you to know what keys are missing. Or you could add an abstraction like `Option.t<T>` and get either one for free: `{[string]: Option.t<T>}`.

Really, this is a symptom of `nil` being used to represent optionality _and_ key removal, when optionality is nested, for instance `Option<T> != Option<Option<T>>`, but Lua only afforded the former.

krakow10:
Thanks for the explanations and function name suggestions.  If the size of a table has to be increased to accommodate this feature, then it's not worth it for what I was proposing.  I was unaware how Lua tables work internally and assumed there must be some sort of internal counter.

I took a moment to look through the Rust standard library implementation (hashbrown), and it really is just a counter maintained from within the library, but it's not necessary for the inner workings of the hash table and is only used for the `.len()` method.  The data structure is 48 bytes, with 8 bytes taken up by the counter.  Lua tables are 40 bytes, as they do not have an internal counter.  Lua tables have to serve as literally all data structures, so the only thing that introducing a counter does is increase memory usage across the board.

For `table.count`, I did not suggest this because it implies some sort of overall count, rather than just counting the dictionary part.  The name also implies that the entire table will be counted then and there rather than accessing an internal counter.  If someone wants to propose a counting function, feel free to do so, but that was not what I was going for here.

In conclusion, Lua tables without a counter provides the best balance of tradeoffs for the language.

PR #135, opened 2025-08-23
RFC: Function Calls with Named Arguments

itsmath:
[Rendered](https://github.com/matlneus/luau-rfcs/blob/function-named-arguments/docs/function-named-arguments.md)

This RFC proposes adding optional __named arguments__ to Luau function calls, allowing arguments to be specified by parameter name rather than position.

(I liked @Wunder-Wulfe's idea in https://github.com/luau-lang/rfcs/pull/91#issuecomment-2676560356)

### DIFF
--- a/docs/function-named-arguments.md
+++ b/docs/function-named-arguments.md
@@ -1,0 +1,341 @@
+# Function Calls with Named Arguments
+
+## Summary
+
+This RFC proposes adding optional __named arguments__ to Luau function calls, allowing arguments to be specified by parameter name rather than position.
+
+## Motivation
+
+Function calls in Luau currently rely on positional arguments, which can make code hard to read when parameters are not self-describing. This is especially problematic for primitive types, such as numbers and booleans, which carry no semantic meaning outside their context.
+
+Consider this example:
+
+```luau
+type FS = {
+    read watch: (path: string, recursive: boolean) -> (),
+} 
+
+local fs: FS
+
+fs.watch("/", true)
+```
+
+Here, the first argument is self-describing (in that for a filesystem function, you're expecting a path), but the second argument is ambiguous: it could be `recursive`, `blocking`, `follow_symlinks`, et al. If you don't read the function definition, you __can't__ know.
+
+The solution proposed by this RFC is to use named arguments:
+
+```luau
+fs.watch("/", recursive=true)
+```
+
+For another example, consider this pattern, commonly seen in game development:
+
+```luau
+-- I'll omit the definition of `Car` to make the point clear
+
+local car: Car
+
+car.new(350.0, 275.0, john_doe)
+```
+
+Do the first two arguments refer to hit points, horsepower, maximum speed, or acceleration? Who is `john_doe`, the owner, or the driver?
+
+Using named parameters makes it clear:
+
+```luau
+car.new(horsepower=350.0, hp=275.0, owner=john_doe)
+```
+
+A less common but still relevant issue is:
+
+```luau
+type Foo = (bar: Bar, qux: Qux, foobar: number) -> string
+type Bar = () -> (...any)
+type Qux = (quux: number, quxquux: number, foobar: number, ...any) -> string
+
+local foo: Foo = function(bar: Bar, qux: Qux, foobar: number): string
+    return qux(bar(), foobar=foobar)
+end
+```
+
+This sort of pattern is typically seen in event handlers, and it is currently not possible to do without making extra unnecessary function calls.
+
+To end the motivation section, here are a few functions from Roblox, a popular runtime for Luau, that could benefit from named arguments:
+
+```luau
+TweenInfo.new(time=2.0, delayTime=0.5, repeatCount=5, reverses=true)
+
+workspace:Raycast(origin=Vector3.zero, direction=-10.0 * Vector3.yAxis)
+
+TweenService:SmoothDamp(current=Vector3.zero, target=target, velocity=Vector3.yAxis, smoothTime=1.0, maxSpeed=10.0, dt=dt)
+```
+
+## Design
+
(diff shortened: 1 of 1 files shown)


jackdotink:
I do not think it is possible to implement this into the VM in any way that is actually any good. It will always be equivalent to or worse than passing a table, so you should probably just pass a table.

Wunder-Wulfe:
> It will always be equivalent to or worse than passing a table, so you should probably just pass a table.

I believe that even if the internal implementations for such behavior ends up involving a temporary table object, it would still be useful, as you would be able to perform functions equivalent to keyword arguments in python.

```py
def example(**kwargs):
    print(kwargs.get("arg", 0))

# alternative definition which accepts ‘arg’ as a positional argument
"""
def example(arg=0):
    print(arg)
"""

# > 0
example()
# > 123
example(arg=123)
```

This provides a clearer context to the function call and definition, removing the need for representations such as `f(args…,{kwd=val})` and `function f(args…, kwds: {kwd: valtype?}?)`, replacing it with `f(args…, kwd=val)` and `function f(args…, kwd: valtype=valdefault)`

deviaze:
Another alternative you can add in alternatives:

luau tooling (namely luau-lsp) already comes with parameter names for nonobvious literals: 

image

jackdotink:
What about imported functions? Additionally, the compiler does not run type analysis - it does not know if a value is a function and if so what the type of that function is. It only knows what it sees is explicitly declared in one file.

matlneus:
> What about imported functions?

I don't see how the function being imported would change things? If the type of the value is known, then named arguments could work.

> Additionally, the compiler does not run type analysis - it does not know if a value is a function and if so what the type of that function is. It only knows what it sees is explicitly declared in one file.

This is a much more interesting point. I assumed that the compiler has at least some sort of type analysis since Luau JIT is said to use types[^1], and if it is true that Luau doesn't run type analysis at compile-time, this feels like something that could be improved, as imported type aliases (newtypes) are quite common.

[^1]: https://luau.org/performance

jackdotink:
The Luau compiler reads type annotations, it does not perform any analysis. It will not perform analysis because that would greatly reduce the speed of the compiler, which is a blocker for roblox, the primary user and funder of Luau.

aatxe:
I haven't read all the discussion here, but my findings weren't really performance-related at all. It was just prototyping a reasonable implementation of named arguments to see how it felt, and it was overall alright. I ended up getting horribly sick with Covid which derailed me really bringing it to its conclusion, but the general implementation approach was to compile named arguments into passing a table and then unpacking the table in the top of the function. It didn't seem like something that would have super significant compilation time impacts, but it certainly comes at a performance cost for every function call, and it's not clear that it would be worth the effort of adding to the language.

aatxe:
> Another alternative you can add in alternatives:
> 
> luau tooling (namely luau-lsp) already comes with parameter names for nonobvious literals:
> 
> image

I don't think it's really fair to characterize "argument name inlay hints" as an alternative to named arguments.

aatxe:
> The Luau compiler reads type annotations, it does not perform any analysis. It will not perform analysis because that would greatly reduce the speed of the compiler, which is a blocker for roblox, the primary user and funder of Luau.

I mean, more generally, I don't believe we have any known users of Luau who are _not_ interested in the compilation speed being very, very fast. If there are users who don't care, they haven't vocalized that to us.

docs/function-named-arguments.md
```diff
@@ -0,0 +1,341 @@
...
+TweenInfo.new(time=2.0, delayTime=0.5, repeatCount=5, reverses=true)
+
+workspace:Raycast(origin=Vector3.zero, direction=-10.0 * Vector3.yAxis)
+
+TweenService:SmoothDamp(current=Vector3.zero, target=target, velocity=Vector3.yAxis, smoothTime=1.0, maxSpeed=10.0, dt=dt)
+```
+
+## Design
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this section are to be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
```
aatxe, line 75: We don't do this IEEE RFC stuff, and I think it's for the best that we don't. It's far too legalistic for the level of process actually intended to be applied here.

docs/function-named-arguments.md
```diff
@@ -0,0 +1,341 @@
...
+foo(
+    foo=404, -- This is the value that would actually be used.
+    foo=409  -- This SHOULD be linted, since it is a duplicate.
+)
+```
+
+If a named argument is used in an union or intersection of functions, two conditions must be met:
+
+1. All functions in the union/intersection MUST allow the named argument by themselves (following the same semantics described by the RFC).
+2. The respective parameter for the named argument MUST have the same position across all functions in the union/intersection.
```
aatxe, line 190: A union of functions is not callable. Period. That's the semantics of the language today. You could instead describe this reasonably as for an overloaded function call to work with named arguments, each overload must support identical argument names with identical positions across all functions, but this would basically amount to "almost no overloaded functions could ever be called" because the vast majority of overloaded functions have different arities and types.

docs/function-named-arguments.md
itsmath: This was my first RFC and I really didn't know how much "legalese" I had to use 😭 Thanks!

docs/function-named-arguments.md
itsmath: I have no idea how I didn't notice that. But yes it would just be for overloaded functions. The overloads could have different amounts of parameters, and different types. They just couldn't have parameters with the same name in different positions, since that means you can't know what argument the name actually applies to.

itsmath:
I think my idea here was just incompatible with the current goals of Luau since it would require the compiler to depend on type analysis, so I'll be closing this RFC for now. I'll update it if I come up with another idea (but it'd most likely be equivalent to table arguments)

PR #136, opened 2025-08-26
Documentation Comments

gaymeowing:
[Rendered](<https://github.com/gaymeowing/rfcs/blob/doc-comments/docs/documentation-comments.md>)


gaymeowing (commented):
Iffy on this, as it differs from the behavior luau has for defining a variable twice in a table ie
```luau
{
     mrrp = "meow", -- is overrided by the next declaration of mrrp
     mrrp = true, 
}
```

docs/doc-comments.md
```diff
@@ -0,0 +1,163 @@
...
+end
+```
+
+```luau
+local module = {
+	--- meow
+	meow = "mrrp"
+}
+
+--- This documentation comment will not override the documentation comment for module.meow
```
gaymeowing: Iffy on this, as it differs from the behavior luau has for defining a variable twice in a table ie
```luau
{
     mrrp = "meow", -- is overrided by the next declaration of mrrp
     mrrp = true, 
}
```

docs/doc-comments.md
```diff
@@ -0,0 +1,162 @@
...
+
+Documentation comments are to be automatically detected by luau, with them requiring no whitespace in between the end of the comment and whatever they're commenting on be it a variable, type, table, or function. Although a single new line is allowed at the end of a comment, as the following would be quite ugly:
+
+```luau
+--[[
+	mrow meow mrrp
+]]local function meow()
+	-- code here
+end
+```
```
Dekkonot: Are doc comments in this Instance a part of the function type, or are they attached to the variable? e.g. would `local foo = meow()` carry the doc comment into `foo`?

If not, that's a break from how Luau LSP does it at the moment:
image

gaymeowing: Its supposed to carry

Dekkonot:
Are local variables meant to have doc comments under this proposal? e.g.

```luau
-- This is an important constant
local SCARY_CONSTANT_NAME = "(>'-')> <('-'<) ^(' - ')^ <('-'<) (>'-')> kirby dance"
```

It's unclear at a glance at the moment. I think this would benefit from a list of things that would actually track doc comments. Obviously things like loops wouldn't, but if functions, types, and variables do, that should be written down more clearly than just a bunch of examples.

docs/doc-comments.md
```diff
@@ -0,0 +1,164 @@
+# Documentation Comments
+
+## Summary
+
+Adds builtin support for documentation comments to luau.
+
+## Motivation
```
deviaze: Should lead with ecosystem fragmentation more directly.

```suggestion
## Motivation

The Luau ecosystem is currently fragmented in respect to documentation comment syntax.

In Luau Language Server, documentation comments are attached to `local` and `type` bindings, and are defined above the binding to be documented, using `---`, `--[=[`, and `--[[` comment syntax.

In Roblox Studio, documentation comments are any comments defined above a function binding <check this for accuracy>.

This RFC aims to unite the Luau ecosystem's tooling by providing first-class support for documentation comments by the language itself.
```

docs/doc-comments.md
```diff
@@ -0,0 +1,164 @@
...
+--[[
+	I'm a documentation comment!
+]]
+type Mrrp = "miaow"
+
+type Mrow = Meow<Mrrp> -- no documentation comment
+```
+
+## Design
+
```
deviaze: A documentation comment (or doc comment) is defined as any comment zero or more newlines above a `local` binding, `function` binding, `type` binding, `function` parameter, `type` property, or `table` property.

This means:

```luau
-- This is a doc comment
local cats = {}
```

```luau
-- I am not a doc comment

-- i am a doc comment
function cats.get(name: string): Cat
```

docs/doc-comments.md
```diff
@@ -0,0 +1,164 @@
...
+### Whitespace Requirements
+
+Documentation comments are to be automatically detected by luau, with them requiring no whitespace in between the end of the comment and whatever they're commenting on be it a variable, type, table, or function. Although a single new line is allowed at the end of a comment, as the following would be quite ugly:
```
deviaze: this should be reworded to be more specific and precise

docs/doc-comments.md
```diff
@@ -0,0 +1,164 @@
...
+
+```luau
+local function make_them_meow(
+	-- The name of the person who will meow
+	person: string,
+	-- The length of the meow
+	len: number?
+)
+	print(`{person} meowed: me{string.rep("o", len or 1)}w`)
+end
```
deviaze: Should param comments on the RHS of the param also count?

```luau
function cats.add(
    name: string, --- the cat's name
    age: number --- the cat's age
)
```

docs/doc-comments.md
```diff
@@ -0,0 +1,164 @@
...
+
+```luau
+---- I have extra dashes!
+```
+
+Would appear as "I have extra dashes" instead of "-- I have extra dashes!".
+
+### Carrying
+
+Documentation comments automatically carry from variable to variable, and from type to type; unless overriden which can be seen in the example of 2 modules below below.
```
deviaze: this part needs more text about how this currently works in luau lsp rather than examples

docs/doc-comments.md
```diff
@@ -0,0 +1,164 @@
...
+### Type functions
+
+Documentation Comments will also be exposed in type functions with the following methods added to the `type` instance:
+
+#### `type` Instance
+
+| Instance Methods | Return Type | Description |
+| ------------- | ------------- | ------------- |
+| `getcomment()` | `string?` | returns the comment attached to the type; if there is no comment attached it returns nil |
+| `setcomment(comment: string?)` | `()` | adds / overrides the type's documentation comment; if comment is nil or is a string with a length of 0, removes the types's comment |
```
deviaze: only documentation comments will be accessible to this api, therefore I recommend naming the apis `:documentation()` and `:setdocumentation()` to more clearly indicate what's actually being gotten and set.

```suggestion
| ------------- | ------------- | ------------- |
| `getcomment()` | `string?` | returns the comment attached to the type; if there is no comment attached it returns nil |
| `setcomment(comment: string?)` | `()` | adds / overrides the type's documentation comment; if comment is nil or is a string with a length of 0, removes the types's comment |
```

docs/doc-comments.md
```diff
@@ -0,0 +1,164 @@
+# Documentation Comments
+
+## Summary
+
+Adds builtin support for documentation comments to luau.
+
```
deviaze: Briefly mention what documentation comments are, why we want them (hovers), and that this RFC does not say anything about what should be in them/how they are rendered (markdown, html, plain text only, etc.)

deviaze (commented):
Left some comments for improved specificity, wording, API, etc.

docs/doc-comments.md
gaymeowing: No, as that's not before the declaration. (this is also now clarified at the top of the design section)

docs/doc-comments.md
gaymeowing: Added a note about how this mimics luau-lsps current behavior, I think that's good enough?

docs/doc-comments.md
gaymeowing: Didn't bother explaining what documentation comments are as they already exist in roblox since forever (didn't always allow user-defined ones but eh), but did specify that this RFC doesn't require any specific format of the comment body.

docs/doc-comments.md
gaymeowing: Roblox and luau-lsp are the same in terms of document comment syntax (except for comment bodies), and I don't really see how this would improve the RFC, as its already explained decently.

deviaze: Luau Language Server doesn't allow `--` to be documentation comments whereas Roblox does.

docs/doc-comments.md
alexmccord: The carrying you see here is literally due to an implementation quirk. `foo` is a _bound type_ to the type of `meow`, which contains a docstring in it, and all Luau LSP does is follow a bound type through until it finds a real type, and voila, you see a docstring.

I would not use existing behavior as something worth preserving.

docs/documentation-comments.md
```diff
@@ -0,0 +1,191 @@
...
+### Type functions
+
+Documentation Comments will also be exposed in type functions with the following methods added to the `type` instance:
+
+#### `type` Instance
+
+| Instance Methods | Return Type | Description |
+| ------------- | ------------- | ------------- |
+| `documentation()` | `string?` | returns the documentation attached to the type; if there is no documentation attached it returns nil |
+| `setdocumentation(documentation: string?)` | `()` | adds / overrides the type's documentation; if documentation is nil or is a string with a length of 0, removes the types's documentation |
```
alexmccord: Doesn't work with table properties/indexers or function parameters.

gaymeowing: > Doesn't work with table properties/indexers or function parameters.

Well keys are types so I would expect the documentation to be attached to the keys, and same with parameters because they're also made up of types.

Possibly also have documentation attached to values? As I think it'd make a bit more sense if the value was able to have its own documentation technically.
for example:
```luau
-- mrrrp
type Meow = "mrrp"

type MeowMap = {
     -- the first type of meow
     meow: Meow,
}

--[[ 
     in this case calling :properties() on the MeowMap table type instance in a type function 
     would have the key have its documentation set as "the first type of meow", and the value would have its documentation set as "mrrrp"
]]
```

matlneus:
Can we also allow the following syntax for documentation comments?:

```luau
local PI: number = 3.14 -- This is an important constant
```

The only reason this is not used today in code is because it is not supported by `luau-lsp` or Roblox. But it is useful if you're defining a lot of simple documentation comments in a row, for example:

```luau
function foobarqux(
    foo: number, -- The foo
    bar: number, -- The bar
    qux: number  -- The qux
): number
    return foo + bar + qux
end
```

docs/documentation-comments.md
```diff
@@ -0,0 +1,193 @@
...
+
+```luau
+--[[
+	I have extra dashes at the end!
+--]]
+```
+
+### Carrying
+
+Documentation comments automatically carry from variable to variable, and from type to type; unless overriden which can be seen in the example of 2 modules below below.
```
itsmath: Since in this RFC, documentation comments are attached to types, does that mean you're cloning the type by overriding the documentation comment? The reason `luau-lsp` has this behavior of carrying the type is (from what I can tell) just because the LSP always uses the comment on the original table type. 

In fact, currently this code does not work in `luau-lsp`:

```luau
--- The foobar comment
type Foobar = {
    foo: number,
    bar: number,
}

--- The quxquux comment, which is useless because it does not override anything.
type Quxquux = Foobar
```

docs/documentation-comments.md
```diff
@@ -0,0 +1,193 @@
...
+This mimics luau-lsp's existing behavior for handling documentation comments.
+
+Module A:
+
+```luau
+-- mrrp
+export type Meow = "mrrp"
+
+type Mrrp = module_a.Meow -- Because Mrrp is just Meow, it'll have the same documentation comment as Meow
```
itsmath: How do we decide if a type A is __just__ type B? What about, for example, type parameters?

```luau
-- internalTypes.luau

--- A comment
export type InternalType<T> = {
    parameter: T,
}
```

```luau
-- publicTypes.luau

local internalTypes = require("./internalTypes")

export type Type<T> = internalTypes.InternalType<T>
```

Should `Type<T>` have the same documentation as `internalTypes.InternalType<T>`? Currently in luau-lsp [this is not how it works](https://github.com/JohnnyMorganz/luau-lsp/pull/956#issuecomment-2700542989).

docs/doc-comments.md
itsmath: I don't think so. The way this currently works in `luau-lsp` is not explicitly defined at all. In the LSP, table types can have documentation comments attached to them and they cannot be overriden, but other type aliases don't carry doc comments at all. See the `luau-lsp` PR [#956](https://github.com/JohnnyMorganz/luau-lsp/pull/956) to see why that is.

docs/documentation-comments.md
```diff
@@ -0,0 +1,193 @@
...
+Documentation Comments will also be exposed in type functions with the following methods added to the `type` instance:
+
+#### `type` Instance
+
+| Instance Methods | Return Type | Description |
+| ------------- | ------------- | ------------- |
+| `documentation()` | `string?` | returns the documentation attached to the type; if there is no documentation attached it returns nil |
+| `setdocumentation(documentation: string?)` | `()` | adds / overrides the type's documentation; if documentation is nil or is a string with a length of 0, removes the types's documentation |
+
+For table fields their documentation will be attached to the key type instances, and function parameters will have their documentation be attached to the type instances that make up the head and tail.
```
itsmath: I think this is confusing. Why not just put all documentation comments on the value types? 

Documentation on the key of a table field should be different from documentation on the value of a table field.

docs/documentation-comments.md
```diff
@@ -0,0 +1,193 @@
...
+	-- I meow! But I get linted :(
+	callback = function()
+		print("meow")
+	end,
+}
+```
+
+### Type functions
+
+Documentation Comments will also be exposed in type functions with the following methods added to the `type` instance:
```
itsmath: What about the type instances that don't change? For example:

```luau
type function IsAnyAlwaysAny()
    return types.singleton(types.any == types.any)
end
```

Since `IsAnyAlwaysAny` is true, this would imply that `types.any` could only ever have one documentation comment that is applied globally to every type alias. 

```luau
type Foo = any
```

gaymeowing:
> Can we also allow the following syntax for documentation comments?:
> 
> ```luau
> local PI: number = 3.14 -- This is an important constant
> ```
> 
> The only reason this is not used today in code is because it is not supported by `luau-lsp` or Roblox. But it is useful if you're defining a lot of simple documentation comments in a row, for example:
> 
> ```luau
> function foobarqux(
>     foo: number, -- The foo
>     bar: number, -- The bar
>     qux: number  -- The qux
> ): number
>     return foo + bar + qux
> end
> ```

I don't think this would be a good idea, as that area is better served by not allowing documentation comments. So things like `--TODO:` comments dont have to be so spread apart:
```luau
-- TODO: meow

local ENABLED = true
```

docs/documentation-comments.md
gaymeowing: > How do we decide if a type A is **just** type B? What about, for example, type parameters?

Its still the same type, it just has its generic set

> Should `Type<T>` have the same documentation as `internalTypes.InternalType<T>`? Currently in luau-lsp [this is not how it works](https://github.com/JohnnyMorganz/luau-lsp/pull/956#issuecomment-2700542989).

Yes

docs/documentation-comments.md
gaymeowing: > What about the type instances that don't change? For example:
> 
> ```luau
> type function IsAnyAlwaysAny()
>     return types.singleton(types.any == types.any)
> end
> ```
> 
> Since `IsAnyAlwaysAny` is true, this would imply that `types.any` could only ever have one documentation comment that is applied globally to every type alias.
> 
> ```luau
> type Foo = any
> ```

I would expect documentation() to return nil for primitives, and setdocumentation() to error. Also types in type functions aren't mutable, they're immutable so you couldn't set documentation for any globally anyways unless you redefined any.

itsmath: Wouldn't that imply that we can't have documentation for newtypes?

```luau
--- The metre (or meter in [US spelling]; symbol: m) is the [base unit] of [length] 
--- in the [International System of Units] (SI)
---
--- [US spelling]: https://en.wikipedia.org/wiki/American_and_British_English_spelling_differences#-re,_-er
--- [base unit]: https://en.wikipedia.org/wiki/SI_base_unit
--- [length]: https://en.wikipedia.org/wiki/Length
--- [International System of Units]: https://en.wikipedia.org/wiki/International_System_of_Units
export type Meters = number
```

Is this not possible?

docs/documentation-comments.md
itsmath: What about an user-defined type function? 

```luau
export type Type<T> = udtf.UserDefinedTypeFunction<T>
```

Should `Type<T>` inherit the documentation comment of the user-defined type function itself, or should it use the documentation comment of whatever the result of `udtf.UserDefinedTypeFunction<T>` is?

matlneus:
> I don't think this would be a good idea, as that area is better served by not allowing documentation comments. So things like `--TODO:` comments dont have to be so spread apart:
> 
> ```luau
> -- TODO: meow
> 
> local ENABLED = true
> ```

To me, that’s a strong argument against `--` comments as doc comments. I get that you're trying to make both the Roblox style and the `luau-lsp` style work, but in my opinion it'd be better to choose a standard that should be used from now on, instead of breaking existing conventions (`--` is a comment, `---` is a documentation comment).

Also, as far as I can tell, this RFC doesn't say that tooling MUST only support one style of documentation comment. That means Roblox could continue supporting their own style as a platform-specific "quirk", without forcing it into the broader standard.

docs/documentation-comments.md
gaymeowing: It should infer the result

docs/documentation-comments.md
gaymeowing: > Wouldn't that imply that we can't have documentation for newtypes?
> 
> ```luau
> --- The metre (or meter in [US spelling]; symbol: m) is the [base unit] of [length] 
> --- in the [International System of Units] (SI)
> ---
> --- [US spelling]: https://en.wikipedia.org/wiki/American_and_British_English_spelling_differences#-re,_-er
> --- [base unit]: https://en.wikipedia.org/wiki/SI_base_unit
> --- [length]: https://en.wikipedia.org/wiki/Length
> --- [International System of Units]: https://en.wikipedia.org/wiki/International_System_of_Units
> export type Meters = number
> ```
> 
> Is this not possible?

Newtypes currently don't exist in luau, so yes unless the luau team does something to make it work in the meantime whilst we don't have Newtypes.

gaymeowing:
> To me, that’s a strong argument against `--` comments as doc comments. I get that you're trying to make both the Roblox style and the `luau-lsp` style work, but in my opinion it'd be better to choose a standard that should be used from now on, instead of breaking existing conventions (`--` is a comment, `---` is a documentation comment).

There is no widely used existing convention, so this RFC just supports any comment in support of the goal to work with the most amount of existing codebases today.

> Also, as far as I can tell, this RFC doesn't say that tooling MUST only support one style of documentation comment. That means Roblox could continue supporting their own style as a platform-specific "quirk", without forcing it into the broader standard.

Roblox funds luaus development, so I would expect them to support it.

docs/documentation-comments.md
gaymeowing: That isn't exactly solidified yet, https://github.com/luau-lang/rfcs/pull/136#discussion_r2306056748

docs/documentation-comments.md
gaymeowing: > Since in this RFC, documentation comments are attached to types, does that mean you're cloning the type by overriding the documentation comment? The reason `luau-lsp` has this behavior of carrying the type is (from what I can tell) just because the LSP always uses the comment on the original table type.

Don't know, this RFC doesn't go into specifics about how documentation comments would work internally. So you'd have to ask the luau team about that

docs/documentation-comments.md
itsmath: I think this is something that ought to be fixed or specified in the RFC, though. Currently I just don't see how the Luau team could allow type aliases for primitive types to have documentation comments (I called them newtypes, but maybe that was poor terminology), and these are quite common, think of ECS libraries for example. 

Even worse, this would also mean that function parameters with primitive types couldn’t have documentation on those parameters.

Maybe instead of mutating the type instance when calling `setdocumentation(documentation: string?)`, we could return a new "cloned" type with the documentation applied. 

```luau
type function SpecialAny()
    return types.any:setdocumentation("Our special any") -- The returned type ~= types.any
end
```

itsmath: We also have `types.copy`, so you could also just specify that primitive types need to be copied before documentation can be applied (but to be honest I don't know if primitive types actually get copied with that function or if they are ignored)

docs/documentation-comments.md
itsmath: Your reasoning in that comment makes sense. Now I think it's fine the way it is right now.

(I'd prefer if function parameters also had an equivalent of the "key" type, but for the parameter name, then we could put the documentation comment in there instead of in the value, and it'd be more consistent)

karl-police:
I was actually trying to re-create the way Roblox Studio may look up comments and got a similar result, it's just unfinished.

It's just a basic retrieval of comments...

&nbsp;

But this is some frontend thing, and I'd **expect** customization options.

For example
```lua
local a -- test
```
Whether that comment is included or not, should be configurable.

&nbsp;


And I wonder about cases like these
```lua
-- does not work
local test = 
	-- does work
	function() end
```


Then there's cases like these
```lua
-- a
-- b
--[[c]] --[[d]]
local test = function() end
```
"c" and "d" count as 2 separate comments, but all of the abcd is included, that's Studio's behavior right now


&nbsp;

```lua
-- discluded
--[[
	a
]]
-- b
-- c
local test = function() end
```
and here a,b,c is included

gaymeowing:
> 
> But this is some frontend thing, and I'd **expect** customization options.
> 
> For example
> 
> ```lua
> local a -- test
> ```
> 
> Whether that comment is included or not, should be configurable.

Whilst this seems like a good idea, I could see some packages requiring that to be enabled. Although this could be a luaurc config property, but this isn't supported currently and also complicates parsing for at least I think little benefit.

> 
>  
> 
> And I wonder about cases like these, though I did expect that "does work" being the only one that works
> 
> ```lua
> -- does not work
> local test = 
> 	-- does work
> 	function() end
> ```



> Then there's cases like these
> 
> ```lua
> -- a
> -- b
> --[[c]] --[[d]]
> local test = function() end
> ```
> 
> "c" and "d" count as 2 separate comments, but all of the abcd is included, that's Studio's behavior right now
> 
>  
> 
> ```lua
> -- discluded
> --[[
> 	a
> ]]
> -- b
> -- c
> local test = function() end
> ```
> 
> and here a,b,c is included
> 
>  
> 
> uhhh then there's this case
> 
> ```lua
> --[[
> discluded for some reason
> ]]function testA() end
> ```
> 
> so that means it only looks for the "end", didn't note that
Studio doesn't use luaus ast and instead does some custom weird stuff, so I'm not worrying about full backwards compatibility.
Although I could see multiline short comments being a configurable thing in luaurc, but no-clue on what'd be a good name. As `allow-multiline-short-comments` is sorta vauge but then again luaurc needs proper documentation anyways, and that could have its meaning clarified there.

docs/documentation-comments.md
```diff
@@ -0,0 +1,350 @@
...
+{
+	"documentation": {
+		"ignoredPrefixes": [ "TODO:" ]
+	}
+}
+```
+
+**Note:** Ignored prefixes are applied after [trimming](#trimming).
+
+#### `inferModuleHeaderAsDocumentationForReturn`
```
gaymeowing: This might be a bit much for luau, and maybe encroaching upon what should be left up to language servers.

But its intentionally very bare bones, despite somewhat going against the idea of the RFC to have stuff just work with the most amount of codebases (because I've seen plenty who do fancy ASCII art work).
But that's only Roblox free models and I wouldn't really consider those as things that'd use strict typechecking or be picky about documentation comments.

docs/documentation-comments.md
gaymeowing: Added in it should be on values for now, for consistency as I can't think of any usecases here for wanting the value to have diffrent documentation besides the fact it makes more sense.

docs/documentation-comments.md
gaymeowing: > We also have `types.copy`, so you could also just specify that primitive types need to be copied before documentation can be applied (but to be honest I don't know if primitive types actually get copied with that function or if they are ignored)

Cloning works, as it makes a new type instance even for primitives.
Screenshot 2025-09-02 at 10 56 49

docs/documentation-comments.md
```diff
@@ -0,0 +1,321 @@
...
+
+export type SpaceEncoding = "+" | "%20" -- SpaceEncoding won't have the header as its documentation comment
+```
+
+Theres also an array of ignored prefixes for documentation comments that can be defined in [`luaurc`](./config-luaurc.md) files.
+This is further explained under the [Configuration](#configuration) section.
+
+### Trimming
+
+#### Whitespace
```
gaymeowing: This area might be a bit much, but I think its best to force the displayed versions of comments to be cleaned up by luau. For visual consistency (at least for plaintext, which is the majority of documentation comments) between language servers, so documentation comments will appear as the writer intends always.

gaymeowing: Also will probably end up doing the same semantics that luau-lsp has currently for formatting the displayed versions of documentation comments, as the current described specification is sorta weird.

docs/documentation-comments.md
```diff
@@ -0,0 +1,321 @@
...
+```
+
+### Configuration
+
+A new key `documentation` will be added to [`luaurc`](./config-luaurc.md) files for conifguring how documentation comments will work, with the following options being underneath it:
+
+#### `ignoredPrefixes`
+
+Array of prefixes that indicate any comments starting with these prefixes are to not be detected as documentation comments.
+With ignored prefixes being applied after [trimming](#trimming), so there is no need to allow for string patterns.
```
gaymeowing: Leaving this as is for now, but it'd probably be fine to apply them before trimming.

docs/documentation-comments.md
gaymeowing: > Also will probably end up doing the same semantics that luau-lsp has currently for formatting the displayed versions of documentation comments, as the current described specification is sorta weird.

(leaving this as a note, as I'm pretty sure this is where luau-lsp reformats whitespace for documentation comments)
https://github.com/JohnnyMorganz/luau-lsp/blob/8b75138d062cf2e0f3df89072f31d4c4ff7531c1/src/DocumentationParser.cpp#L405

docs/documentation-comments.md
gaymeowing: Changed my mind on this a bit, as today this is valid code:
```luau
type Meow = () -> ()

type function Mrow()
	Meow:setparameters({ types.number })
	return Meow
end

-- type of the variable MRRP is (number) -> ()
local MRRP: Mrow
```
So the limitation will only be for primitive type instances that are on the types library, as luau can just make a clone for every other primitive given.

andyfriesen:
This RFC appears only to describe a set of conventions that other tools should follow and not something that Luau itself should do anything with.

As such, I don't think it makes sense to require that `.luaurc` files to be extended to support documentation.

gaymeowing:
> This RFC appears only to describe a set of conventions that other tools should follow and not something that Luau itself should do anything with.

Well I was thinking it'd work how require-by-string works, where luau provides the infrastructure for it but its ultimately up to runtimes (or in this case language servers) to then integrate themselves.
Its also virtually impossible for language servers to allow for passing documentation comments thru type functions, unless the language server hardcodes it for a select set of type function names or something. 

> As such, I don't think it makes sense to require that `.luaurc` files to be extended to support documentation.

This was done as to address concerns:
https://github.com/luau-lang/rfcs/pull/136#issuecomment-3229981854

gaymeowing:
> This is all fine, but I don't see why any of this requires changes to `.luaurc`.

The luaurc changes could be hardcoded instead, as come to think of it `inferModuleHeaderAsDocumentationForReturn` doesn't need to be configurable as automatic detection would be good enough.

But I don't see a good way of removing `ignoredPrefixes`, as you either have to hardcode some ignored prefixes like `TODO:`, and then account for every language (I would imagine people using languages other than english would use the version of TODO in their language). 
But then you run into the issue of there being a valid case for `TODO:` not being ignored by default, as to be a reminder to people that "hey this needs doing".
There's also the added issue of what was mentioned here https://github.com/luau-lang/rfcs/pull/136#issuecomment-3229981854 where one doesn't want their documentation comments to be things that don't apply to them.
Plus having `ignoredPrefixes` just included with luau without any need for some external tooling, would be a huge benefit for everyone as:
* It would work anywhere
* Existing codebases would be easier to port over if you made some previous odd choices like using `ROBLOX TODO:` instead of just `TODO:` for example
* Custom `ignoredPrefixes` are nice for allowing internal details to not be surfaced on exports, as you could have `!` as an ignored prefix to indicate a comment is intended for internal use only
* Its also relatively simple and would provide good benefit to everyone

Alternatively the luaurc changes could be removed, but I'm intending for this to hopefully be the only RFC necessary for documentation comments, so I'm just covering all the bases that are reasonable with this RFC.

docs/documentation-comments.md
```diff
@@ -0,0 +1,305 @@
...
+-- I'm a multiline
+-- short comment!
+```
+
+## Design
+
+Documentation comments are to be detected by luau as any comment or short comments that is before (with one newline allowed), or after (with no newlines allowed) a `type`, `variable`, `function` parameter, `type` union component, or `table` property declaration.
+There are some caveats to this though:
+
+* Only short comments can be merged together, as short comments merging with long comments goes against how one would think documentation comments should work. The same goes for merging long comments with other long comments.
```
gaymeowing, line 53: This was already stated above, but this is here just to be very clear with how comments should be merged.
Although this would be a breaking change for Roblox, but I've never seen anyone ever relying on long comments being merged together.

andyfriesen:
> > I'd prefer if we didn't rely at all on documentation tools to have to specify what a documentation comment is. Documentation generation tools should be able to just depend on the Luau parser (and type solver because of user-defined type functions) and get all comments in the file from there.
> 
> I would imagine the luau API would provide the raw comment, so tooling could require `---` if they wished.

This makes much more sense to me.

It's fine to write down conventions, but if someone wants to write a documentation generator, I frankly don't think we have any business telling them what features they can offer or how.

gaymeowing:
> For type functions, only a way to set a comment can be added. This will simply the implementation and also avoid users developing custom DSLs inside type comments and making type function generation dependent on comment data.

Could there be a copy documentation method instead? for copying documentation from one type to another type
```luau
type:copydocumentationto(other_type)
```

docs/documentation-comments.md
```diff
@@ -0,0 +1,248 @@
...
+
+	```luau
+	--[[
+		It'd be very weird
+	]]
+	--[[
+		if
+	]]
+	-- we merged together
+	```
```
alexmccord, line 63: This is an argument to reject the idea of having multiple comment types and stick to one and only one. Because you're also going to have to define which one _is_ the documentation string (let alone whether you would consider any of them to be a documentation string). Is it the first? The last? Or none, because it's not "well-formed?"

docs/documentation-comments.md
```diff
@@ -0,0 +1,248 @@
...
+	-- A spherical shape.
+	| "Ball"
+	-- A block shape.
+	| "Block"
+	-- A cylinder shape oriented along the X axis.
+	| "Cylinder"
+	-- A wedge shape with a slope on one side.
+	| "Wedge"
+	-- A wedge shape with slopes on two sides.
+	| "CornerWedge"
```
alexmccord: Another formatting option:
```luau
-- The PartType enum controls the Shape of a Part.
type PartType =
	| "Ball" -- A spherical shape.
	| "Block" -- A block shape.
	| "Cylinder" -- A cylinder shape oriented along the X axis.
	| "Wedge" -- A wedge shape with a slope on one side.
	| "CornerWedge" -- A wedge shape with slopes on two sides.
```

Not every line necessarily needs to be vertically aligned here, that's a formatter concern. And not every line necessarily needs to be immediately above or to the right, some can be above and others can be to the right. But if it is on the right, then only single-line comment is allowed.

docs/documentation-comments.md
```diff
@@ -0,0 +1,248 @@
...
+	-- I meow! But I get linted :(
+	callback = function()
+		print("meow")
+	end,
+}
+```
+
+</br>
+
+If a field isn't defined in a table literal, fields will only allow their comment to be the first place where the field is defined:
```
alexmccord: With type states, this would be inconsistent. And also consider what if you do this in a branch?
```luau
local m = {}

if flag then
  -- a
  m.f = function() end
else
  -- b
  m.f = function() end
end
```
What documentation string does `m.f` yield? The first one, `-- a` because it occurs first lexically, or since "union of documentation" is nonsense (tagged unions exempted), we don't include any?

Put another way, what would it mean if you did this:
```luau
local function f(x)
  local m = {}

  -- hello! I am an implementation detail. But also probably not.
  m.x = x

  return m
end

return f(5)
```
Does `m.x` have the documentation string `"hello! I am an implementation detail. But also probably not."`? In some cases, it _is_ an implementation detail that type inference happened to scoop up, but you don't want it exposed. In other cases, it's _not_ e.g. you're constructing a module with functions and fields. There's an ambiguity here.

vegorov-rbx:
Community is not a magic entity which doesn't make design mistakes @jackdotink.
And I don't want Luau to follow a design suggestion (---) that will be a mistake.

I have to reiterate once again that tooling can still make those and other arbitrary choices when all `--` comments are captured, but it should not be a choice that the language specification places. You can read my earlier comment on this.

As a side note, I stand by Roblox making the right choice in our tooling, Luau comment syntax was preserved from Lua to document code.

docs/documentation-comments.md
gaymeowing: > With type states, this would be inconsistent. And also consider what if you do this in a branch?
> 
> ```luau
> local m = {}
> 
> if flag then
>   -- a
>   m.f = function() end
> else
>   -- b
>   m.f = function() end
> end
> ```
> 
> What documentation string does `m.f` yield? The first one, `-- a` because it occurs first lexically, or since "union of documentation" is nonsense (tagged unions exempted), we don't include any?

The first I'd assume unless anyone has any objections

> Put another way, what would it mean if you did this:
> 
> ```luau
> local function f(x)
>   local m = {}
> 
>   -- hello! I am an implementation detail. But also probably not.
>   m.x = x
> 
>   return m
> end
> 
> return f(5)
> ```
> 
> Does `m.x` have the documentation string `"hello! I am an implementation detail. But also probably not."`? In some cases, it _is_ an implementation detail that type inference happened to scoop up, but you don't want it exposed. In other cases, it's _not_ e.g. you're constructing a module with functions and fields. There's an ambiguity here.

It would be scooped up either way with the current state of the RFC.

docs/documentation-comments.md
gaymeowing: > Not every line necessarily needs to be vertically aligned here, that's a formatter concern. And not every line necessarily needs to be immediately above or to the right, some can be above and others can be to the right. But if it is on the right, then only single-line comment is allowed.

This is already allowed in the RFC, although perhaps I should be more explicit about it (see the first sentence of the design section).

docs/documentation-comments.md
gaymeowing: > This is an argument to reject the idea of having multiple comment types and stick to one and only one. Because you're also going to have to define which one _is_ the documentation string (let alone whether you would consider any of them to be a documentation string). Is it the first? The last? Or none, because it's not "well-formed?"

It would be the last because the last would be closest to the declaration (assuming the declaration is directly below it).
```luau
--[[
     I can't merge with the comment below me because I'm long and not short!
]]
-- I'm the documentation!
type meow = "mrrp"
```

jackdotink:
The core feature being requested here is luau tracking comments on items. This feature on its own is pretty meaningless and so some kind of usage needs to be envisioned. The usecase I see is language servers providing in-editor information, particularly docs. With that in mind, this is absolutely about documentation and not just about tracking comments. If this is about documentation, then it not only makes sense but is prudent to have a conversation about what good documentation looks like in luau, and what should be encouraged. If we are not going to have a conversation about documentation (the primary and maybe only usecase of this feature) then I think this feature should not be accepted.

If we are going to have a conversation about documentation, then I'd like to go back to my previous questions.
> On the side of having luau capture all comments as a documentation comment, I see no motivation beyond "roblox does it". This cannot be right, so I would love if someone could enlighten me as to any benefits it provides for my workflow instead of telling me about your workflow. Or alternatively, tell me about how using `---` makes your workflow impossible.

vegorov-rbx:
Thank you @gaymeowing for the work on trying to improve Luau and to address all the comments and feedback around this RFC.

However, we on the Luau language team have decided that we are not going to try and define the rules on what is and isn't a comment of interest for external tooling.

Tooling today already has enough information using definition module, definition location and AST locations to extract comments that may be attached to specific entities in code, so most of the RFC only defines guidelines that do not affect Luau implementation and will not be enforceable.
Guidelines for tooling to agree on how to extract this info is still good motivation, but it is out of scope of this part of the project for now.

With that said, if there are issues with missing data, entities that are hard to link together using current AST structure, that should be brought up as issues for us to fix, but those do not require an RFC.

The motivation around type functions not preserving identity of entities (types) is good as it highlights a gap we have in current serialization of types for the type runtime.
After looking into that, we can quickly determine that it's not just about the attached comment, it also currently affects definition module, location, embedder-defined tags and other attached type metadata fields which exist internally, but are not exposed in the type function environment.
This can also be brought up as an issue to solve, we do recognize this problem.

PR #137, opened 2025-08-28
RFC: Function Parameter Names in User-Defined Type Functions

itsmath:
[Rendered](https://github.com/matlneus/luau-rfcs/blob/udtf-function-parameter-names/docs/udtf-function-parameter-names.md)

This RFC proposes a method to add names to function parameters in user-defined type functions.

### DIFF
--- a/docs/udtf-function-parameter-names.md
+++ b/docs/udtf-function-parameter-names.md
@@ -1,4 +1,4 @@
-# Function Parameter Names in User-Defined Type Functions
+# Function parameter Names in User-Defined Type Functions
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


gaymeowing (commented):
Shouldn't names be an array of strings?

alexmccord:
Array of strings might not work. `<A, B, C>(p1: A, B, p3: C) -> ()` is valid. That would correspond to `{type.singleton("p1"), nil, type.singleton("p3")}` which we all know is bad due to `nil` showing up in the middle of it.

matlneus:
> Array of strings might not work. `<A, B, C>(p1: A, B, p3: C) -> ()` is valid. That would correspond to `{type.singleton("p1"), nil, type.singleton("p3")}` which we all know is bad due to `nil` showing up in the middle of it.

Right. That's why I think it should be more like an unordered map. But should we use `type.singleton("name")` or just `"name"`?

matlneus:
From what I can tell, in the Luau source code, [argument names in function types look like this](https://github.com/luau-lang/luau/blob/5189ad927ecae483fc3d26641fc21dc5992ae008/Analysis/include/Luau/Type.h#L381C1-L381C59)

```cpp
struct FunctionType
{
    /* snip */
    std::vector<std::optional<FunctionArgument>> argNames;
    /* snip */
}
```

To me, this suggests we should go with a type that looks something like one of the following:

```luau
-- I'm using [number] here because the table isn't necessarily contiguous,
-- but it could also be written as {string?}?
type ArgumentNames = { [number]: string? }?
```

or

```luau
-- The types should be string singletons
type ArgumentNamesType = { [number]: type? }?
```

Right now I'm wondering if we could get rid of the optional values in the table to make it a nice contiguous array. Perhaps a zero-length string (or string singleton) could represent missing parameter names instead of `nil`? 

I'm just not sure if it's worth it.

matlneus:
@deviaze suggested this approach, which I think is a much nicer solution, but it would unfortunately break existing code. If that's acceptable, since user-defined type functions are relatively new, I think this would be a much better option:

```luau
type Parameters = {
    { name: string?, type: type }
}
```

gaymeowing:
Going to say this RFC should probably not be accepted until https://github.com/luau-lang/rfcs/pull/136 is figured out, as with documentation comments as currently proposed names should be type instances because of this case, so when hovering over the type of the `maow` parameter it'll show the documentation for `Meow`.

```luau
-- Meow
type Meow = "mrrp"

local function mrow(
     -- im the doc comment for the meow parameter
     maow: Meow
)
     -- code here
end
```

PR #138, opened 2025-08-31
Add a note about explicit type instantiation and metatables

alexmccord:
Eventually people will try to do this. We want to be able to point at this section and say why you can't without playing customer support whack-a-mole.

### DIFF
--- a/docs/explicit-type-parameter-instantiation.md
+++ b/docs/explicit-type-parameter-instantiation.md
@@ -115,7 +115,7 @@ In the case of explicit instantiation `f<<T>>` fragment where `f` has a metatabl
 
 1. The forall quantifier is nested under the metamethod for that table's metatable: that's three levels of indirection.
 2. A metatable can have multiple metamethods that are polymorphic: it's not even clear what metamethod to instantiate without a way to disambiguate.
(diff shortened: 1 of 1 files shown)


docs/explicit-type-parameter-instantiation.md
```diff
@@ -109,6 +109,32 @@ local f2 = f<<number>>
 
 When this new syntax is used on a value that is not a function, or does not take those types, a static-time error will be given. This syntax has zero runtime effect.
 
+### Interop with metatables
```
aatxe:
```suggestion
### Interoperability with metatables

```

docs/explicit-type-parameter-instantiation.md
```diff
@@ -109,6 +109,32 @@ local f2 = f<<number>>
 
 When this new syntax is used on a value that is not a function, or does not take those types, a static-time error will be given. This syntax has zero runtime effect.
 
+### Interop with metatables
+
+In the case of an explicit instantiation fragment `f<<T>>` where `f` has a metatable, there are three problems, namely:
```
aatxe:
```suggestion
There are three problems that arise when attempting to instantiate a table with a metatable attached, namely:

```

PR #140, opened 2025-09-11
RFC: function `coroutine.noyield(fn, ...)`

Aseminaunz:
[Rendered](https://github.com/Aseminaunz/rfcs/blob/rfc-function-coroutine-noyield/docs/function-coroutine-noyield.md)

This proposal creates a new standard library function in `coroutine`:
```luau
function coroutine.noyield<A..., R...>(fn: (A...) -> R..., ...: A...): R...
```
It runs a function within the same thread as the caller, without allowing it to yield, reflecting it in `coroutine.isyieldable()` and causing `coroutine.yield()` to error.

### DIFF
--- a/docs/function-coroutine-noyield.md
+++ b/docs/function-coroutine-noyield.md
@@ -1,0 +1,438 @@
+# `coroutine.noyield()` function
+
+## Summary
+
+A new standard library function, `coroutine.noyield(fn, ...)`, which runs a function without allowing it to yield.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


jackdotink:
@vrn-sn This is true, but unimportant because this is already how these mechanisms work today when attempting to yield in non-yieldable contexts.

vegorov-rbx:
This is something we might return to in the future with an alternative proposal, but RFC in this form is not accepted because of the problems mentioned before.

PR #141, opened 2025-09-19
RFC: NoInline Attribute for Functions

TechHog8984:
This RFC proposes a noinline function attribute to explicitly tell the compiler to not inline a function.
I based many elements of this RFC off of the native and deprecated function attribute RFCs so I apologize for any deja vu.

Also, I have already made an implementation for this since I am not too familiar with the RFC process and wasn't properly informed, so should I add `Status: Implemented` or wait until the potential acceptance of this RFC?


TechHog8984:
Closing in favor of 128, oops!

PR #142, opened 2025-09-22
Create string-build-function.md

ssynical:
Propose `string.build()`

### DIFF
--- a/docs/string-build-function.md
+++ b/docs/string-build-function.md
@@ -1,0 +1,172 @@
+# string.build() function proposal
+
+## Summary
+
+I propose to add a `string.build` function to provide a convenient API for multi-argument string concatenation, addressing the usability issues of the current `table.concat` approach while maintaining linear time complexity.
+
+## Motivation
+
+String concatenation in Luau using the `..` operator has quadratic time complexity (`O(n²)`) when used repeatedly, creating significant performance issues in string-heavy code.
+
+Performance testing demonstrates the severity of this problem:
+
+| Iterations | Naive `..` | `table.concat` | Performance Gap |
+|------------|------------|----------------|-----------------|
+| 1,000      | 2.1ms      | 0.2ms          | **10.5x slower** |
+| 5,000      | 35.4ms     | 0.8ms          | **44.3x slower** |
+| 10,000     | 137.7ms    | 1.5ms          | **91.8x slower** |
+| 20,000     | 503.9ms    | 3.3ms          | **152.7x slower** |
+
+For example, a developer may have such logic:
+
+```lua
+local result = ""
+for i = 1, 10000 do
+    result = result .. "item " .. i .. "\n"
+end
+```
+where processing becomes increasingly slow with large `n`, taking 138ms for 10,000 iterations.
+
+The current workaround requires managing arrays manually with `table.concat` as such:
+```lua
+local parts = {}
+for i = 1, 10000 do
+    table.insert(parts, "item ")
+    table.insert(parts, tostring(i))
+    table.insert(parts, "\n")
+end
+local result = table.concat(parts)
+```
+which is efficient at 1.5ms, but it's verbose and requires developers to manually separate values into individual table elements.
+
+This pattern is in many real codebases, such as:
+- Message/Text systems that build formatted messages with player names, timestamps, and content
+- UI text generation combining player stats, currency formatting, and game state
+- Debug output combining object properties, coordinates, and state information
+- Command parsing and help text generation for admin tools and game commands
+
+Every single one of these scenarios forces developers to choose between performance and maintainability, which is a choice that shouldn't exist in a modern programming language.
+
+The `table.concat` approach itself forces developers to:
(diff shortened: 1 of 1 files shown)


docs/string-build-function.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Alternatives
+
+- **Maintain status quo.** The current situation actively harms developer productivity and code quality. Unacceptable.
+
+- **Do nothing.** Continue requiring developers to manually manage tables for efficient concatenation. This leaves the ergonomic issues in place but maintains optimal performance for those who understand the patterns.
+
+- **Optimize the `..` operator.** Make `..` automatically efficient for multiple concatenations. This would require significant runtime changes and might break existing performance expectations of the operator.
+
+- **Add string interpolation syntax.**  A much larger language change requiring parser modifications. While potentially valuable, it doesn't address the programmatic string building use cases.
```
itsmath, line 168: Luau already has string interpolation, for example:

```luau
local world = "world"
local hello_world = `Hello, {world}!`

print(hello_world) --> Hello, world!
```

ssynical: Yes, Luau does have great string interpolation syntax. My meaning behind the alternative is as so:

`string.build()` addresses a different problem that interpolation doesn't solve. The key difference between what I'm proposing and the existing string interpolation is that string interpolation is good for templates where you know the structure at write-time, and good for single variable and single expression substitution and string creation, but `string.build` will be used for string construction where the number/type of parts varies and programmatic building where you don't know the exact structure ahead of time, along with multiple argument concatenation without sacrificing code readability. 

For example, consider these snippets:

```lua
local function format_user_stats(user, include_premium, include_level, include_coins)
    local result = `Name: {user.name}`
    if include_level then
        result = result .. `\nLevel: {user.level}`
    end
    if include_coins then
        result = result .. `\nCoins: {user.coins}`
    end
    if include_premium then
        result = result .. `\nPremium: {user.is_premium and "Yes" or "No"}`
    end
    return result
end
```
It looks rather awkward with interpolation, whereas with my proposal, the structure turns into something like this:

```lua
local function format_user_stats(user, include_premium, include_level, include_coins)
    local parts = {`Name: {user.name}`}
    if include_level then
        table.insert(parts, string.build("Level: ", user.level))
    end
    if include_coins then
        table.insert(parts, string.build("Coins: ", user.coins))
    end
    if include_premium then
        table.insert(parts, string.build("Premium: ", user.is_premium and "Yes" or "No"))
    end
    return table.concat(parts, "\n")
end
```

which cleans up the logic and makes it more readable.

Interpolation doesnt fix the performance issue I addressed earlier, the problem that `string.build` would fix.

```lua
local result = ""
for i = 1, 10000 do
    result = result .. `Item {i}: {items[i].name}\n`  -- Still quadratic, still O(n²)
end
```

String concatenation is also inefficient with 2-3+ args as shown.

The core issue remains that developers still face the choice between readable code and performant code when building strings programmatically, which `string.build()` would resolve.

Cooldude2606: Using the snippets above, the following code achieves both objectives of this RFC, namely being readable and performant, using existing string interpolation and only uses `table.concat` where your proposal would also require it.

```luau
local function format_user_stats(user, include_premium, include_level, include_coins)
    local parts = { `Name: {user.name}` }
    if include_level then
        table.insert(parts, `Level: {user.level}`)
    end
    if include_coins then
        table.insert(parts, `Coins: {user.coins}`)
    end
    if include_premium then
        table.insert(parts, `Premium: {user.is_premium and "Yes" or "No"}`)
    end
    return table.concat(parts, "\n")
end
```

```luau
local parts = table.create(10000)
for i = 1, 10000 do
    parts[i] = `Item {i}: {items[i].name}` -- Same time complexity as string.build would have
end
local result = table.concat(parts, "\n")
```

I would also like to point out that all snippets in the RFC are directly replaceable with `table.concat{}` without any added boilerplate. And that the primary motivating example still requires the use of `table.contact(parts)` as follows:

```luau
local parts = table.create(20_000)
for i = 1, 20_000 do
    parts[i] = string.build("item ", i)
    -- parts[i] = "item " .. i             -- x220 speedup over `result ..=`
    -- parts[i] = table.concat{"item ", i} -- x100 speedup over `result ..=`
    -- parts[i] = `item {i}`               -- x280 speedup over `result ..=`
end
local result = table.concat(parts, "\n")
```

```lua
local greeting = string.build("Hello, ", name, "! Welcome to ", game_name, ".")
-- is equivalent to below, no reason to add `parts` as shown in the correct usage section
local greeting = table.concat{"Hello, ", name, "! Welcome to ", game_name, "."}
local greeting = `Hello, {name}! Welcome to {game_name}.` -- even better as str interop
```

Do you have a counter example where `string.build` cannot be replaced with string interpolation?

jackdotink:
My problem is that this is the same thing as string interpolation. I have not seen example usage of this function which is not trivially replaceable with string interpolation.

PR #143, opened 2025-10-02
RFC: Single-quoted strings as singletons.

andyfriesen:
[Rendered](https://github.com/luau-lang/rfcs/blob/7be17502fda542ffc23969244a47a87809f89e8e/docs/syntax-single-quotes-for-singleton-strings.md)

### DIFF
--- a/docs/syntax-single-quotes-for-singleton-strings.md
+++ b/docs/syntax-single-quotes-for-singleton-strings.md
@@ -1,0 +1,133 @@
+# Use `'` for Singleton Strings
+
+## Summary
+
+Adjust type inference to assume that string literals written using single quotes (eg `'string'`) should always be inferred to have a singleton type, and that strings written using `"double quotes"` should always be typed `string`.
+
+## Motivation
+
+It's quite tricky for us to consistently infer a singleton type in exactly all of the places where it's what the developer intended, and for us to always infer `string` where it's what the developer intended.
+
+Today, our inference strategy is to assume that a string literal should have type `string` if possible, but that type can be constrained to a singleton if necessary.
+
+This strategy is broadly successful, but fails in a few ways.
+
+### Bugs
+
+We have bugs that prevent us from deducing the proper type in certain cases.  For example: https://github.com/luau-lang/luau/issues/1483
+
+```lua
+type Form = "do-not-register" | (() -> ())
+
+local function observer(register: () -> Form) end
+
+observer(function()
+	if math.random() > 0.5 then
+		return "do-not-register" -- TypeError: Type 'string' could not be converted into '"do-not-register" | (() -> ())' Luau(1000)
+	end
+	return function() end
+end)
+```
+
+### Singleton Strings and Generics
+
+The currently implemented inference strategy doesn't always do what the developer intended in the presence of generics.  Whenever a string literal is passed to a generic, it is possible for it to be typed `string`, and so that's what Luau infers.
+
+In the following snippet using the [luaup](https://github.com/jackdotink/luaup) project, a developer clearly wanted `Kind` to be bound to a string singleton and not `string`.  They were instead forced to write a bunch of casts:
+
+```lua
+local function new_token<Kind>(kind: Kind, text: string?): cst.TokenKind<Kind>
+    return { kind = kind, text = text or kind :: any, span = span, trivia = trivias }
+end
+
+return table.freeze {
+    semicolon = new_token(';' :: any) :: cst.TokenKind<';'>,
+    equals = new_token('=' :: any) :: cst.TokenKind<'='>,
+    colon = new_token(':' :: any) :: cst.TokenKind<':'>,
+    comma = new_token(',' :: any) :: cst.TokenKind<','>,
+    dot = new_token('.' :: any) :: cst.TokenKind<'.'>,
+    endd = new_token('end' :: any) :: cst.TokenKind<'end'>,
+}
+```
+
+## Design
+
+Languages like Lisp and Erlang offer a concept called an _atom_, which is essentially an interned string.  Atoms and Luau string singletons are largely used for the same purpose: People use them to tag their unions and to build enumerations.
+
+Given that this is the most common use case, it makes quite a lot of sense to offer separate syntax to allow developers to be precise about what they want.
+
+In this RFC, we propose the syntax `"foo"` for a string of type `string`, and `'bar'` for a string with the singleton type `"bar"`.
+
+This makes a whole bunch of inference scenarios completely trivial.
+
+```lua
+-- fruits : {string}
+local fruits = {}
+table.insert(fruits, "avocado") -- These are just strings
+table.insert(fruits, "hazelnut")
+table.insert(fruits, "carolina reaper")
+```
+
+```lua
+type Ok<T> = {
+    tag: 'ok', -- the tag of a tagged union should use single quotes
+    data: T
+}
+type Err<E> = {
+    tag: 'error',
+    error: E
+}
+type Result<T, E> = Ok<T> | Err<E>
+
+function try_very_hard()
+    local n = math.random()
+    if n > 0 then
+        return {tag='ok', data=n}
+    else
+        return {tag='err', error="I failed"} -- no ambiguity: The tag is a singleton and the error message is not.
+    end
+end
+
+local r: Result<number, string> = try_very_hard()
+```
+
+And
+
+```lua
+local function new_token<Kind>(kind: Kind, text: string?): cst.TokenKind<Kind>
+    return { kind = kind, text = text or kind :: any, span = span, trivia = trivias }
+end
+
+return table.freeze {
+    semicolon = new_token(';'), -- The string here uses single quotes, and so the generic Kind can only be ';'
+    equals = new_token('='),
+    colon = new_token(':'),
+    comma = new_token(','),
+    dot = new_token('.'),
+    endd = new_token('end'),
+}
+```
+
+### Implementation
+
+This is a very simple adjustment to the constraint generation phase of type checking.  Most of the actual work involves disabling the code to infer when a particular string literal must be a singleton.
+
+It is so easy to implement, in fact, that we've already done it as an experiment.  As of this writing, you can try it out by setting the flag `DebugLuauStringSingletonBasedOnQuotes`.
+
+## Drawbacks
+
+There are a number of pretty serious drawbacks to this proposal:
+
+First, and arguably most importantly: The inference we already have is performing quite well!  I could only find one ticket that points out a bug in the system: https://github.com/luau-lang/luau/issues/1483.  This bug is a little bit esoteric and can probably be fixed in a reasonable timeframe, so it is not itself strong evidence that we should change the language.
+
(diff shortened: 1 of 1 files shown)


docs/syntax-single-quotes-for-singleton-strings.md
```diff
@@ -0,0 +1,133 @@
...
+
+It is so easy to implement, in fact, that we've already done it as an experiment.  As of this writing, you can try it out by setting the flag `DebugLuauStringSingletonBasedOnQuotes`.
+
+## Drawbacks
+
+There are a number of pretty serious drawbacks to this proposal:
+
+First, and arguably most importantly: The inference we already have is performing quite well!  I could only find one ticket that points out a bug in the system: https://github.com/luau-lang/luau/issues/1483.  This bug is a little bit esoteric and can probably be fixed in a reasonable timeframe, so it is not itself strong evidence that we should change the language.
+
+Secondly, most code formatters in modern use will automatically change all string literals to use double quotes because that is considered good style.  If this RFC is implemented, those tools will need to be updated so they do not break type inference.
```
Dekkonot, line 123: The tooling side of this makes this whole thing a dealbreaker to me. This is current default behavior of StyLua:
ezgif-8317ad5f9596de

Notably:
- It enforces double quoted strings
- It converts strings that only escape double quotes into single-quoted strings

This feature being added to Luau would kill both of those formatter options. While that's not its own justification, I think the loss of them would be enough of a downside that I can't support this.

I want my formatter to be able to enforce string-style, and to be able to make things easier to read without changing the semantic meaning of my program, including the type system. I don't want to have to review pull requests to double check that they're using the right quote style for the right situation. That's work that is pointless and completely avoidable.

jackdotink:
Note that I don't use a formatter for Luau, so I've used this without problems others have had.

This has been wonderful to use. It hasn't caused any problems for me when writing or reading code. The worst that happens is that I type a singleton with `"` out of habit, then change it to the singleton syntax.

I for one do not think the loss in formatting options is enough to eliminate this feature, nor do I think it will be possible to infer singleton vs string in every position, so some kind of singleton syntax is needed. This syntax is good, and there are other good options too.

Ukendio:
I am not *opposed* of this personally, as I think it would make sense for single quotes to have different semantics over types than double quotes. And the syntax is pretty great, not that it is hard when it is just standing to the side of casting to literals.

Although the alternative of fixing the bugs and allowing bounded generics to be able to have a strict subtype constraint makes more sense to me. It would likely involve less work for my own codebases.

alexmccord:
```luau
-- this unfortunately becomes legal 
local meow::meow=:meow
```

Regarding this, it doesn't have to be. Try doing `type Id<T>= T`.

MagmaBurnsV:
If you're going to start inferring some string literals as singletons, why not all of them? Almost every style guide like [Roblox's](https://roblox.github.io/lua-style-guide/#literals) prefers double quotes and it just seems oddly alienating not to support it.

My guess is there's some concern with backwards compatibility with the likes of `table.create(50, "hello")` inferring as `{"hello"}` instead of `{string}`. But I'd argue that's preferrable because it's more "correct" even if a little unergonomic.

Regardless, having two different inference rules for what are normally interchangeable style preferences seems a little surprising. Whatever the inference rule is, it should stay consistent for all styles in my opinion.

docs/syntax-single-quotes-for-singleton-strings.md
```diff
@@ -0,0 +1,133 @@
...
+}
+```
+
+## Design
+
+Languages like Lisp and Erlang offer a concept called an _atom_, which is essentially an interned string.  Atoms and Luau string singletons are largely used for the same purpose: People use them to tag their unions and to build enumerations.
+
+Given that this is the most common use case, it makes quite a lot of sense to offer separate syntax to allow developers to be precise about what they want.
+
+In this RFC, we propose the syntax `"foo"` for a string of type `string`, and `'bar'` for a string with the singleton type `"bar"`.
```
alexmccord, line 59: Earlier, the RFC says ["This strategy is broadly successful, but fails in a few ways"](https://github.com/luau-lang/rfcs/pull/143/files#diff-e4cb57eb44865d4022355d35ab3063fb6f189b1a6e073c35063d4f52b2301675R13) and also has [another supporting argument](https://github.com/luau-lang/rfcs/pull/143/files#diff-e4cb57eb44865d4022355d35ab3063fb6f189b1a6e073c35063d4f52b2301675R121), but the RFC as proposed is removing this nice ergonomic feature without a reason?

There's a half-way solution here.
```suggestion
In this RFC, we propose the syntax `'foo'` for a string with the singleton type `'foo'`, and `"bar"` to retain its current behavior standing for either the singleton type `'bar'` or `string`, depending on the bounds.
```

docs/syntax-single-quotes-for-singleton-strings.md
```diff
@@ -0,0 +1,133 @@
...
+end
+
+return table.freeze {
+    semicolon = new_token(';' :: any) :: cst.TokenKind<';'>,
+    equals = new_token('=' :: any) :: cst.TokenKind<'='>,
+    colon = new_token(':' :: any) :: cst.TokenKind<':'>,
+    comma = new_token(',' :: any) :: cst.TokenKind<','>,
+    dot = new_token('.' :: any) :: cst.TokenKind<'.'>,
+    endd = new_token('end' :: any) :: cst.TokenKind<'end'>,
+}
```
alexmccord, line 50: For what it's worth, this could also be solved with explicit instantiation too. `new_token<<";">>(";")`. But I must say, I don't understand why `new_token(';' :: any) :: cst.TokenKind<';'>` is necessary here when `new_token(';' :: ';')` would work just as well?

docs/syntax-single-quotes-for-singleton-strings.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Singleton Strings and Generics
+
+The currently implemented inference strategy doesn't always do what the developer intended in the presence of generics.  Whenever a string literal is passed to a generic, it is possible for it to be typed `string`, and so that's what Luau infers.
+
+In the following snippet using the [luaup](https://github.com/jackdotink/luaup) project, a developer clearly wanted `Kind` to be bound to a string singleton and not `string`.  They were instead forced to write a bunch of casts:
+
+```lua
+local function new_token<Kind>(kind: Kind, text: string?): cst.TokenKind<Kind>
+    return { kind = kind, text = text or kind :: any, span = span, trivia = trivias }
```
alexmccord, line 40: This specific cast is arguably because of the lack of generic bounds. If `TokenKind` defines `text: string`, then `Kind` must be bounded by `string` for `text or kind` to be valid, so this specific example is unsound because you can write `new_token(5)`. If we could utter `Kind: string`, then `string | Kind` always reduces to `string`, and so `text = text or kind` will type check.

alexmccord: If you write `type TokenKind<Kind> = { read kind: Kind, read text: string, .. }`, then `new_token(";")` simply infers `TokenKind<";">` because of polarity. This should already work today, and if it doesn't then I would consider that a different bug orthogonal to this RFC.

alexmccord:
> If you're going to start inferring some string literals as singletons, why not all of them? Almost every style guide like [Roblox's](https://roblox.github.io/lua-style-guide/#literals) prefers double quotes and it just seems oddly alienating not to support it.

Because that's pathologically horrendous. If you have a type `{"a"}`, you cannot pass it where `{string}` is expected. See this wikipedia page on [variance](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)).

The problem here is this: if you write `local t = {"a"}`, then what type does `t` have: `{string}` or `{"a"}`? Trick question, it's neither. The type of `t` is `{"a" <: 'a <: string}` which is equivalent to `{"a"} <: 'a <: {"a"} | {"a" | "b"} | {"a" | "b" | "c"} | {"a" | "b" | "c" | "d"} | ... | {"a" | "ab"} | {"a" | "ac"} | ...` (ignoring read/write indexers). If `t` is used in such a way that accepts `{"a" | "b"}`, then we have a constraint `'a <: "a" | "b"` and so `t` has the type `{"a" | "b"}`. But due to invariance, `{"a"}` is not a subtype of `{string}`, and because of that, the suggestion of "every literal infers the singleton" is not the same thing that the current type solver does with string literals when it interacts with tables.

Even worse, the approach will not work across functions.
```luau
function make()
  return {"a"}
end

local t = make()
```
With your approach here, you can't even say `local t: {string} = make()` because it's `make` itself that has the type `() -> {"a"}`, and `{"a"}` is not a subtype of `{string}`. This entire class of problems exists due to invariance because you can read (covariant) _and_ write (contravariant) to the table. That `and` keyword means that passing `{"a"}` where `{string}` is expected is a type error because both type must match _exactly_, and evidently it doesn't.

That's why the RFC is proposing _one_ knob, have single quote strings infer the singleton type (to replace the stupid `"foo" :: "foo"` code smell) and have double quotes continue the current behavior as-is.

> My guess is there's some concern with backwards compatibility with the likes of `table.create(50, "hello")` inferring as `{"hello"}` instead of `{string}`. But I'd argue that's preferrable because it's more "correct" even if a little unergonomic.

This is funny to me because in the old type system, I put in a lot of work to fix various unification bugs to force singletons to generalize into their top primitive type, like this:
```luau
function f(t, x)
  if x ~= "x" then
    return
  end

  table.insert(t, x)
end
```
In the old solver, unification of two types were more or less an equality constraint (the story is more complicated than that), so if the subtype or the supertype were a free metavariable, it would bind one to the other and be done. That meant the above snippet would infer the type `f : ({"x"}, unknown) -> ()`, which is obviously utter nonsense and overly restrictive. The correct type is `f : ({string}, unknown) -> ()`.

That means adding a rule in unification where if the subtype is a singleton (distributive over union) and the supertype is a free metavariable, then we generate a new "replaced subtype" where the singletons are replaced by their top primitive types (`true`/`false` -> `boolean`, and any string singletons -> `string`) and then bind the free metavariable to the replaced subtype. This solved the problem, and those bug reports disappeared overnight. It's a shame that the bug reports got replaced by people having a hard time understanding how literal generalization worked, which is unfortunate because it was a necessary evil in that world, and trying to do anything else would break the consistency in the typing algebra. (even I hated the literal generalization rule because of the DX issues it brought to the table)

If we did make every string literals infer the string singleton type in the new solver, we're going back to that world for the second time. The current strategy we have in the new solver is subtly different, which is why doesn't seem to have any glaring DX issues and doesn't clash with invariances. I was very happy when I had the brainwave to suggest that singleton inference should be `"a" <: 'a <: string` for all strings (ditto booleans) in the new solver.

MagmaBurnsV:
If that's the case, I'd rather have literals inferred consistently as a boring `string` instead of the language making an opinionated choice on which style gets special treatment.

These variance issues won't go away if you push them over to someone's personal preference of which quotation marks to use. Someone's bound to do `return {'Bob says "Hi!"'}` and have no clue why defining the function return type as `{string}` doesn't work.

There's got to be a better way to handle the core issue here, which is avoiding the `f("a" :: "a")` syntax that you say looks stupid. For example, higher order generic constraints (if/when they come) could force literal arguments to infer as singletons, e.g. the signature `f<t: singleton>(t) -> t` could infer `t` in `f("a")` as `"a"`.

PR #146, opened 2025-10-29
RFC: Empty Statement Syntax

Aseminaunz:
[Rendered](https://github.com/Aseminaunz/rfcs/blob/rfc-empty-statement-syntax/docs/empty-statement-syntax.md)

Allow empty statements as introduced in Lua 5.2:
```luau
;do ; end;
```


docs/empty-statement-syntax.md
```diff
@@ -0,0 +1,86 @@
...
+```luau
+print("Hello")
+(_G.events["Foo"] :: Event):fire()
+print("World")
+```
+> Error:2: Ambiguous syntax: this looks like an argument list for a function call, but could also be a start of new statement; use ';' to separate statements
+
+To solve it, the tool will likely want to insert a semicolon at the beginning, and, to be sure, end of the statement. Doing so would also be dangerous, as a syntax error would occur if the statement already had any surrounding semicolons or no statement was preceding it in the block. Tools need to check the context to skip introducing a semicolon where it'd cause an error, adding unnecessary and probably unexpected complexity. If empty statements were allowed, tools could safely surround statements with semicolons, guarding against ambiguity without having to worry about context.
+
+This is especially useful for simple "insert-replace" tools that don't perform a syntactic analysis of the code, such as search-and-replace operations in code editors, where currently each replacement would need review to avoid syntax errors.
```
aatxe, line 45: The tooling argument is compelling, but this argument is not. Using find-and-replace to make your code a horrible mess sounds like an antipattern, and the fact that you today might get parsing errors to correct intentionally actually sounds much better in comparison.

docs/empty-statement-syntax.md
```diff
@@ -0,0 +1,86 @@
...
+block ::= {stat | ';'} [laststat {';'}]
+```
+
+As already is the case, empty statements have no impact in generated bytecode.
+
+## Drawbacks
+
+Previously erroring code would now be valid, though uses of Luau relying on this error would be exceptionally rare (and admittedly odd).
+
+Luau is designed to be backwards compatible with Lua 5.1, and adopting this syntax would diverge from that compatibility. However, Luau already diverges from Lua 5.1's syntax, and this change would align Luau more closely with modern Lua.
```
aatxe: Supporting things not in Lua 5.1 does not affect our backwards compatibility story by definition. Backwards compatibility means that you can run valid Lua 5.1 programs. This is not a change that has any impact on backwards compatibility.

docs/empty-statement-syntax.md
```diff
@@ -0,0 +1,86 @@
...
+More generally, similar "void" separators could also be allowed for other list-like syntaxes, such as generic lists, parameters, table constructors, or expression lists:
+
+```luau
+local function test<,,,,A,, B,, C...,,,,>(,,a: A,,,, b: B,, ...: C...,,): (,number,,, string,)
+	local _ = {1,,, 2;;,;, 3;;;;,}
+	return ,,42,,,, "hello",,;
+end
+```
+
+This syntax however, unlike with statement blocks, is considerably confusing and unusual, has an unclear purpose, and is very likely a mistake, with little to no practical use cases. In table constructors, a single trailing separator is already allowed.
```
aatxe, line 86: Why are these obviously a mistake, but the proposed syntax isn't obviously a mistake? That feels like it needs to be articulated much more crisply.

Aseminaunz: `f(1, , 3)` isn't clear as to whether the intention was to do `f(1, nil, 3)` or `f(1, 3)`. In my opinion, the most likely thing is the user inserted a comma as placeholder and forgot to insert an actual value later.

docs/empty-statement-syntax.md
Aseminaunz: Writing prettier or uglier code with more or fewer semicolons is up to the user. If they insert unnecessary semicolons, they can manually remove them where they see fit or pass the code through a formatter. "Ugly syntax" that is not ambiguous should not be a parsing error, regardless of how ugly it is.

Aseminaunz: If unnecessary semicolons are too horrendous to pass normally, Luau's linter could add a "DuplicatedSemicolon" lint when two or more semicolons are used in a row without whitespace in between. But this is merely style and imo should be up to an external tool. And hey, maybe someone wants to use double semicolons to be extra explicit about a no-op. They should be able to without errors or lint warnings.

docs/empty-statement-syntax.md
Aseminaunz: Unlike with statements, an "empty element" in these other list syntaxes does not make sense and needs to either be `nil` or skipped entirely. I don't believe I can find any benefits to allow "skipping" in those other contexts, and silently defaulting to either of those two solutions would do more harm than good.

docs/empty-statement-syntax.md
Aseminaunz: To make a clearer and more reasonable example: in my code, I might have a specific comment that I want to replace with a series of debugging statements when built for debug. For this, I use a simple CLI tool that replaces occurrences of that comment with my statements. If I want to avoid ambiguity, I'd need to either add a preceding semicolon where appropriate to the comments, or use a more sophisticated and unnecessary solution that performs a syntactic analysis. Both unideal for something theoretically dead simple.

docs/empty-statement-syntax.md
aatxe: I'm not asking you to convince me here, I'm asking you to revise the RFC to be more convincing.

Aseminaunz: Just stating it here for documentation and amending with any further observations when I can

docs/empty-statement-syntax.md
```diff
@@ -58,9 +58,9 @@ As already is the case, empty statements have no impact in generated bytecode.
 
 ## Drawbacks
 
-Previously erroring code would now be valid, though uses of Luau relying on this error would be exceptionally rare (and admittedly odd).
+Previously erroring code would now parse. This is not considered to be a problem, as such programs were never valid in the first place.
 
-Luau is designed to be backwards compatible with Lua 5.1, and adopting this syntax would diverge from that compatibility. However, Luau already diverges from Lua 5.1's syntax, and this change would align Luau more closely with modern Lua.
+This syntax, when used, is not backwards-compatible. However, it can easily be removed without changing the meaning of code (which isn't the case for many other syntax features introduced by Luau).
```
aatxe, line 63: We do not care about running Luau code in Lua 5.1 VMs. That's not a goal of the project.

Aseminaunz: That has already been removed.

docs/empty-statement-syntax.md
```diff
@@ -0,0 +1,86 @@
...
+A tool may want to replace the statement `GLOBAL_EVENT "Foo"`, with for example `(_G.events["Foo"] :: Event):fire()`. Inserting it naively would cause an error due to the ambiguity of the parentheses: the first two lines are a single statement. Luau currently errors with statements like this, refusing to parse them.
+
+```luau
+print("Hello")
+(_G.events["Foo"] :: Event):fire()
+print("World")
+```
+> Error:2: Ambiguous syntax: this looks like an argument list for a function call, but could also be a start of new statement; use ';' to separate statements
+
+To solve it, the tool will likely want to insert a semicolon at the beginning, and, to be sure, end of the statement. Doing so would also be dangerous, as a syntax error would occur if the statement already had any surrounding semicolons or no statement was preceding it in the block. Tools need to check the context to skip introducing a semicolon where it'd cause an error, adding unnecessary and probably unexpected complexity. If empty statements were allowed, tools could safely surround statements with semicolons, guarding against ambiguity without having to worry about context.
```
alexmccord, line 43: > the tool will likely want to insert a semicolon at the beginning, and, to be sure, end of the statement

Beginning _and_ end? Why? It's trivial to decide whether a semicolon is necessary by literally just stringifying the next statement, and then see if that string starts with `(`. If it starts with `(`, then add a `;` to the current line, then `\n` and then add the indentation and finally, the statement string in question.

Something like this:
```luau
function print_block(stmts)
  local result = table.create(#stmts)

  for i, stmt in stmts do
    table.insert(result, print_stmt(stmt))
  end

  for i, stmt_str in result do
    local next_str = result[i + 1]
    if next_str == nil then
      break
    end

    local ends_with_semi_colon = string.byte(stmt_str, -1) == string.byte(";")
    local starts_with_open_paren = string.byte(next_str, 1) == string.byte("(")

    if not ends_with_semi_colon and starts_with_open_paren then
      result[i] ..= ";"
    end
  end

  return table.concat(result, "\n")
end
```

You can avoid an intermediate table if you wish, but I don't care. This argument is weak. Obviously, the specific code I've given you will still generate `;` even when it isn't strictly necessary, but you can just replace `ends_with_semi_colon` with a function that takes some `stmt` and returns a boolean indicating whether the given `stmt` fragment ends with an expression and without a semi-colon. Still trivial.

InfraredGodYT:
I'm just going through RFC's I have thoughts on at the moment, and I know this one is 6 months old, but I'm not really sure if the functionality of empty statements is particularly useful, especially in a language like Luau.

PR #147, opened 2025-11-04
Allow attributes to be used on types, variables, fields, and classes

gaymeowing:
[Rendered](https://github.com/gaymeowing/rfcs/blob/attributes-everywhere/docs/syntax-attributes-for-types-variables-fields-classes.md)

### DIFF
--- a/docs/syntax-attributes-for-types-variables-fields-classes.md
+++ b/docs/syntax-attributes-for-types-variables-fields-classes.md
@@ -138,7 +138,7 @@ local puppy = Puppy {
 }
 ```
 
-Class fields work the same as [Table Fields](#table-fields). Attributes placed before a `public` field declaration are attached to that class member.
+Class fields work the same as [Table Fields](#table-fields). Attributes placed before a `public` field declaration are attached to that class member:
 
 ```luau
 class PetSounds
--- a/docs/syntax-attributes-for-types-variables-fields-classes.md
+++ b/docs/syntax-attributes-for-types-variables-fields-classes.md
@@ -66,6 +66,7 @@ module.parrot = "cracker, now"
 return table.freeze(pet_sounds)
 ```
 
+#### Table Fields
 If the value of a field has attributes, those will be merged with the attributes defined on the field.
 With the attributes on the field having priority over the attributes on the value.
 For example: if both the value and the field have a `@deprecated` attribute, the `@deprecated` attribute on the value will be ignored.
@@ -137,7 +138,7 @@ local puppy = Puppy {
 }
 ```
 
-Class fields work the same as [Table Fields](#tables). Attributes placed before a `public` field declaration are attached to that class member.
+Class fields work the same as [Table Fields](#table-fields). Attributes placed before a `public` field declaration are attached to that class member.
 
 ```luau
 class PetSounds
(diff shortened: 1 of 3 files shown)


gaymeowing:
Didn't bother adding ebnf for each bit of syntax where attributes will now be allowed, as its the same as how function attributes are applied except adjusted for each bit of syntax this RFC adds attributes to.

I think tables might need some ebnf defined (because of how attributes are merged)? Although unsure how to do that, so if someone wants to add that in it'd be appreciated.

itsmath:
Is there a difference between these?

```luau
local on_type: @deprecated T

@deprecated
local on_variable
```

Seems like the answer is _yes_? For example:

```luau
type deprecate<T> = @deprecated T

@deprecated
type deprecated<T> = T

local use_deprecate: deprecate<number> = 67 -- This is fine
local use_deprecated: deprecated<number> = 67 -- This is not fine, you're using a deprecated type

print(use_deprecate) -- This is not fine, you're using a deprecated variable
```

I think this is something you need to specify in the RFC. I feel like this will require a major redesign on how the `@deprecated` attribute works

<sub>Also, what about attributes on loops? Something like an `@unroll` attribute would be pretty useful</sub>

MagmaBurnsV:
I think instead of adding attribute syntax to virtually every language construct in one fell swoop, we should leave grammar incorporation to separate RFCs that actually introduce useable attributes for a specific construct, e.g. an `@unroll` attribute for for-loops would make attributes before the `for` keyword valid.

This way, we won't accidently introduce attribute syntax in places where we'll never end up supporting any attributes, and discussion about semantics like merging deprecation messages can be specialized to the attribute being considered.

In this RFC's case, I think it should just focus on adding `@deprecation` support for local variables and types.

docs/syntax-attributes-for-types-variables-fields-classes.md
```diff
@@ -0,0 +1,124 @@
...
+
+Consumer:
+
+```luau
+-- No lint occurs on the import
+local module = require("@module")
+
+-- DeprecatedApi: Member 'puppy' is deprecated, use 'dog' instead Luau(22)
+print(module.puppy)
+```
```
gaymeowing, line 52: Will probably end up not having it work like this, as this is just sorta weird behavior?
But variables I would like to keep, as I could imagine it being useful for those who were not doing best practices before.

Like in the case of someone writing their entire game in one script, where they want to deprecate a variable because they split out its logic to a different module. With find and replace being time consuming when they're likely to just be rewriting most of everything. So having the `@deprecated` attribute on the variable would work as a reminder to not use that variable as they refactor.

But this is literally only good for this one somewhat niche usecase, so perhaps ditching variable support would be a good idea?

docs/syntax-attributes-for-types-variables-fields-classes.md
```diff
@@ -0,0 +1,162 @@
...
+Note: With this RFC the internals for attributes will be able to state they can be applied to only certain bits of syntax, so one can't apply [`@native`](./syntax-attribute-functions-native.md) to a type, variable, class, or field and have it work. Although this RFC does not state how this would be done in Luau's internals.
+
+The following list proposes how attributes should be attached to each bit of syntax in Luau (except functions and comments):
+
+### Variables
+
+This exists as it could be useful for if for instance a constant is defined as a local variable and then exported:
+
+```luau
+@deprecated { use = "dog"}
```
PhoenixWhitefire, line 23: Shouldn't it be `@[deprecated{ use = "dog" }]` to be consistent?

gaymeowing: It just looks nicer to do it this way + square brackets are only required for multiple attributes

PR #150, opened 2025-11-20
RFC: Multiline string interpolation

jadarve:
[Rendered](https://github.com/jadarve/luau-rfcs/blob/multiline_string_interpolation/docs/syntax-multiline-string-interpolation.md).
See https://github.com/luau-lang/luau/issues/2095


jackdotink:
its possible and good to use \ and \z for multiline interpolated strings today

PR #152, opened 2025-11-27
RFC: math.truncate

karl-police:
``math.truncate`` it cuts off digits after the decimal point, without having to use ``math.floor`` or ``math.clamp``.
🤔

[Render](https://github.com/karl-police/rfcs/blob/patch-4/docs/function-math-truncate.md)

### DIFF
--- a/docs/function-math-truncate.md
+++ b/docs/function-math-truncate.md
@@ -8,7 +8,7 @@ A method that returns the integer part of a number by removing any fractional di
 
 ``math.floor`` does not give the same effect as ``Math.trunc()`` from JavaScript, for instance. But ``math.floor`` and ``math.ceil`` can be used together, to re-produce the same result that ``Math.trunc()`` would give, but a function like ``math.truncate`` directly as a built-in would be more straightforward.
 
-For instance:
+For instance, negative numbers:
 ```lua
 print( math.floor(-0.005) )
 -- Returns: -1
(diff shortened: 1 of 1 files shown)


aatxe:
> Since that one returns a tuple, What happens to the the second tuple in memory? Does it just get reliably garbage collected?

Luau does not have tuples. `modf` returns two numbers on the stack, and if they are not used, they are not used. There's no heap allocation associated with returning two numbers.

karl-police:
> > Since that one returns a tuple, What happens to the the second tuple in memory? Does it just get reliably garbage collected?
> 
> Luau does not have tuples. `modf` returns two numbers on the stack, and if they are not used, they are not used. There's no heap allocation associated with returning two numbers.

Ah, I see. Tuples to me are just a list of values, not sure if it has any other meanings.


> You can already achieve number truncation with [`math.modf`](https://www.lua.org/manual/5.1/manual.html#pdf-math.modf).

Is``.modf`` more expensive than ``math.ceil`` and or ``math.floor``?

vegorov-rbx:
Both motivation and design of this RFC lack clarity expected from a proposal.
The feature that I can only infer that is being proposed here does it worse than the language it takes inspiration from.
We are not looking to add this to Luau in this form.

Something to note is that feature requests don't have to be in the form of an RFC, you can suggest features in https://github.com/luau-lang/luau

PR #154, opened 2025-12-04
Implement sorting in index.md

ari-party:
This PR:
- Implements the suffixing of `(Implemented)` to the links of RFCs that contain `**Status**: Implemented`
- Implements guaranteed alphabetical sorting, and sorting based on implementation status (implemented RFCs come last)

You can preview this [here](https://ari-party.github.io/luau-lang-rfcs/).
Links don't work on this preview due to root URL differences.

The reason for these changes is the matter of RFC relevance, and to add a form of categorization.

(Should `(Implemented)` be in the link or outside of it?)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,6 +1,22 @@
 # RFCs
 
-{% assign sortedPages = site.pages | sort: 'title' %}
+{% assign allPages = site.pages %}
+{% assign implementedPages = '' | split: '' %}
+{% assign notImplementedPages = '' | split: '' %}
+
+{% for page in allPages %}
+{% assign ext = page.name | split:'.' | last %}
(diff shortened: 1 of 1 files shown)


ari-party:
> Sorting should be in order of most recently accepted to least recently accepted. Alphabetical ordering is just annoying and doesn't actually help us find recent RFCs. Like every time I need to check an RFC I go to the PRs tab and hope the branch wasn't deleted. 

I'll look into that. Afaik Jekyll doesn't support that by default, would require some git integration—or adding implementation dates as metadata to each RFC

ari-party:
I unfortunately do not have the time for this right now, and won't for a while. I'll hopefully come back to do this, later this year, but only if Arseny hasn't done something already!

PR #160, opened 2025-12-23
Align tonumber with current Luau number syntax

gaymeowing:
[Rendered](https://github.com/gaymeowing/rfcs/blob/align-tonumber-with-number-syntax/docs/align-tonumber.md)

### DIFF
--- a/docs/align-tonumber.md
+++ b/docs/align-tonumber.md
@@ -7,9 +7,16 @@ allowing for binary literals and underscores in number strings passed to `tonumb
 
 ## Motivation
 
-Currently `tonumber` supports regular numbers and hexadecimals, making `tonumber` support half of luaus number syntax currently.
-This is annoying for users as it means if they want to support `_` in numbers, if they are say making a math expression evaluator they cannot do so without doing some jank things with buffers and the bit32 library.
-Same goes for binary literals, despite not being nearly as useful.
+Currently `tonumber` supports regular numbers and hexadecimals, via the `base` argument and adding `0x` as a prefix to the string being passed to `tonumber`.
+Given `tonumber` supports hexadecimal via `0x`:
+```luau
+print(tonumnber("0xFFFFFF")) --> 16777215
+```
+
+Its currently odd that it doesn't support binary literals via `0b` like luau currently does for numbers.
+Additionally allowing underscores in numbers passed to `tonumber` is fairly easy to implement, and is nice syntax sugar for making parsers that handle numbers in luau.
+Alongside the fact allowing underscores would also make the behavior of `tonumber` simpler for users. As `tonumber` would then parse numbers in the exact same way as Luau,
+rather than how it is now with an odd caveat for hexadecimals.
 
 ## Design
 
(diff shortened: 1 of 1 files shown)


jackdotink:
imo the backwards compatibility loss isnt worth the change here

alicesaidhi:
what backwards compatibility will be lost here?

gaymeowing:
> imo the backwards compatibility loss isnt worth the change here

I haven't found any code that this would break, and majority of the cases where I've seen people ensuring a string is valid ASCII uses the string library to do so and not `tonumber`. Albeit my search wasn't that comprensive, but if its had any significance I would expect it to appear in a github search.
https://github.com/search?q=(language%3ALua+OR+language%3ALuau)+is+ascii+tonumber&type=code

jackdotink:
okay sure, its still breaking backwards compatibility without adding significant value - use gsub first if you really want this. as is frequently the case, the solution to your problem is not more language features, but is instead writing more code

alicesaidhi:
what backwards compat is it breaking?

gaymeowing:
> that code is not valid today - you have to pass a second argument to tonumber to make it parse hexadecimal. just because you cant find code that would break doesnt mean that a case this simple doesnt exist somewhere

what do you mean i ran that with the luau cli

docs/align-tonumber.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Summary
+
+This RFC proposes aligning `tonumber` with Luau's current number syntax,
+allowing for binary literals and underscores in number strings passed to `tonumber`.
+
+## Motivation
+
+Currently `tonumber` supports regular numbers and hexadecimals, making `tonumber` support half of luaus number syntax currently.
+This is annoying for users as it means if they want to support `_` in numbers, if they are say making a math expression evaluator they cannot do so without doing some jank things with buffers and the bit32 library.
```
alexmccord: Not too annoying.
```luau
local function my_tonumber(s: string, base: number?)
  s = string.gsub(s, "_", "")
  if base == nil or base == 2 then
    local matched = string.match(s, "0b(.+)")
    if matched ~= nil then
      s = matched
      base = 2
    end
  end
  return tonumber(s, base)
end
```

gaymeowing:
> It is pretty common in programming languages to have literals in the source code and runtime number parsing differences.
> 
> `tonumber` relies on the underlying host implementation of double number parsing and supports many forms that are not valid as Luau number literals (nan, inf, hex floats, decimal separators if locale is changed from C). This RFC doesn't achieve alignment. Since all C99 features have to be supported for backwards compatibility, we will have to reimplement number parsing to preserve all C99 features to add additional forms. Approach taken in the current parser with temporary string construction while filtering '_' will negatively impact runtime performance. Even if manually implemented, many of the valid C99 strtod/strtoul strings will remain to be invalid literals in Luau.

Would it be possible as an alternative to this, to instead support passing a buffer instead of a string to `tonumber`?
As the parser I'm currently working on benefits a lot from using a buffer (in native, although table.concat is marginally faster with the interpreter) for converting number strings with underscores into a number string without underscores to then be passed on to `tonumber`. And if `tonumber` could take in buffers I could avoid allocating a temporary string using `buffer.readstring`.

PR #165, opened 2025-12-27
[RFC] Amend Global Type Functions to Accomodate Read & Write Keys

hardlyardi:
[Rendered.](https://github.com/hardlyardi/rfcs/blob/amend-global-type-functions-to-accomodate-readwrite-indexing/docs/amend-global-type-functions-to-accomodate-readwrite-indexing.md)

### DIFF
--- a/docs/amend-global-type-functions-to-accomodate-readwrite-indexing.md
+++ b/docs/amend-global-type-functions-to-accomodate-readwrite-indexing.md
@@ -1,0 +1,68 @@
+# Amend Global Type Functions to Accomodate Read/Write Indexing
+
+This RFC supersedes [PR 162](https://github.com/hardlyardi/rfcs/blob/newindex-and-rawset-type-functions/docs/type-functions-newindex-and-rawset.md)
+
+## Summary
+
+Amend some global type functions (`keyof`, `rawkeyof`, `index`, `rawget`) to take in an optional trailing argument,
+limiting their use to the "read" or "write" types of a table.
+
+## Motivation
+
+Currently, there are no built-in type functions with functionality to access the read/write variants of keys in a
+table-like type. Additionally, `rawget` and `rawset` are currently typed unsafely, so this RFC improves the soundness
+of existing globals.
+
+## Design
+
+The following existing type functions:
+
+- `keyof<TableLike>`
+- `rawkeyof<TableLike>`
+- `index<TableLike, Key>`
+- `rawget<TableLike, Key>`
+
+Do not support table-like types with differing read and write types; including, but not limited to
+[read-only properties](https://rfcs.luau.org/property-readonly.html). This RFC proposes adding an optional trailing
+argument, a string literal type, to each. The current type expression `keyof<TableLike>` will become equivalent to the
+expression `keyof<TableLike, "read"> | keyof<TableLike, "write">`.
+
+Similarly for other global type functions:
+
+```luau
+type Meow = {
+    read purr: true,
+}
+-- Write Key 'purr' does not exist in type
+type Mrrp = index<Meow, "purr", "write">
+-- 'true'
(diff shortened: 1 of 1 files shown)


docs/amend-global-type-functions-to-accomodate-readwrite-indexing.md
```diff
@@ -0,0 +1,68 @@
...
+
+- `keyof<TableLike>`
+- `rawkeyof<TableLike>`
+- `index<TableLike, Key>`
+- `rawget<TableLike, Key>`
+
+Do not support table-like types with differing read and write types; including, but not limited to
+[read-only properties](https://rfcs.luau.org/property-readonly.html). This RFC proposes adding an optional trailing
+argument, a string literal type, to each. The current type expression `keyof<TableLike>` will become equivalent to the
+expression `keyof<TableLike, "read"> | keyof<TableLike, "write">`.
```
alexmccord, line 28: Can't you just do composition? `keyof<readonly<T>>` and `keyof<writeonly<T>>` where `readonly` removes the `write` modifier, and conversely `writeonly` with `read` modifier.

docs/amend-global-type-functions-to-accomodate-readwrite-indexing.md
```diff
@@ -0,0 +1,68 @@
...
+Similarly for other global type functions:
+
+```luau
+type Meow = {
+    read purr: true,
+}
+-- Write Key 'purr' does not exist in type
+type Mrrp = index<Meow, "purr", "write">
+-- 'true'
+type Mrrp = index<Meow, "purr", "read">
```
alexmccord, line 39:
```luau
type T = index<{ write p: number }, "p">
```
Ugh, `index<{ write p: number }, "p">` getting normalized to `number` is without doubt a bug. `index` and `rawget` are intended to only work on the `read` part!

docs/amend-global-type-functions-to-accomodate-readwrite-indexing.md
hardlyardi: how would that work with extern types?

hardlyardi:
> This is actually a bug report, not an RFC, I think. The type system generates types like `index<T, K>` as part of deferring any unsolved type variables in the new constraint solver, and the fact that `index<T, K>` can normalize to the `write` part of the property means the type system will not derive correct types.

This is definitely an RFC. Getting `write` indices is useful! That said, I'm aware there are a few bugs w.r.t. table-like-types and indexing, not sure how it all plays out yet.

docs/amend-global-type-functions-to-accomodate-readwrite-indexing.md
MagmaBurnsV: Generic type functions sound like the better option here. I don't think there should be any support for extern types since trying to normalize them seems impossible, e.g. `readonly<vector>` would have to resolve to something wacky like `vector & ~{write x: number, write y: number, write z: number}` to work. If someone really needs that then they can just use UDTFs.

hardlyardi: it isn't unrealistic to normalize the read/write access of extern types, it is unrealistic to normalize the extern types themselves. That's what I'm addressing.

Additionally, there are many examples of the dissonance between extern types and table types being harmful to ease-of-use, as well as the development of the language. I think if it can be avoided, you shouldn't need to use a UDTF to get the `read`/`write` keys of an extern type. That's a very reasonable use-case for keyof IMO.

aatxe:
The particular part Alex was commenting on, that `type T = index<{ write p: number }, "p">` doesn't give `never` or an error, is definitely a bug, but yeah, the main part of this RFC is a feature request (the ability to _ask_ for the write-types).

PR #166, opened 2026-01-09
Const Keyword for Local Declarations

bradsharp:
This document introduces the `const` keyword for local variable bindings, explaining its syntax, semantics, and potential drawbacks.

[Rendered](https://github.com/luau-lang/rfcs/blob/rfc-const-keyword/docs/const-keyword.md)

### DIFF
--- a/docs/const-keyword.md
+++ b/docs/const-keyword.md
@@ -6,9 +6,33 @@ Introduce a `const` keyword for local variable bindings that prevents reassignme
 
 ## Motivation
 
-When writing code, many variables are declared once and never intended to change such as module imports, function declarations, service references, and configuration objects. However, there is no way to prevent them from being rebound, which can lead to accidental rebinding, more time spent validating variables do not change during code reviews, and reducing the effectiveness of certain compiler / typechecker optimizations
+When introducing the [`export`](https://github.com/luau-lang/rfcs/pull/42) keyword we will need to introduce the concept of `const`-ness to the language. Modules with exports return a frozen table of values; however, the variables used to declare those exports are not themselves immutable. This makes it possible to write code such as the following:
 
-Additionally, with the proposal of the [export keyword](https://github.com/luau-lang/rfcs/pull/42), variables declared with it are frozen in the module exports table and therefore expected to refer to stable values. Without const-ness they can accidentally be redefined in confusing ways, within the module definition.
+```luau
+export foo = 5
+
+export function increment()
+	foo += 1
+end
+```
+
+This results in different values for `foo` depending on where it is accessed from:
+
+```luau
+-- Inside the module
+print(foo) --> 5
+increment()
+print(foo) --> 6
+
+-- Outside the module
+print(module.foo) --> 5
+module.increment()
+print(module.foo) --> 5
+```
+
+To address this, we propose that all exported values are `const` by default, preventing reassignment within the module and ensuring that exported bindings represent stable values.
+
+However, this is not the only use-case for `const`. A `const` binding provides a clear guarantee that a name will always refer to the same value for the lifetime of its scope. This property is useful beyond module exports: it makes code easier to reason about, avoids accidental rebinding in closures or long-lived scopes, and allows tools and the typechecker to treat such bindings as stable references rather than values that may change over time. Since `export` already requires this guarantee to behave correctly, exposing `const` as a general language feature avoids introducing special-case rules and provides a simple, consistent way to express immutability of bindings where it is desired.
 
 ## Design
 ### Syntax
--- a/docs/const-keyword.md
+++ b/docs/const-keyword.md
@@ -1,0 +1,128 @@
+# Const Keyword
+
+## Summary
+
+Introduce a `const` keyword for local variable bindings that prevents reassignment after initialization. A const declaration makes the binding immutable, not the value, so tables and other mutable objects remain mutable through the binding.
+
+## Motivation
+
+When writing code, many variables are declared once and never intended to change such as module imports, function declarations, service references, and configuration objects. However, there is no way to prevent them from being rebound, which can lead to accidental rebinding, more time spent validating variables do not change during code reviews, and reducing the effectiveness of certain compiler / typechecker optimizations
+
+Additionally, with the proposal of the [export keyword](https://github.com/luau-lang/rfcs/pull/42), variables declared with it are frozen in the module exports table and therefore expected to refer to stable values. Without const-ness they can accidentally be redefined in confusing ways, within the module definition.
+
+## Design
+### Syntax
+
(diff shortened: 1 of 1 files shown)


docs/const-keyword.md
```diff
@@ -0,0 +1,152 @@
...
+end
+```
+
+### Semantics
+
+A `const` variable is equivalent to a variable declared with `local` however `const` variables cannot be reassigned after they are initialized:
+
+```luau
+const x = 1
+x = 2 -- error
```
SPY, line 74: What kind of error? Static(Parsing Error) or dynamic?
Difference is more obvious inside a function definition.
```
const a = 42
function foo() -- 1
  a = 43
end
foo() -- 2
```
Should we see an error at parsing of 1 or at the call of 2?
JavaScript picks option 2, but honestly I'm not aware of advantage of it. So I would prefer option 1.

docs/const-keyword.md
```diff
@@ -0,0 +1,152 @@
...
+```luau
+const x = 1
+const function f()
+	x = 2 -- error
+end
+```
+
+### Grammar / Parsing
+
+`const` becomes a keyword in the `local` declaration position. If `const` is currently allowed as an identifier, this is a source-breaking change only for code that uses const as a local variable name in positions where a local declaration is parsed. (See Drawbacks.)
```
SPY: I don't think we need to color `const` as a keyword for now(as we do for `continue`). `<ident> <ident>` construction is not syntactically correct, so `const <ident>` shouldn't be present in existing code. Also, `local const = 42` will be a correct construction.

docs/const-keyword.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Drawbacks
+
+- Keyword and compatibility surface: if `const` is currently used as an identifier in existing code, reserving it as a keyword can be source-breaking in some contexts.
```
SPY: As I mentioned above it is not an issue right now.
But it can be in conflict with developing proposals, like destructive assignment.
Theoretically you can have something like that in you code now:
```
const {a, b}
```
and it sould be treated as a call to `const` with a single param.
Assuming destructive assignment will look like 
```
const {a, b} = x
```
It can be an issue

bradsharp:
> The motivation is very weak

`export` is the primary motivation for introducing `const`-ness to the language. When we return from a module with exports we freeze the result (see why [here](https://github.com/luau-lang/rfcs/pull/42#issuecomment-3386556127). This leads to an issue where the variables can be reassigned but the return value itself doesn't change. If we're going to introduce `const`-ness to the language for `export` then it seems reasonable to expose it for general use at the same time.

I've added more details about this to the RFC.

> the drawbacks of breaking backwards compatibility are... bad.

Instead of reserving `const` as a reserved keyword it can be introduced as a contextual keyword that is only valid in positions that `local` is also valid. This makes the change compatible with existing code as `const foo = 5` would fail to parse today.

I've covered this in the RFC also.

docs/const-keyword.md
bradsharp: I updated the RFC to reflect this, going to resolve

docs/const-keyword.md
bradsharp: As `const` variables must always be initialized `const {a, b}` would be invalid (and therefore treated as a function call) while `const {a, b} = x` remains unambiguous as this is invalid syntax today. The one downside I see if that this could become a gotcha, so we might want to consider a linter warning for `const {a, b}`.

docs/const-keyword.md
bradsharp: I think it can be implemented as a parsing error but we may want a runtime error as a backstop for cases like `loadstring`. I would expect the error to be on the line of the actual assignment, in this case `a = 43`.

bradsharp: As an aside `JavaScript` likely picks (2) because it's the line that leads to `a` being mutated and without it the value remains constant. For sanity, I think we should stick to erroring on `a = 43` even if the code is unreachable.

vegorov-rbx: JS goes crazier here:
```js
const a = 42;

function bar(overrides) {
  with (overrides) {
    let foo = function() {
      console.log(a);
      a = 43;
      console.log(a);
    }

    foo();
  }
}

bar({ a: 50 }); // prints 50, 43
bar({ b: 50 }); // prints 42 and throws TypeError: Assignment to constant variable.

```

docs/const-keyword.md
jackhexed: @bradsharp While `const {a, b] = x` is not ambiguous, it is still not reasonably parsable.

jackhexed: It almost makes one want for a "Do not add destructuring syntax" RFC

docs/const-keyword.md
```diff
@@ -0,0 +1,152 @@
...
+const t = { a = 1 }
+```
+
+Typed declarations remain valid:
+
+```luau
+const x: number = 5
+const t: { a: number } = { a = 1 }
+```
+Multi-assignment is also supported:
```
SPY, line 54: Probably worth mentioning vararg multi assignment explicitly.

it should be fine from runtime point of view to have something like
```
function f() return 1 end

const a, b = f()
```
Because f can return multiple values in last position(same with `...`). In that case `b` should be initialized with `nil`. But typechecker probably should complain here.

SPY (approved):
I think RFC in a good shape and can be merged.

SPY:
Merged as a dependency for #42

MagmaBurnsV:
I must strongly object to merging this RFC.

First, I think it is dishonest to say this is needed as a "dependency" for the export-by-value RFC. This is not true and even contradicts the RFC's stated motivation:

> Since `export` already requires this guarantee to behave correctly, exposing `const` as a general language feature avoids introducing special-case rules and provides a simple, consistent way to express immutability of bindings where it is desired.

Nothing about exporting *needs* this, and it shouldn't be framed like that.

Second, as I've already said before, this motivation is very weak as its only substantial rationale is to close a feature gap caused by the current export-by-value semantics. This can instead be solved by adopting different export semantics that don't cause feature gaps, such as my proposed alternative semantics.

Even if the team is convinced that current export semantics are the best for static optimizations (I disagree, my alternative can also obtain these optimizations without introducing const-ness), that still doesn't justify implementing `const` *before* export-by-value has even been accepted.

alexmccord:
Rebinding `const` can all be done at parse time, there's no need for a runtime error to rebind `const`. `loadstring` can only affect globals, which `const` isn't, it occupies a register (modulo `const`-of-`const` elimination) which do not carry into `loadstring`. See `local x = 5 assert(loadstring("x = 7; print(x)"))() print(x)` which prints `7` then `5`.

PR #170, opened 2026-01-29
RFC: Interpolated string function calls

rofrankel:
## Summary

Allow calling functions with interpolated strings without parentheses, enabling DSL patterns like structured logging, SQL escaping, and HTML templating. The call desugars into two arguments: the original template string (with interpolation expressions intact) and a table of evaluated values.

```luau
log:Info `Hello {name}`
-- Desugars to: log:Info(string.interpolatedstring.create({"Hello ", ""}, {name}))
```

Two new core library functions are also introduced:
* `string.interpolate` executes the default string interpolation algorithm for such a template string, and
* `string.interpolatedstring.create`, which creates a wrapping table and ascribes to it a metatable that provides a few useful operator overloads.

## Motivation

The [string interpolation RFC](https://github.com/luau-lang/rfcs/blob/master/docs/syntax-string-interpolation.md) noted the restriction on parentheses-free calls was "likely temporary while we work through string interpolation DSLs." This RFC proposes lifting that restriction with semantics designed for structured logging and similar use cases.

## Key design decisions

- Parentheses-free interpolated string calls desugar to an interpolated string object
- Any expression valid in a regular interpolated string is also valid here (no restrictions)
- Currying pattern enables passing additional context without grammar changes

### DIFF
--- a/docs/syntax-interpolated-string-function-calls.md
+++ b/docs/syntax-interpolated-string-function-calls.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-Allow calling functions with interpolated string literals without parentheses, optionally followed by a context table literal argument. This enables domain-specific language patterns like structured logging where the template, interpolation values, and optional additional context are all passed to the function.
+Allow calling functions with interpolated string literals without parentheses. The call is desugared into a function call with three arguments: a format string with `%s` placeholders, a table of the evaluated interpolation values, and a table of the original expression source texts. This enables domain-specific language patterns like structured logging, SQL escaping, and HTML templating.
 
 ## Motivation
 
@@ -22,69 +22,79 @@ print `Hello {name}`  -- currently a parse error
 
 The [string interpolation RFC](https://github.com/luau-lang/rfcs/blob/master/docs/syntax-string-interpolation.md) noted this restriction was "likely temporary while we work through string interpolation DSLs."
 
-This proposal lifts that restriction and extends it further to support an optional trailing table literal, enabling powerful DSL patterns. The primary motivating use case is structured logging, where it's valuable to capture both the formatted message and the template with its interpolation values for machine processing:
+This proposal lifts that restriction with semantics that decompose the interpolated string into its constituent parts, passing them to the called function. This enables the function to process the template, values, and expression metadata however it sees fit, without needing to implement its own string parser.
+
+### Structured logging
+
+The primary motivating use case is structured logging, where it is valuable to capture both the formatted message template and its interpolation values for machine processing:
 
 ```luau
-local name = "Alice"
-local log = require("@rbx/logging")
+local a = 1
+local function double(x: number)
+  return x * 2
+end
 
--- Calls log:Info with three arguments:
---   "Hello Alice" (formatted string)
---   "Hello {name}" (template)
---   {name = "Alice"} (interpolation values)
-log:Info `Hello {name}`
-
--- Calls log:Info with four arguments, adding structured context:
---   "Hello Alice" (formatted string)
---   "Hello {name}" (template)
---   {name = "Alice"} (interpolation values)
---   {userId = 12345, region = "us-east"} (additional context)
-log:Info `Hello {name}`, {userId = 12345, region = "us-east"}
+-- Desugars to `log:Info("The double of %s is %s", {a, double(a)}, {"a", "double(a)"})`
+-- Note that the second argument evalutes to {1, 2} at runtime, but not at the desugaring step.
+log:Info `The double of {a} is {double(a)}`
 ```
 
-This pattern is common in structured logging, where preserving the template enables aggregating logs by message pattern while the interpolation values and context provide searchable structured data for platforms like Splunk, Datadog, or Elasticsearch.
+The template string enables aggregating logs by message pattern (e.g. grouping all "The double of %s is %s" messages), while the values and expression names provide searchable structured data for platforms like Splunk, Datadog, or Elasticsearch.
 
-Without this feature, there are two alternatives:
+Without this feature, developers must either manually construct all arguments (tedious and error-prone) or use a logging library that implements its own template parsing at runtime (duplicating language functionality).
 
-1. Manually construct all arguments:
+### SQL escaping
 
-   ```luau
-   log:Info(`Hello {name}`, "Hello {name}", {name = name}, {userId = 12345, region = "us-east"})
-   ```
+Interpolated string calls enable safe, ergonomic SQL query construction where the function can automatically escape interpolated values:
 
-   This is tedious, duplicates the `name` variable twice (and the template string once), and is easy to get wrong when refactoring.
+```luau
+local sqlite = require("luau_sqlite")
 
-2. Use a logging library that performs its own template parsing at runtime:
+local function takeUserInput(db: sqlite.DB, user: string, comment: string)
+  -- Auto-escape inputs to guard against SQL injection
+  db:Exec `INSERT INTO user_inputs (user, comment) VALUES ({user}, {comment})`
+  -- Desugars to: db:Exec("INSERT INTO user_inputs (user, comment) VALUES (%s, %s)", {user, comment}, {"user", "comment"})
+end
+```
 
-   ```luau
-   log:Info("Hello {name}", {name = name, userId = 12345, region = "us-east"})
-   ```
+### HTML templating
 
-   This doesn't require new syntax, but requires the logging library to implement its own string interpolation separate from the language's built-in interpolation. It also cannot benefit from compile-time optimizations or static analysis that language-level support would enable.
+Similarly, HTML renderers can automatically escape interpolated values to prevent XSS attacks:
+
+```luau
+local tmpl = require("luau_html_renderer")
+
+local function renderPage(r: tmpl.Renderer, userinput: string)
+  -- Automatic XSS protection through escaping
+  return tmpl.HTML `The user asked about {userinput}`
+  -- Desugars to: tmpl.HTML("The user asked about %s", {userinput}, {"userinput"})
+end
+```
 
 ## Design
 
 ### Grammar
 
-The grammar for function calls is extended to allow an interpolated string as the argument, optionally followed by a comma and a table literal:
+The grammar for function calls is extended to allow an interpolated string as the argument:
 
 ```
 functioncall ::= prefixexp args
-args ::= '(' [explist] ')' | tableconstructor | LiteralString | stringinterp [',' tableconstructor]
+args ::= '(' [explist] ')' | tableconstructor | LiteralString | stringinterp
 ```
-
-Note that the optional `, tableconstructor` is only valid following `stringinterp`, not following a regular `LiteralString` or standalone `tableconstructor`.
 
 ### Semantics
 
-When a function is called with an interpolated string literal in this style, the call receives multiple arguments derived from the interpolated string:
+When a function is called with an interpolated string literal in this style, the compiler desugars the call into a function call with three arguments:
 
-1. **Formatted string**: The fully interpolated result (what you would get from the expression today)
-2. **Template string**: The original template with placeholders intact, e.g. `"Hello {name}"`
-3. **Interpolation table**: A table mapping placeholder names to their values at call time, e.g. `{name = "Alice"}`
-4. **Context table** (optional): If a table literal follows the interpolated string, it is passed as a fourth argument
+1. **Format string**: The template with each `{expression}` replaced by `%s`, e.g. `"The double of %s is %s"`
+2. **Values table**: A sequential table of the interpolation values, e.g. `{a, double(a)}`
+3. **Expressions table**: A sequential table of the original expression source texts, e.g. `{"a", "double(a)"}`
 
-For method calls (`:` syntax), `self` is passed first as usual, followed by these arguments.
+The format string and expressions table are determined at compile time. The values table is evaluated at runtime.
+
+For method calls (`:` syntax), `self` is passed first as usual, followed by these three arguments.
+
+If the interpolated string contains no interpolation expressions, the call passes a single string argument (the literal string), identical to a regular string literal call.
 
 ### Examples
 
@@ -92,24 +102,48 @@ For method calls (`:` syntax), `self` is passed first as usual, followed by thes
 local id = 42
 local user = {name = "Alice"}
 
--- Basic usage: passes three arguments
+-- Simple identifier
 log:Info `Processing item {id}`
--- Desugars to: log:Info("Processing item 42", "Processing item {id}", {id = 42})
+-- Desugars to: log:Info("Processing item %s", {42}, {"id"})
 
--- With context table: passes four arguments
-log:Info `User {user.name} logged in`, {timestamp = os.time()}
--- Desugars to: log:Info("User Alice logged in", "User {user.name} logged in", {user = {name = "Alice"}}, {timestamp = 1234567890})
+-- Member expression
+log:Info `User {user.name} logged in`
+-- Desugars to: log:Info("User %s logged in", {"Alice"}, {"user.name"})
+
+-- Multiple expressions
+log:Info `{user.name} is processing item {id}`
+-- Desugars to: log:Info("%s is processing item %s", {"Alice", 42}, {"user.name", "id"})
 ```
+
+### No expression restrictions
+
+Unlike some alternative designs that use named keys for interpolated values, this design uses positional tables. This means **any expression valid in a regular interpolated string is also valid here**, including function and method calls:
+
+```luau
+-- Function calls are allowed
+log:Info `Result is {compute()}`
+-- Desugars to: log:Info("Result is %s", {compute()}, {"compute()"})
+
+-- Method calls are allowed
+log:Info `Name is {user:getName()}`
+-- Desugars to: log:Info("Name is %s", {user:getName()}, {"user:getName()"})
+
+-- Repeated expressions are fine (each is a separate positional entry)
+log:Info `{increment()} and then {increment()}`
+-- Desugars to: log:Info("%s and then %s", {increment(), increment()}, {"increment()", "increment()"})
+```
+
+Since values are stored positionally rather than keyed by expression text, there is no ambiguity when the same expression appears multiple times or when expressions have side effects.
 
 ### Behavior with variadic functions
 
-Functions that accept variadic arguments will receive all the desugared arguments. For example, Roblox's `print` concatenates its arguments with spaces:
+Functions that accept variadic arguments will receive the three desugared arguments. For example, Roblox's `print` concatenates its arguments with spaces:
 
 ```luau
 local name = "Alice"
 print `Hello {name}`
--- Desugars to: print("Hello Alice", "Hello {name}", {name = "Alice"})
--- Output: Hello Alice Hello {name} table: 0x...
+-- Desugars to: print("Hello %s", {"Alice"}, {"name"})
+-- Output: Hello %s table: 0x... table: 0x...
 ```
 
 This is likely not the desired output. Developers wanting simple string interpolation should use parentheses:
@@ -118,37 +152,7 @@ This is likely not the desired output. Developers wanting simple string interpol
 print(`Hello {name}`)  -- Output: Hello Alice
 ```
 
-The parentheses-free form is intended for functions specifically designed to receive the expanded arguments, such as structured logging APIs.
-
-### Handling of duplicate placeholder names
-
-If the same identifier appears multiple times in a template, it appears once in the interpolation table:
-
-```luau
-log:Info `{x} + {x} = {result}`
--- Interpolation table is {x = 10, result = 20}, not {x = 10, x = 10, result = 20}
-```
-
-### Interpolation table keys for member expressions
-
-When a placeholder contains a member expression like `{user.name}`, the interpolation table uses nested tables mirroring the access path:
-
-```luau
-log:Info `{user.name} is {user.age} years old`
--- Interpolation table is {user = {name = "Alice", age = 30}}
-```
-
-This approach is more ergonomic for consuming code, which can use natural table access like `context.user.name`, and is more idiomatic to Lua/Luau's table-centric design.
-
-When multiple member expressions share a common root, their values are merged into a single nested structure. If a simple identifier `{user}` and a member expression `{user.name}` both appear, the simple identifier provides the full object, which already contains the nested values.
-
-An alternative design would use flat string keys:
-
-```luau
--- Alternative (not proposed): {["user.name"] = "Alice", ["user.age"] = 30}
-```
-
-This would be simpler to implement (no merging logic required) but less ergonomic, requiring consuming code to use string keys like `context["user.name"]`.
+The parentheses-free form is intended for functions specifically designed to receive the expanded arguments.
 
 ### Interaction with existing syntax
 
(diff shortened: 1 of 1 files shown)


docs/syntax-interpolated-string-function-calls.md
```diff
@@ -0,0 +1,306 @@
...
+```luau
+log:Info `{x} {mutate(x)} {x}`
+-- Could produce: {
+--   ["x#1"] = {value = 1},
+--   ["mutate(x)#2"] = {result = nil, args = {x = {value = 1}}},
+--   ["x#3"] = {value = 2}
+-- }
+```
+
+This adds significant complexity. An alternative would be to document that the context captures values at an unspecified point during evaluation, and that templates with side effects may produce inconsistent results. This is left for future consideration.
```
Cooldude2606: Because of metamethods this alternate will be required since accessing a property can mutate it. Or you could ban property access from being used and deviate further from what current string interpolation allows.

```luau
local user = setmetatable({
    id = 0,
}, {
    __index = function(self, key)
        if key == "name" then
            self.id += 1
            return "User" .. self.id
        end
    end,
})

log:Info `Name is {user.name} but also {user.name}`
-- Formated: "Name is User1 but also User2"
-- Template: "Name is {user.name} but also {user.name}"
-- Context: what is user.name? Same key two values, no direct call in the syntax
```

docs/syntax-interpolated-string-function-calls.md
```diff
@@ -0,0 +1,306 @@
...
+
+   This is tedious, duplicates the `name` variable twice (and the template string once), and is easy to get wrong when refactoring.
+
+2. Use a logging library that performs its own template parsing at runtime:
+
+   ```luau
+   log:Info("Hello {name}", {name = name, userId = 12345, region = "us-east"})
+   ```
+
+   This doesn't require new syntax, but requires the logging library to implement its own string interpolation separate from the language's built-in interpolation. It also cannot benefit from compile-time optimizations or static analysis that language-level support would enable.
```
hgoldstein: We should extend the motivation here. Structured logging is cool, but we talked offline about how it does not strictly need this addition to work as needed with _decent_ UX. Off the top of my head, this would probably be useful for escaping when the target needs special handling:
```luau
local sqlite = require("luau_sqlite")

local function takeUserInput(db: sqlite.DB, user: string, comment: string)
    -- Auto-escape our inputs to guard against SQL injection
    db:Exec `INSERT INTO user_inputs (user, comment) VALUES ({user}, {comment})`
end
```
```luau
local tmpl = require("luau_html_renderer")

local function renderPage(r: tmpl.Renderer, userinput: string)
    -- Similar guard against XSS
    return tmpl.HTML `<p>The user asked about {userinput}</p>`
end
```

docs/syntax-interpolated-string-function-calls.md
rofrankel: If we go the direction this RFC proposes (about which, to be clear, I have mixed feelings myself) then I think banning property access would be the way to go.  If we're already deviating from standard string interpolation to prevent potential mutations inside `{}`, we might as well be consistent about it.

(One thing the RFC doesn't note explicitly is that the results of property access and function/method calls can, of course, be cached as local variables.  It's a bit cumbersome and inelegant though.)

docs/syntax-interpolated-string-function-calls.md
```diff
@@ -0,0 +1,306 @@
...
+### Grammar
+
+The grammar for function calls is extended to allow an interpolated string as the argument, optionally followed by a comma and a table literal:
+
+```
+functioncall ::= prefixexp args
+args ::= '(' [explist] ')' | tableconstructor | LiteralString | stringinterp [',' tableconstructor]
+```
+
+Note that the optional `, tableconstructor` is only valid following `stringinterp`, not following a regular `LiteralString` or standalone `tableconstructor`.
```
hgoldstein: Allowing the `tableconstructor` bit is going to create an ambiguity:
```luau
foo(bar `baz`, { bing })
```

It's not clear if this is:
```luau
foo((bar `baz`), { bing })
```
... or ...
```luau
foo((bar `baz`, { bing }))
```

I don't think it's immediately obvious which one would be correct, so dropping the `tableconstructor` seems like the only option here.

docs/syntax-interpolated-string-function-calls.md
```diff
@@ -0,0 +1,306 @@
...
+
+### Semantics
+
+When a function is called with an interpolated string literal in this style, the call receives multiple arguments derived from the interpolated string:
+
+1. **Formatted string**: The fully interpolated result (what you would get from the expression today)
+2. **Template string**: The original template with placeholders intact, e.g. `"Hello {name}"`
```
hgoldstein: As is, the way string interpolation appears to work is that we de-sugar:
```luau
print(`Hello {name}, it is {currtime}!`)
```
... to something like ...
```luau
print(string.format("Hello %*, it is %*!", name, currtime))
```

I'm imagining semantics like:
```luau
template `Hello {name}, it is {currtime}`
```
... desugaring to something like ... 
```luau
-- Could also be an array for the second argument
template("Hello %*, it is %*!", name, currtime)
```

Would that be enough for your use case? I think it also makes the "mistaken" case of ...
```luau
print `Hello, my name is {name}`
```
... a _little_ less unfortunate.

docs/syntax-interpolated-string-function-calls.md
```diff
@@ -0,0 +1,306 @@
...
+
+- Regular string literals: `print "hello"` continues to pass a single string argument
+- Table literals: `print {1, 2, 3}` continues to pass a single table argument
+- Parenthesized calls: `print(`hello {name}`)` continues to pass a single formatted string
+
+The new behavior only applies to calls without parentheses using interpolated strings.
+
+### Expression restrictions in templates
+
+When used in this calling style, the interpolated expressions within the template may not contain function or method calls:
```
hgoldstein: This seems like a weird asymmetry. My personal hesitance with this RFC is that the statements:
```luau
log "hello!"
log(`hello {name}!`)
```
... both _look_ like the syntax you're proposing, act _exactly_ the same, but now the new syntax ...
```luau
log `hello {name}!`
```
... does something different than _either_ snippet. 

On the one hand this restriction makes it clearer that it's new syntax. But on the other it feels like it breaks the mental model users have for interpolated strings. In the below snippets, I do not know how a user is supposed to determine that, say, binary and unary expressions are allowed, and functionally don't appear in the output, but function calls _aren't_ allowed.

docs/syntax-interpolated-string-function-calls.md
MagmaBurnsV: You would have to ban nearly all the operators because of metamethods, e.g. `__add` and `__concat`.

The alternative of having unspecified evaluation is far worse imo. It breaks the strict evaluation rule which leads to confusing semantics for both the programmer and implementor. I believe this is called "local reasoning" in language design.

docs/syntax-interpolated-string-function-calls.md
```diff
@@ -0,0 +1,306 @@
...
+
+## Design
+
+### Grammar
+
+The grammar for function calls is extended to allow an interpolated string as the argument, optionally followed by a comma and a table literal:
+
+```
+functioncall ::= prefixexp args
+args ::= '(' [explist] ')' | tableconstructor | LiteralString | stringinterp [',' tableconstructor]
```
MagmaBurnsV: Wouldn't this syntax be ambiguous in the case of function calls and table construction?

```luau
local history = {
	log `answer: {value}`, {id = 1}
}

-- becomes
local history = {
	log("answer: _", "answer: {value}", {value = value}, {id = 1})
}
-- or
local history = {
	log("answer: _", "answer: {value}", {value = value}),
	{id = 1}
}
```

hgoldstein (changes requested):
The main sticking points here are:
1. The ambiguity noted by the extra parameter(s);
2. The restrictions on what can be passed to interpolated strings.

docs/syntax-interpolated-string-function-calls.md
Bottersnike: Personally me one issue with this is you still end up needing to write your own parser to consume this API. It's not _too_ bad, but you can't just do a simple find and replace because you need to account for `%`-escaping in the passed format string. I think you could cook up a nasty gsub using `"([^%%])%%%*", "%1" .. replace` with an extra check for if the `%*` is at the start of the string, but this feels pretty ugly to me.

hgoldstein: Yeah there's a couple decent outcomes here. My immediate thought, well before I read the RFC, is you'd get the "decomposed" interpolation, so an array of alternating `string` and value parts, e.g.:
```luau
log `Foo {name} bar {1 + 2}`
```
... becomes something like ...
```luau
log({"Foo ", name, " bar ", 3 })
```
I'll admit wanting to preserve the exact string came out of one of the goals of logging. Another option is that we can _probably_ embed the actual text of the interpolated string, e.g. you'll get:
```luau
log `Foo {name} bar {1 + 2}`
```
... becoming ...
```luau
log("Foo {name} bar {1 + 2}", "Foo ", name, " bar ", 3 })
```
.... you can still reconstruct it by `table.concat`ing everything but the first argument.

docs/syntax-interpolated-string-function-calls.md
rofrankel: (from live discussion) This example makes most sense to me if we pass the format string (requiring a subsequent string.format call) rather than the rendered/interpolated string, since in the latter case the rendered string is thrown away.  Luau would also have to pass two tables:  a list of arguments (usable for string.format) as well as a map (since some use cases, like structured logging, want key-value context).  Or perhaps a parallel array with the string representations of the interpolated placeholders, e.g. "foo" or "bar()".

Will consider as we write a new draft.

docs/syntax-interpolated-string-function-calls.md
rofrankel: Yep, agreed - the next draft will remove this.  There's a way to achieve this without explicit language support anyway.

docs/syntax-interpolated-string-function-calls.md
rofrankel: Addressed above, but yep agreed, as we discussed separately we don't need this anyway.

docs/syntax-interpolated-string-function-calls.md
rofrankel: From live discussion with Hunter, perhaps something like:

```luau
local a = 1
local function double(x: number)
  return x * 2
end

-- Desugars to `log:Info("The double of %* is %*", {1, 2}, {"a", "double(a)"})`
log:Info "The double of {a} is {double(a)}"`
```

docs/syntax-interpolated-string-function-calls.md
rofrankel: I think this is resolved in the latest version of the RFC, but please reopen if you disagree.

docs/syntax-interpolated-string-function-calls.md
rofrankel: Updated!  PTAL.

docs/syntax-interpolated-string-function-calls.md
rofrankel: New draft pushed

docs/syntax-interpolated-string-function-calls.md
rofrankel: I've updated the draft, PTAL!

docs/syntax-interpolated-string-function-calls.md
rofrankel: Yes I agree.  This is resolved in the latest draft.

rofrankel:
I've updated this RFC with a substantially rewritten and simplified approach addressing feedback from @MagmaBurnsV, @Cooldude2606, @hgoldstein, and @Bottersnike.  Thanks to @hgoldstein for the live brainstorming.

I am much happier with this version than the original version I wrote up (which came out of a 30 minute group brainstorming session that perhaps needed to be longer :) ).

The latest draft doesn't directly address @Kampfkarren's comment about potential future support for format specifiers, but perhaps it does so implicitly?  E.g. if the `expressions` argument looked like `{"name", "money, .2f"}` would that work or is there some subtle problem?

rofrankel:
> One alternative not mentioned is that you can get pretty close (albeit not with a nice interpolated string) to this behaviour just using existing functionality:
> 
> ```lua
> local function Log(fmt: string)
>   local function log_handler(args: { any })
>     print("Log event. Format string:", fmt, "Evaluated:", string.format(fmt :: any, table.unpack(args)))
>   end
>   return log_handler
> end
> 
> local user = "Bottersnike"
> Log "Hello, %*" { user }
> ```
> 
> I like the changes, though `%*` would make more sense than `%s` given the types of the arguments aren't strictly limited to strings.

Good call, updated.

> > Functions designed for parentheses-free interpolated string calls would need to be written **(or updated)** to accept the three-argument format.
> 
> While not an issue directly with the RFC or it's wording, it's worth noting that updating a function isn't possible in many cases. For example, `print` has no way of knowing if it was called with an interpolated string and should itself perform the formatting, or the user in fact called it as `print(...)` and passed those three values explicitly. I think for the most part any variadic function can't be updated, but any function that currently takes a single string argument could be augmented to support the additional arguments optionally.

Yes this is a valid point - that's why I describe the `Message` approach as an alternative.  I've rephased slightly to try and make that clearer.

docs/syntax-interpolated-string-function-calls.md
```diff
@@ -0,0 +1,318 @@
...
+Functions that accept variadic arguments will receive the three desugared arguments. For example, Roblox's `print` concatenates its arguments with spaces:
+
+```luau
+local name = "Alice"
+print `Hello {name}`
+-- Desugars to: print("Hello %*", {"Alice"}, {"name"})
+-- Output: Hello %* table: 0x... table: 0x...
+```
+
+This is likely not the desired output. Developers wanting simple string interpolation should use parentheses:
```
hgoldstein: As mentioned, I dislike this, but I recognize there's also precedent, [check this snippet out](https://play.luau.org/#N4IgbiBcBMA0IGMogDYHsEEMUAIBmArgHYIAuAlmkfgBQCUkONAzqQE7lEDmsOrH3OgB1qOMWwCmpAm2pCQeNGnm95AI0xt5IiUQAmIkQAcBpGnnrCiJzmfOW6IeMyigAtmj0TkRDAAsJBABrJxA0Iwo3cgAvTAoqABkJMAkUKGgAXwygA)
```luau
local function f(): (string, string)
    return "foo", "bar"
end

print(f()) -- prints "foo\tbar"
print((f())) -- prints "foo"
```

docs/syntax-interpolated-string-function-calls.md
```diff
@@ -0,0 +1,318 @@
...
+
+Structured logging APIs can also inspect the Message's fields directly:
+
+```luau
+function structured_log(msg: Message)
+  emit(tostring(msg), msg.fmt, msg.values, msg.expressions)
+end
+```
+
+While not part of this proposal, this pattern also opens the door for existing functions to be updated to special-case a single `Message` argument, extracting structured data for telemetry while preserving their current behavior for all other callers.
```
hgoldstein: I'm not sure this needs to be here explicitly.

docs/syntax-interpolated-string-function-calls.md
```diff
@@ -0,0 +1,318 @@
...
+A Luau adaptation would pass two tables, string parts and values:
+
+```luau
+tag `Hello {name}, you have {count} messages`
+-- Would call: tag({"Hello ", ", you have ", " messages"}, {name, count})
+```
+
+This was considered and the proposed design shares the same spirit: decomposing the interpolated string into parts that don't require the consumer to parse Luau expressions. The proposed design differs in using a format string with `%*` placeholders instead of a string parts array, and in providing an additional expressions table with the source text of each interpolated expression. The format string approach:
+
+1. Provides a single template string usable as a log aggregation key or cache key
```
hgoldstein: I think we discussed it offline, but wouldn't it be enough to (effectively) reconstruct the format string by doing something like:
```luau
local function mylog(templateparts, values)
    local fmt = table.concat(templateparts, "%*")
    print(string.format(fmt :: any, table.unpack(values)))
end
```
... including metadata about the values _in_ the template aside.

docs/syntax-interpolated-string-function-calls.md
rofrankel: Happy to remove it if example usage is not stylistically appropriate, let me know.

docs/syntax-interpolated-string-function-calls.md
rofrankel: This kind of works, but with two drawbacks:

1. There's collision risk for simple templates that have the same overall structure, e.g. `{foo}: {bar}` and `{baz}: {qux}`.  False positives for aggregation/deduplication may in some cases be even worse than false negatives.
2. Replacing the placeholder expressions with just `"%*"` may hurt readability...in some cases it may be obvious what the placeholders were, but in other cases, less so.

Speaking purely as the user, these drawbacks seem significant enough that I'd rather have the placeholder expressions as well.  And users who are happy without the expressions can just ignore them. I know there may be some performance hit, but as we discussed, all the identified realistic use cases for this functionality are likely going to do something much more expensive anyway with the result (e.g. make a network request), so the language performance cost is not a primary concern.

docs/syntax-interpolated-string-function-calls.md
```diff
@@ -0,0 +1,331 @@
...
+end
+
+-- Desugars to `log:Info("The double of %* is %*", {a, double(a)}, {"a", "double(a)"})`
+-- Note that the second argument evalutes to {1, 2} at runtime, but not at the desugaring step.
+log:Info `The double of {a} is {double(a)}`
+```
+
+The template string enables aggregating logs by message pattern (e.g. grouping all "The double of %* is %*" messages), while the values and expression names provide searchable structured data for platforms like Splunk, Datadog, or Elasticsearch.
+
+Without this feature, developers must either manually construct all arguments (tedious and error-prone) or use a logging library that implements its own template parsing at runtime (duplicating language functionality).
```
hgoldstein: > ... use a logging library that implements its own template parsing at runtime (duplicating language functionality).

I'm not sure that duplicating language functionality is the issue here: it's more that this is _also_ error prone in its own way. For example:
```luau
log:Info("Hello, my name is: {{nome}}", { name = "Hunter" })
```

... whereas we already have a mechanism for constructing strings that can provide analysis input (intellisense, error checking): it's more missed opportunity than duplicating language features being a problem.

docs/syntax-interpolated-string-function-calls.md
```diff
@@ -0,0 +1,331 @@
...
+local function renderPage(r: tmpl.Renderer, userinput: string)
+  -- Automatic XSS protection through escaping
+  return tmpl.HTML `The user asked about {userinput}`
+  -- Desugars to: tmpl.HTML("The user asked about %*", {userinput}, {"userinput"})
+end
+```
+
+## Design
+
+### Grammar
```
aatxe:
```suggestion
### Syntax
```

Nitpick, but "syntax" and "semantics" are the 'right' way to describe the two bits of design here.

andyfriesen:
Would it be better (or worse?) if the desugaring instead included byte offsets into the original string for each substitution?

```lua
log`{timestamp}: Count is {count}.  Total is {count + total}`
-- desugars to
log("{timestamp}: Count is {count}.  Total is {count + total}", {timestamp, count, count + total},
    -- (offset, length) pairs for each substitution
    {{0, 11}, {22, 7}, {45, 15}}
)
```

This trivializes any custom parsing you might want to do.

andyfriesen:
> One alternative not mentioned is that you can get pretty close (albeit not with a nice interpolated string) to this behaviour just using existing functionality:
> 
> ```lua
> local function Log(fmt: string)
>   local function log_handler(args: { any })
>     print("Log event. Format string:", fmt, "Evaluated:", string.format(fmt :: any, table.unpack(args)))
>   end
>   return log_handler
> end
> 
> local user = "Bottersnike"
> Log "Hello, %*" { user }
> ```

I like this interface quite a lot.

I'd like to see this approach added to the "Alternatives" section alongside a strong argument for why it is unviable.

rofrankel:
> Would it be better (or worse?) if the desugaring instead included byte offsets into the original string for each substitution?
> 
> ```lua
> log`{timestamp}: Count is {count}.  Total is {count + total}`
> -- desugars to
> log("{timestamp}: Count is {count}.  Total is {count + total}", {timestamp, count, count + total},
>     -- (offset, length) pairs for each substitution
>     {{0, 11}, {22, 7}, {45, 15}}
> )
> ```
> 
> This trivializes any custom parsing you might want to do.

I'll leave it to the reviewers to say how they feel about this alternative from a language perspective, but I agree with you this would support the motivating use case (and is generically powerful).

rofrankel:
> > Would it be better (or worse?) if the desugaring instead included byte offsets into the original string for each substitution?
> > ```lua
> > log`{timestamp}: Count is {count}.  Total is {count + total}`
> > -- desugars to
> > log("{timestamp}: Count is {count}.  Total is {count + total}", {timestamp, count, count + total},
> >     -- (offset, length) pairs for each substitution
> >     {{0, 11}, {22, 7}, {45, 15}}
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This trivializes any custom parsing you might want to do.
> 
> I'll leave it to you and the other reviewers to say how you all feel about this alternative from a language perspective, but I agree with you this would support the motivating use case (and is generically powerful).
> 
> Happy to update the RFC to reflect this if that's the recommended path.

On reflection, I just updated it to reflect this.

docs/syntax-interpolated-string-function-calls.md
```diff
@@ -100,26 +100,71 @@ args ::= '(' [explist] ')' | tableconstructor | LiteralString | stringinterp
...
 ```
 
-For method calls (`:` syntax), `self` is passed first as usual, followed by these three arguments.
+Parses an interpolation template and returns a sequential table of the stringified interpolation expressions.
 
-If the interpolated string contains no interpolation expressions, the call passes a single string argument (the literal string), identical to a regular string literal call.
+```luau
+string.interpparse("Hello {name}")          -- returns {"name"}
+string.interpparse("{a} + {b} = {a + b}")   -- returns {"a", "b", "a + b"}
+string.interpparse("No interpolations")     -- returns {}
```
Bottersnike: This seems like a downgrade in functionality from previous, where now there's no easy way for code to inspect the non-interpolated parts of an interpolated string. I honestly can't think of a good example use case for when you would want to, but I think it would be nice if `interparse` also gave the offsets for those things. As-is, this function doesn't really seem especially useful; Luau doesn't have `loadstring` so the things returned from this function have very limited utility. It also doesn't scream "parsing the interpolation string" given the limited information it returns.

docs/syntax-interpolated-string-function-calls.md
```diff
@@ -100,26 +100,71 @@ args ::= '(' [explist] ')' | tableconstructor | LiteralString | stringinterp
...
+```luau
+string.interp(template: string, values: {any}) -> string
+```
+
+Renders an interpolation template by replacing each `{expression}` with the `tostring` of the corresponding value from the values table, matching the behavior of normal string interpolation. The values table may be either sequential (positional matching) or associative with string keys (matching by expression name):
+
+```luau
+-- Sequential: values matched positionally to expressions
+string.interp("Hello {name}", {"Alice"})          -- returns "Hello Alice"
+string.interp("{a} + {b} = {a + b}", {1, 2, 3})   -- returns "1 + 2 = 3"
```
Cooldude2606: This is an unclear example, is the final expression "3" because `1 + 2` is `3` or because the positional value is "3"? I assume the latter is intended. Therefore an example should be used that shows that the contents of the expression is not a constrained on the value.

e.g. `string.interp("{a} + {b} = {a + b}", {1, 2, "three"})   -- returns "1 + 2 = three"`

docs/syntax-interpolated-string-function-calls.md
```diff
@@ -100,26 +100,71 @@ args ::= '(' [explist] ')' | tableconstructor | LiteralString | stringinterp
...
+
+-- Associative: values matched by expression name
+string.interp("Hello {name}", {name = "Alice"})   -- returns "Hello Alice"
```
Cooldude2606: What happens in the following scenarios?
`string.interp("Hello {name}", {"bob", name = "Alice"})`
`string.interp("{a} + {b} = {a + b}", {b = 1, 2, "three"})`

I would assume it matches by name, then any unmatched expressions consume positional values in order:
`Hello Alice` and `2 + 1 = three`

Bottersnike: As another scenario, what about `string.interp("{2} {1}", {"First", "Second"})` (ie literal values, but that are also array indexes}

Bottersnike:
**Edit, and TLDR**: I don't think there's any good solution involving conversion of the interpolated string into any other string representation that is actually viable to consume. I believe the only reasonable solution is to pass the invoked function each chunk of the interpolated string (be that string chunks or value chunks) with associated metadata.

Any string format including the original representation is ripe for parsing issues, or substantial duplication of existing language code. Any string format that removes the original representation (eg `%*`) needs extra baggage to annotate what that value was (for structured logging).

The rest of this comment explains the issues with the currently proposed `{{`/`}}` escaping and where even the example implementation falls flat regarding this.

---

Something that came to mind is that `{{` as an escape sequence is different to how `{`s are normally escaped. For example, to normally escape a brace in an interpolation string you have to do:

```luau
print(`\{{1}\}`)
```

Which would print `{1}`.

If a user was to write

```luau
print `\{{1}\}`
```

Then that would desugar to `print("{{{1}}}", ...)` if we're treating `{{` as an in-string escape sequence for a non-interpolation-literal `{`. I'm not sure if there _is_ a good way to make this nice and consistent with the rest of the language, but I thought it was worth raising at least as a mild concern. As a similar thought, `{{` in an interpolated string is considered invalid syntax, reminding users that `Double braces are not permitted within interpolated strings`. This is useful both as a way to catch errors, and to drastically ease parsing.

If we consider, for example, the string `"{{{1}}}"`, matching the brackets is now a non-trivial operation. A naive scan for a non-duplicate `{` followed by a non-duplicate `}` is impossible to do with just string matching.

Let's consider the reference implemented for `string.interp` provided at https://github.com/luau-lang/luau/pull/2261/changes#diff-5754ebda9ef82adaa3498deff7066d2cef90204707aa9cb3067101238f49bebbR1726

We're going to use

```luau
capture `\{{"Hello }} world"}\}`
```

As an interpolated string, this evaluates to `{Hello }} world}`. If we parse that by hand, we first have a literal `{`, then a value of `"Hello }} world"` followed by a final literal `}`. If we think about escaping this into a string, we convert the literal braces into brace pairs, and end up with either `"{{{\"Hello }} world\"}}}"` or `"{{{\"Hello }}}} world\"}}}"` depending on if we also escape the braces within the value. Passing that "escaped" string into the example `string.interp` implementation we get an output of `{Hello }}world}}world"}}` or `{Hello }}}}world}}world"}}`. Needless to say these are both nonsense.

The reason for this is the example `string.interp` implementation naively counts `{` and `}` symbols to calculate the "depth" of nesting. This is thwarted by the reality of language syntax. A properly working implementation of `string.interp` rapidly approaches the complexity of the entire language's lexer.

At this time, I don't have an ideal _solution_ that just drops into the existing RFC and fixes this all. The best I can think of off the top of my head would be to entirely change the signature of an interpolated string function call to return pre-chunked parts of the string. Something like `({str1, str2, str3, str4}, {repr1, repr2, repr3}, {value1, value2, value})` (always with `N+1`, `N+1`, and `N` items in each of those three tables) would resolve this, allowing the lexing pass that has already been performed on the language to be utilised, rather than reconstructing a faux representation of the interpolated string for runtime use. This is obviously a large divergence from the currently suggested method.

Edit: A different example I thought of that bypasses this problem while retaining all the useful functionality would be

```luau
fn `hello {name}, do you like {food}?`
-> fn("hello ", {repr="name", value="Bottersnike"}, " do you like ", {repr="food", value="Oranges"}, "?")
```

gaymeowing:
This RFC as is isn't type safe, which I would consider an absolute requirement for any new luau feature. Given there's been all this work into making the type system as great as it is today, it'd be rather stupid to throw all that to the wayside.

So I've thought of a possible type safe alternative to what this RFC is trying to achieve which adds a new 'keyword' `tagged` that goes before the `function` keyword:
```luau
tagged function log:Info(`{rule_id: number}: {action: string} {src_ip: string}:{src_port: string} -> {dst_ip: string}:{dst_port: string} proto={proto: buffer} bytes={bytes: buffer}`)
     -- code here
end
```

Where a "interpolated" string is used to define the functions arguments, with the brackets containing function arguments with their type.
The non bracket stuff could be given via a string argument called `full` that works similarly to the `self` argument when a function is created by using `:` on a table.
There could also be an array argument that contains the stuff not in brackets given similar to `self`, although dunno what it'd be called, as I can't think of a good name for it.

Perhaps there could be a variable provided called `format`, which would be a format string version of "interpolated" string used to define arguments. Which you could call `string.format` on to then format the arguments.
Alternatively `format` could be a function which the compiler sees and then de-sugars into a `string.format` call with all of the arguments:
```luau
tagged function log:Info(`{rule_id: number}: {action: string} {src_ip: string}:{src_port: string} -> {dst_ip: string}:{dst_port: string} proto={proto: buffer} bytes={bytes: buffer}`)
     --[[
          this would de-sugar into:
          print(string.format("%*: %* %*:%* -> %*:%* proto=%* bytes=%*", rule_id, action, src_ip, src_port, dst_ip, dst_port, proto, bytes))
     ]]
     print(format())
end
```

rofrankel:
Thanks for all the feedback and apologies for the churn/lack of responsiveness.  Some of the updates are in response to internal conversations that are happening in parallel with this quasi-public process.  (meta:  there's probably room for improvement there)

There might be one more substantive change to this RFC in the works, which might implicitly address some of the current feedback (and/or make it obsolete).  If not then I'll respond point by point.  Thanks again for the thoughtful comments.

andyfriesen:
Hey everyone.

After quite a lot of spirited internal debate, I've made some changes to this proposal.

We have dropped the feature where the original source text of the substitution tokens is part of the desugaring because the only use-case anyone had that called for it was structured logging.

Secondly, we wanted to minimize the impact of use of this API with functions that aren't intended to be used with interpolated function calls.  To handle this, the desugaring now results in a single value that represents the interpolated string expression.  This value provides a `__tostring` metamethod (and a few others) so that the result can easily be used in a string-like way if so desired.

The proposed desugaring now works as follows:

```luau
sql:Exec `SELECT * FROM users WHERE id = {userId}`

-- desugars to:

sql:Exec(string.interpolatedstring.create(
    {"SELECT * FROM users WHERE id = ", ""},
    {userId}
))
```

Bottersnike:
Personally, this solution seems like a really clean one with how it Just Works with existing things but allows access to its internals when you want to get to them. Two thoughts pop into my mind:

- How does this interact with types, will there be a `type interpolatedstring = <big mess here>` or will it be considered a primitive? Will it be compatible with a `string` or would a consumer need to use `string | interpolatedstring` as their type?
- Would it make sense to also pass the original literal value (presumably normalised given the reality of AST->back to luau) to `interpolatedstring.create()` for use-cases like log aggregation?

While additional overhead, it could also be interesting for the `create` function to capture the filename+line of the caller (again, thinking for a use case like log aggregation, or something like python's `warnings` library that only prints any given warning exactly once). Capturing name+line is less of a concern than capturing the literal though because the function being _passed_ the `interpolatedstring` could also do that logic.

Cooldude2606:
Many, if not all, standard library functions are fast callable. However, in this case `create` is contained within a table that must be indexed. How does this interact with fast calling?

For greater string compatibility, the table could include an index metamethod and define the same methods as string. This increases compatibility because it is common for strings passed as parameters to use method calls. E.g. `my_string:sub(4)`. Has this been considered?

gaymeowing:
> Many, if not all, standard library functions are fast callable. However, in this case `create` is contained within a table that must be indexed. How does this interact with fast calling?

It would just compile as `GETIMPORT`, just like how referencing `table.insert` doesn't use `GETTABLEKs` but instead just uses a single `GETIMPORT`. With a fastcall instruction emitted after the `GETIMPORT`.

gaymeowing:
Whys there still the `string.interpparse()` and other interpolated string methods still directly on the `string` library, rather than being under `string.interpolatedstring`?
Whilst it would be annoying to type out `string.interpolatedstring.parse`, autocomplete could be improved so that keys of the `interpolatedstring` sub table get recommended when you type `string.i`. So the user isn't required to type out `string.interpolatedstring.` before being able to complete the `parse` method. 

It would also be better if it was just called `interpolated` rather than `interpolatedstring`, as its already clear its a string because its on the `string` library.

andyfriesen:
> How does this interact with types, will there be a type interpolatedstring = <big mess here> or will it be considered a primitive? Will it be compatible with a string or would a consumer need to use string | interpolatedstring as their type?

It's not a primitive.  It's just a table with a metatable.

It's not compatible with string.  We could add a type alias to make `string | interpolatedstring` a more convenient thing to write.

> Would it make sense to also pass the original literal value (presumably normalised given the reality of AST->back to luau) to interpolatedstring.create() for use-cases like log aggregation?

Maybe someday?  One of the reasons I wanted the desugaring to include the wrapping table is that it gives us an easy extension point.

> Whys there still the string.interpparse() and other interpolated string methods still directly on the string library, rather than being under string.interpolatedstring?

`string.interpolated` sounds okay.

`string.interpparse()` is gone, replaced by the vaguely analogous `string.interpolate()`.  Now that you mention it, though, we could probably drop it entirely.  You can just write `tostring(string.interpolated.create({"Hello ", "!"}, {name}))`

MagmaBurnsV:
If it's not supposed to be compatible with string, should it even include metamethods? The only argument for them seems to be to just make variadic functions like `print` format to the interpolated string instead of the table, which I find weak.

If we drop the metamethods, then we could go back to the design where interpolated calls desugar into a plain table literal that can be reconstructed by the consumer with a `string.interpolate` function.

gaymeowing:
> It's not a primitive. It's just a table with a metatable.

Could it be a builtin userdata so `typeof()` can be used to check for an interpolated string? As I can imagine that being useful for debugging if somewhere in a codebase wrong values are being passed to functions.

Bottersnike:
Personally I think you'd want either `typeof()` to work or `tostring()` to work, for the case where a function takes, say, `any` as a type but wants to stringify it for whatever reason. If there's no support for `tostring`, you would need a way to check if it's an interpolated string, but `typeof(x) == "table"` would match far more than just legal values for `string.interpolated[string].XXX` functions.

gaymeowing:
> Could it be a builtin userdata so `typeof()` can be used to check for an interpolated string? As I can imagine that being useful for debugging if somewhere in a codebase wrong values are being passed to functions.

Forgot `getmetatable` exists and can be used here:
```luau
getmetatable(meow) == string.interpolated
```

PR #173, opened 2026-02-09
RFC: Unique types

athar-adv:
[Rendered](https://github.com/athar-adv/unique-types/blob/master/docs/unique-types.md)

Introduces the syntax `unique type TypeName` to define unique opaque types

### DIFF
--- a/docs/unique-types.md
+++ b/docs/unique-types.md
@@ -20,7 +20,7 @@ A structure that is an intersection between some type T and a unique type U will
 ---
 The proposed syntax to create a unique type is to define it using `unique type TypeName`, it does not have an = symbol after it because all unique types are are opaque unique types, they do not contain any additional information beyond that.
 
-The name `unique` was chosen as it clearly conveys that this specific type is unique and is not the same as any other type
+The name `unique` was chosen as it clearly conveys that this specific type is unique and is not the same as any other type.
 Unique types of the same name defined in different files will, of course, still be unique as their own "primitive" type, and cannot be cast to eachother.
 
 ### Behavior with intersections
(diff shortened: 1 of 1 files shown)


Bottersnike:
Ref: https://github.com/luau-lang/rfcs/pull/123

docs/unique-types.md
```diff
@@ -0,0 +1,265 @@
+# Unique Types
+---
+## Summary
+---
+This RFC proposes adding support for unique types to luau.
```
alexmccord: This line is redundant, RFCs by definition already suggests this.

docs/unique-types.md
```diff
@@ -0,0 +1,265 @@
...
+
+Unique types are an extension to the type checker, and make no changes to runtime semantics.
+
+## Motivation
+---
+Since Luau uses structural typing, there is no way to make a primitive distinct. If there are two types PlayerId and AssetId and they are both strings, the type checker allows a PlayerId to be passed into a function expecting AssetId because they are both just string.
+
+Current workarounds like tagging (string & { _tag: "PlayerId" }) are messy and confuse autocomplete.
+
+Unique types solve this by being completely unique from any other type, therefor allowing programmers to bar casts between different unique types.
```
alexmccord:
```suggestion
Unique types solve this by being completely unique from any other type, therefore allowing programmers to bar casts between different unique types.
```

What is "bar" here?

deviaze:
I think this has basically been solved by classes, which are first class user definable nominal types.

surfbryce:
@deviaze You may want to reread the RFC, a large portion of it is dedicated to non-classable data-types.

For example, the PlaceId/UserId examples. They are both numbers (non-classes), but the distinction between their purposes is extremely important, and *should* bear a distinction in the type system (but still operate as numbers).

PR #176, opened 2026-02-16
Extend design/drawbacks/alternatives in integer type proposal

vegorov-rbx:
No functional changes, added clarification on behavior that was not explicitly written out and provided better drawbacks and alternative sections.

[Rendered](https://github.com/luau-lang/rfcs/blob/integer-extra-notes/docs/type-long-integer.md)

### DIFF
--- a/docs/type-long-integer.md
+++ b/docs/type-long-integer.md
@@ -6,11 +6,11 @@ A builtin 'integer' type to represent 64 bit integer numbers.
 
 ## Motivation
 
-Current methods of representing 64-bit integers in Luau with default environment are done with high-low pairs or by splitting them among vectors which leads to poor ergonomics and may not be sufficient for cases where performance is important.
+Current methods of representing 64-bit integers in Luau with the default environment are done with high-low pairs or by splitting them among vectors which leads to poor ergonomics and may not be sufficient for cases where performance is important.
 
-In the case of a high-low pair implementation, each individual number takes 16 bytes (or 24) and needs to be be handled together which can be confusing.
+In the case of a high-low pair implementation, each individual number takes 16 bytes (or 24) and needs to be handled together which can be confusing.
 
-In the case of vector implementations, ergonomics are improved by storing the entire integer in one value however you have the same issues of implementation complexity and lack of integration with the type system.
+In the case of vector implementations, ergonomics are improved by storing the entire integer in one value. However you have the same issues of implementation complexity and lack of integration with the type system.
 
 In both cases, performance is lacking compared to what could be provided by a native implementation.
 
@@ -21,11 +21,11 @@ Besides the above solutions in userland, runtimes are also able to define their 
 
 Further, representing 64-bit numbers as heap-allocated userdata is sufficient for correctness, but also adds a layer of indirection for something that could fit within the existing value size. In domains where the use of these numbers is more common, the performance characteristics of these varying implementation techniques could be problematic or even prohibitive.
 
-Userdata remain the right mechanism for complex or embedder-specific numeric types, but 64-bit integers have a sufficiently large problem domain to justify their inclusion as a core value type, and bring significant benefits for performance and interoperability in the overall Luau ecosystem.
+Userdata remains the right mechanism for complex or embedder-specific numeric types, but 64-bit integers have a sufficiently large problem domain to justify their inclusion as a core value type, and bring significant benefits for performance and interoperability in the overall Luau ecosystem.
 
 ## Design
 
-This will be implemented a with new type called `integer`.
+This will be implemented with a new type called `integer`.
 
 An additional character may be specified at the end of numeric literals `i` which will signify an 64 bit integer literal.
 64 bit integer literals will support separators, hex, and binary values:
@@ -43,9 +43,17 @@ Operations performing a string formatting of an integer should format the number
 
 Integer values will have a built-in equality comparison, but will not have any other operators or metamethods defined.
 
+Integers are never automatically converted to numbers or strings, and vice-versa.
+Passing an integer to a function expecting a number (or string) will result in a type error.
+
+While there are no arithmentic operators defined for integers in this RFC, it is still worth noting that operations between `number` and `integer` are also not supported.
+
 Functions for creating and manipulating this type will exist in a new library called 'integer`.
 
 ### Library
+
+Unless otherwise specified, all operations interpret integer values as signed two's complement 64 bit numbers.
+Operations that treat the value as unsigned are explicitly prefixed with 'u'.
 
 `function integer.create(n: number): integer?`
 
@@ -64,7 +72,7 @@ String is allowed to have leading and trailing spaces and number can be preceded
 If base is not specified, number can have a `0x` or `0X` prefix to be converted in base 16, otherwise base 10 is used.
 
 When base is specified, it has to be in range between 2 and 36 inclusive.
-In base 10 and base 16, number is allowed to have a `0x` or `0X` prefix.
+In base 10 and base 16, numbers are allowed to have a `0x` or `0X` prefix.
 
 Returns `nil` if the string doesn't contain a number.
 
@@ -102,7 +110,7 @@ Overflow wraps around according to rules of two-complement arithmetic.
 
 `function integer.div(a: integer, b: integer): integer`
 
-Performes signed truncated division of `a` by `b`.
+Performs signed truncated division of `a` by `b`.
 
 If `b` is 0, throws a division by zero error.
 
@@ -118,7 +126,7 @@ If `a` is -2^63 and `b` is -1, result is 0.
 
 `function integer.idiv(a: integer, b: integer): integer`
 
-Performes signed floored division of `a` by `b`.
+Performs signed floored division of `a` by `b`.
 
 If `b` is 0, throws a division by zero error.
 
@@ -126,7 +134,7 @@ If `a` is -2^63 and `b` is -1, throws an overflow error.
 
 `function integer.mod(a: integer, b: integer): integer`
 
-Performes signed floored modulus division of `a` by `b`.
+Performs signed floored modulus division of `a` by `b`.
 
 If `b` is 0, throws a division by zero error.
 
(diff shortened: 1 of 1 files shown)


docs/type-long-integer.md
```diff
@@ -332,8 +340,46 @@ Included in `luaL_openlibs`.
 
 ## Drawbacks
 
-This increases the complexity of the VM as it may need to implement separate paths for mathematical operations.
+This introduces a library with a large set of functions which will likely take a large number of fastcall slots.
+
+The lack of operator overloads causes code with integers to be more verbose, decreasing readability in complex expressions.
```
hgoldstein, line 345: Do we want to have a similar section here about signed-ness? All integers being signed by default, but only exposing signed-ness via specific builtins means its easier to misuse the unsigned builtins.

docs/type-long-integer.md
```diff
@@ -332,8 +340,46 @@ Included in `luaL_openlibs`.
...
+Polymorphic overloaded operators present many challenges and can cause unexpected drops of performance.
+This is already true when working with `vector` types without type annotations and given the focus of this proposal, we do not want to provide additional pitfalls.
+Monomorphic library functions allow compiler and runtime to optimize more aggressively.
+
+Heavy uses of 64-bit integers often rely on bitwise operations.
+Since Luau does not have bitwise operators in its syntax, even with a subset of overloaded operators in place, code will still end up using library functions.
+
+Integer division has multiple conflicting definitions (truncated vs. floored).
+Operators like `/` and `%` would force a specific default behavior that might not match the specific rounding that the user intended.
+Default behavior might also be less performant, causing developers to still rely on library methods.
```
hgoldstein, line 373: I'm not sure if its worth referencing something like Python, which ended up adding additional operators for _specifically_ flooring integer division.

hgoldstein (approved):
This seems fine as-is, just left some suggestions.

docs/type-long-integer.md
vegorov-rbx: I go over the signed/unsigned topic in the alternatives, not sure it has a place here.

hgoldstein: Maybe something like:
```
Not distinguishing between signed and unsigned integers may make it easier to misuse the unsigned APIs, as there is no mechanism to tell the user they're making a mistake.
```

docs/type-long-integer.md
hgoldstein: We spoke offline: it probably would not add to the RFC given Python has made _other_ very different choices (they have a _ton_ of overloads for `+`, for example).

docs/type-long-integer.md
vegorov-rbx: The use of unsigned APIs might also be intentional as mentioned in alternatives section now.

docs/type-long-integer.md
```diff
@@ -332,8 +340,54 @@ Included in `luaL_openlibs`.
...
+
+### Operator overloads
+
+This proposal does not introduce overloaded operators aside from equality checks.
+This omission is deliberate.
+
+While the lack of operators makes code more verbose, Luau's primary numeric type remains `number`.
+The `integer` type is intended to cover specialized use cases, many of which will specifically prefer explicit library functions over operators for performance.
+
+Polymorphic overloaded operators present many challenges and can cause unexpected drops of performance.
```
MagmaBurnsV, line 364: I understand this rationale and how it's basically a blocker for adding any operator support, but I also think this is just non-negotiable.

First, just reflect on the absurdity of supporting an integer type in a language but leaving out operator support. That's nuts. It heavily discourages casual use and has terrible ergonomics. As someone else pointed out, not even `-1i` is valid because there's no `__unm`.

Second, making users have to choose which signedness to use for every operation makes integers feel like a really esoteric compiler-intrinsic feature. Even C doesn't need that level of specification. If special optimizations and bit-twiddling is all integers will ever be used for, does that really make them worth the cost of adding?

hgoldstein: What you're observing is that, yes, `integer` is a specialized data type in Luau, which is not the case for other programming languages that tend to either have disjoint types (effectively all static languages) or make no distinction between integers and floats at all (JavaScript, Lua, Luau prior to this RFC).

In that context, adding a specialized type needs to balance ergonomics against usage of the rest of the language. The bet we're making is that most folks will use continue to use `number` (the as mentioned primary data type), and some will reach for `integer` as it suits their needs, and it doesn't make sense to jeopardize the performance of `number` for this new tool. FWIW: I don't think anything in this RFC locks us out of supporting operators in the future.

MagmaBurnsV: This RFC has *never* been advertised in that light before, to my knowledge. In fact, I would argue this interpretation goes against the entire spirit of the documented motivation, which is to provide first-class support for integers as a core value type instead of relying on a userdata implementation. People generally wanted integers for casual stuff like, y'know, adding and multiplying them, not for doing niche optimizations.

Even if it's theoretically possible to shoehorn operators back in under the restraints given in this RFC, let's face it, that probably won't ever happen. If it will jeopardize VM performance now it will in the future. The assumption should be that integers will never get operators, not that we'll worry about it later.

Bottersnike: Operators are mostly locked out unless you want to only support a subset of operators, with the subset that can't be implemented (such as comparison) still falling back to the existing issue of no type safety.

How would `-1i` be implemented? Will a special syntax need implemented explicitly to handle `'-' INTEGER` as an atom? Assumably in that case `'-' [whitespace stripped by lexer] INTEGER` would be fine? In that case, how would `5 - 1i` parse? Would that be `NUMBER '-' INTEGER` (valid production rule `exp`, invalid runtime) or `NUMBER NEGATIVE_INTEGER` (invalid syntax)?

Looking at the compiler itself, I'd guess it would just be parsed as an `unary` as normal, then a special case added in for `LOP_MINUS`? In that case though would `-(1i)` be considered valid? `-(a bunch of integer code here)`?

It feels like one operator _is_ being implemented in that case, which feels antithesis to the goal of not having operators, and falls right into the problem of `-integer.udiv(...)`

Bottersnike: [merged into the above]

Bottersnike: And to add more thoughts to it, if it's done in the parser we now have to content with `- -5 = 5` but what does `- -1i` equal? Do we do it in the parser and break parity with how unaries are expected, or do it in the VM and produce a nonsense `-(unsigned)` operator?

docs/type-long-integer.md
Cooldude2606: Why is this a concern when floor division has already been implemented as an operator in lauu as `//` (See [RFC](https://github.com/luau-lang/rfcs/blob/master/docs/syntax-floor-division-operator.md)). This means that `/` would truncate and `//` would floor. I could see `%` maybe being an issue, however modulo is defined as being "signed floored modulus division" by this RFC, so why would it truncate?

vegorov-rbx: Because even in presence of `//` this loses the relationship of dividend, remainder and original number.
Additionally, this would mean `/` is truncating division in Luau and flooring division in Lua.

Cooldude2606: From what I read in the spec, 5.3 and above, `/` operating on integers will convert them to floats, using the closest higher or lower representable value if it can not be represented exactly, and will always return a float. So lua does not truncate nor floor with standard division on integers. Because of this luau should be free to decide which it uses, as all choices (including not implementing it) are deviations from lua spec.

vegorov-rbx: That's not correct, spec says that conversion to float is only performed for float division, meaning when both operands are not integers.

docs/type-long-integer.md
hgoldstein: >  In fact, I would argue this interpretation goes against the entire spirit of the documented motivation, which is to provide first-class support for integers as a core value type instead of relying on a userdata implementation.

I see why you've come to that conclusion, but I think the key point in the motivation is that while you _can_ do this without native support, the domain is large enough to justify native support: we'd rather have _a_ integer implementation that works and is paired with a native integer type than having `N` different implementations.

docs/type-long-integer.md
Cooldude2606: You are incorrect because there is no defination for anything other than "float division" (`/`) and "floor division" (`//`) in the [reference](https://www.lua.org/manual/5.5/manual.html#3.4.1). This means that `int / int` performs "float division" and `int // int` performs "floor division". This can be plainly seen in the following example where both values are represented as ints internally and yet the output is a float. (executed on Lua 5.4.6)
```lua
local a = 1 << 62
local b = 3 << 61
local c = a / b
print(a, b, c)
-- 4611686018427387904	6917529027641081856	0.66666666666667
```
Because int -> float coercion has been rejected by this RFC, it is not possible to impliment "float division" for integers; therefore allowing Luau to instead define "integer division" using `/` as "trucated division". The spec also defines modulo `%` as the remainder from floor division. There is no deviation from Lua by these definitions, other than that already introduced by the RFC in the rejection of coercion.

vegorov-rbx: Ah yes, I was looking into idiv opcode of the Lua VM.

docs/type-long-integer.md
MagmaBurnsV: The way I see it is that I'd rather use a nonstandard userdata implementation than a standard native implementation with terrible ergonomics.

Without operators, this feature falls short at providing even a remotely better solution for anything besides fine-tuned intrinsics, which I have seen *no one* outside the Luau team voice support for. I'm not even sure this was a part of the original author's intent.

Luau shouldn't be designing new features "just to have it in case, we hope people will use it," especially for something as groundbreaking as integers. I'm almost certain that this feature set and motivation would've been firmly rejected by the team 2-3 years ago considering how narrow it has evolved.

petrihakkinen:
Interesting RFC.

For reference, I have implemented uint64s on user side as tagged lightuserdata type. I register a metatable for lightuserdata (shared between all tagged lightuserdata types) and in the metamethod on native side I lookup up function pointer to the arithmeric operation based on the lightuserdata tag. This way I can have operators (and tostring) for uint64s but also for any other tagged lightuserdata types without any changes to the Luau VM. Granted, the metatable dispatch adds some overhead, but I don't use uint64s in high performance code.

Side-note: It would be interesting if the size of the lightuserdata could be expanded with a configuration option to allow using the full size of the tagged value for the contents. This would make lightuserdata even more useful with very minimal change to Luau itself.

docs/type-long-integer.md
petrihakkinen: @MagmaBurnsV note that you are not restricted to slow heap allocated userdata for integers in user code in order to support operators. See my other message below.

PR #177, opened 2026-02-17
Type only imports for Luau

VegetationBush:
[Rendered](https://github.com/VegetationBush/rfcs/blob/VegetationBush-type-only-imports/docs/type-only-importing.md)

Implements a type-only import for Luau similar to that of TypeScript. This feature aims to introduce type extraction without runtime imports.

### DIFF
--- a/docs/type-only-importing.md
+++ b/docs/type-only-importing.md
@@ -106,4 +106,4 @@ For the case of extending the functionality of `require`:
 - No longer a function.
 
 ## Alternatives
-Use a header file for type definitions. Although, this scales poorly and can become cumbersome in larger projects.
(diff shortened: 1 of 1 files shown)


YarikSuperpro:
I get the intent, but this feels like it's solving an IntelliSense limitation by adding more structural overhead to the language.

Types in Luau are already compile-time and don't need runtime representation, so introducing a separate "type-only import" still creates an extra conceptual container in the module system just for tooling to interpret differently.

That ends up working around IntelliSense constraints rather than addressing them directly, and it can reinforce the idea that types need their own import mechanism instead of improving how existing type information is inferred and reused.

PR #179, opened 2026-02-25
Mutable before return Export-by-Value Semantics

MagmaBurnsV:
This proposes alternative semantics for #42, which treats exports as syntax sugar for assigning keys to an export table before sealing.

This allows reassigning exported values before the module is returned, solving the original RFC's issues with mutually dependent functions and shorthand aliasing. Furthermore, it introduces export semantics that are already representable in the language, dismissing the need for introducing `const`-ness to the language.

[Rendered](https://github.com/MagmaBurnsV/rfcs/blob/34e2efcc3d5f0895570cea8e104bac66505a72ae/docs/export-keyword.md)

### DIFF
--- a/docs/export-keyword.md
+++ b/docs/export-keyword.md
@@ -1,7 +1,7 @@
 # Export by Value
 
 ## Summary
-Extend the `export` keyword to support variables and functions as syntax sugar for constructing a table and returning it from a module.
+Extend the `export` keyword to support variables and functions as syntax sugar for constructing a table and returning it from the module.
 
 ## Motivation
 Today, type aliases are able to be exported from a module using the `export` keyword.
@@ -14,12 +14,10 @@ However, this mechanism is currently only supported by types.
 
 Extending  `export`  to also support variable and function declarations would provide a consistent way for users to expose a stable API from modules.
 
-Furthermore, introducing a language-level concept of exporting opens the door to many optimizations not currently possible with dynamic module returns, such as cross-module inlining and constant-folding.
+Furthermore, introducing a language-level concept of exporting opens the door to many optimizations not currently possible with dynamic module returns, such as cross-module inlining and constant folding.
 
 ## Design
 The `export` contextual keyword will now be allowed anywhere before variable and function declarations at the top level of a module, including `local`, `const` and `function` declarations.
-
-Exporting declarations that are in nested `do end` blocks under the top scope are also permitted. Attempting to export a declaration outside the module scope will result in a parse error.
 
 Exported variables will count towards the local variable limit, as they can be optimized into locals by the compiler.
 
@@ -30,16 +28,6 @@ export const TAU = math.pi * 2
 export function init() -- exported functions are always const
 	-- TODO
 end
-
-do
-	local counter = 0
-	
-	export function increment(): number
-		counter += 1
-		return counter
-	end
-end
--- counter and increment not visible here
 
 if foo then
 	export local bar = 1 -- not allowed, syntax error
@@ -55,13 +43,11 @@ export local d -- same as export d = nil
 export const MAX_ITEMS = 10 -- cannot be reassigned
 ```
 
-Exporting a variable with the same identifier twice is a parse error, regardless of the scope it was defined in.
+Exporting a variable with the same identifier twice is a parse error.
 
 ```luau
 export local foo = 1
-do
-	export local foo = 2 -- syntax error
-end
+export local foo = 2 -- syntax error
 ```
 
 However, exporting a variable with the same identifier as another non-exported variable is allowed, following conventional lexical scoping and shadowing rules.
@@ -83,7 +69,7 @@ animal = "bird"
 print(animal) -- bird
 ```
 
-Exported function declarations are also always treated as const, as reassigning function declarations is usually a mistake.
+Exported function declarations are also always treated as const, as reassigning function declarations is unidiomatic and typically a mistake.
 
 ```luau
 export function f() end
@@ -91,7 +77,7 @@ f = 1 -- illegal
 ```
 
 ### Desugared Form
-Exporting variables desugars into assigning keys to a table that is then frozen and returned once the module scope ends.
+Exporting variables desugars into assigning keys to a table that is then frozen and returned once the module scope ends. Desugared examples are purely illustrative, as the compiler is free to optimize and restructure table construction internally.
 
 ```luau
 export local a = 1
@@ -112,7 +98,7 @@ if math.random(0, 1) == 1 then
 end
 
 export local f, g
-function f()
+function f() -- same as f = function() ...
 	g()
 end
 function g()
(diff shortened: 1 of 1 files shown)


Cooldude2606:
The following cases are currently unclear in their behaviour, root causing being a lack of definition for the interaction of export statements with existing variables. I.e. you currently only explain that two exports cannot share an identifier.

```luau
-- This is mentioned in alternatives, but not in body
local function foo() end
export foo -- syntax error?

local bar = 0
export bar = 1 -- syntax error?
print(bar) -- 0 or 1?

fruit = "apple" -- global
export fruit -- shadowing or reference?
fruit = "pear" -- is fruit now local?
print(fruit) -- pear?
print(_G.fruit) -- apple?

export animal
animal = "dog"
local animal = "cat" -- syntax error?
print(animal) -- cat or dog?
```

MagmaBurnsV:
I've clarified that the restriction only applies to exported variables, and exports follow conventional shadowing rules in regard to non-exported variables, so your example would desugar into:

```luau
local _EXP = {}

local function foo() end
_EXP.foo = nil

local bar = 0
_EXP.bar = 1
print(_EXP.bar) -- 1

fruit = "apple"
_EXP.fruit = nil
_EXP.fruit = "pear"
print(_EXP.fruit) -- pear
print(_G.fruit) -- apple

_EXP.animal = nil
_EXP.animal = "dog"
local animal = "cat"
print(animal) -- cat

-- {foo = nil, bar = 1, fruit = "pear", animal = "dog"}
return table.freeze(_EXP)
```

Or alternatively if optimized into locals by the compiler:

```luau
local function foo() end
local foo

local bar = 0
local bar = 1
print(bar)

fruit = "apple"
local fruit
fruit = "pear"
print(fruit)
print(_G.fruit)

local _animal = nil
_animal = "dog"
local animal = "cat"
print(animal)

return table.freeze({foo = foo, bar = bar, fruit = fruit, animal = _animal})
```

(Uninitialized exports looking like shorthand exports may be a good reason not to support them. The only reason I allowed them in the design was to obtain parity with how the `local` keyword works, but I am fine with removing them if that's what the team wants.)

Cooldude2606:
> Uninitialized exports looking like shorthand exports may be a good reason not to support them.

While this is true, it should be really easy to detect most cases through linting because export follows existing shadowing rules. In fact the same lint would catch another common error which could result from mistypes or copy paste errors.

I would propose "Uninitiated export is never assigned to" as a lint rule.
This also mirrors common lint rules such as unussed local variable, or variable is never accessed.

```luau
local function foo() end
export foo -- Uninitiated export is never assigned to
-- The above also catches the incorrect assumption of exporting locals
-- A common lint rule "no shadowing" would also catch this error

export bar -- No lint because bar is used in an assignment
if condition then
    bar = function() end
end

export f, g -- No lint because f and g is used in an assignment
function f() g() end
function g() f() end

export user -- Uninitiated export is never assigned to
User = {

}

export fruit = nil -- Uninitiated export is never assigned to
-- Although assigned nil explictly, there would never be a reason to export nil
```

Lint rules are not necessarily part of the langauge, but the possibilty of it being easily lintable helps to remove confussion for users and maintains feature parity with the local keyword as you described.

SPY:
We discussed internally different approaches resolving mutual recursive functions and related issues. I think we all agreed mutable-before-return semantic will be least evil here. I intend to merge this RFC at the end of the week if no strong evidence against it will be provided.

docs/export-keyword.md
```diff
@@ -0,0 +1,255 @@
...
+
+Extending  `export`  to also support values and functions would provide a consistent way for users to expose a stable API from modules. Furthermore, static exports would allow for many future optimizations, such as cross-module inlining and constant folding.
+
+## Design
+Allow the `export` contextual keyword anywhere the `local` keyword may be used at the top level of a module, as well as within nested `do end` blocks. Attempting to export a variable outside the module scope will be a parse error.
+
+Exported variables will count towards the local variable limit, as they can be optimized into locals by the compiler.
+
+```luau
+export version = "5.1"
```
SPY: Also we can think about `export const version = "5.1"` syntax to make assignment to `version` a syntax error(or even make it default). In this case `export local x` can be symmetric syntax explicitly signaling `x` will be modified later.

docs/export-keyword.md
```diff
@@ -0,0 +1,255 @@
...
+export [a] = t -- looks like export[a] = t
+```
+
+Uninitialized exported variables may also cause confusion due to looking like "shorthand exports" that export previously defined variables. For example, the following does not export the function `foo` but instead declares an exported variable that shadows it. This can be solved by implementing a lint that warns about uninitialized exports that are never assigned to.
+
+```luau
+local function foo(x) return x * 2 end
+export foo -- exports nil, shadows foo
+```
+
```
tommyscholly, line 237: One additional drawback is, in large Luau files, reassignable exports can introduce a potential footgun. That is the tradeoff to enable cleaner mutual recursion, but I think it should be mentioned at least.

docs/export-keyword.md
```diff
@@ -0,0 +1,255 @@
...
+export tau = math.pi * 2
+print(tau)
+
+-- becomes
+local tau = math.pi * 2
+print(tau)
+return table.freeze({tau = tau})
+```
+
+Furthermore, static exports that are never reassigned to are capable of being inlined and constant folded across modules, assuming future support for static imports.
```
tommyscholly: The "staticness" of the export is not required for potential optimizations. Because export statements are top-level statements, we can determine it's final exported value at compile time.

docs/export-keyword.md
MagmaBurnsV: Honestly, since I made this RFC with the expectation that const would be rolled back (because the whole rationale for const was to fill the "can't reassign" gap immutable exports brought), I never had any plans for introducing const-ness to exported tokens. Now that const is here but immutable exports aren't, things are awkward.

I guess I can try to retrofit everything with const in mind, but my fear is the result is going to be a little "sore" since the features here are going to added in backwards.

MagmaBurnsV: I guess the only rational way to do this is to make `export` a keyword that goes before declarations like you've suggested. Functions/classes might have to be special-cased where they are always exported const, but that should be fine.

docs/export-keyword.md
MagmaBurnsV: I guess a more correct way to phrase this would be "first-class exports allow for the compiler to assume exported tokens cannot be reassigned after module return, unlocking the ability to bring constant-folding and inlining across module boundaries."

docs/export-keyword.md
```diff
@@ -0,0 +1,253 @@
...
+However, this mechanism is currently only supported by types.
+
+Extending  `export`  to also support variable and function declarations would provide a consistent way for users to expose a stable API from modules.
+
+Furthermore, introducing a language-level concept of exporting opens the door to many optimizations not currently possible with dynamic module returns, such as cross-module inlining and constant-folding.
+
+## Design
+The `export` contextual keyword will now be allowed anywhere before variable and function declarations at the top level of a module, including `local`, `const` and `function` declarations.
+
+Exporting declarations that are in nested `do end` blocks under the top scope are also permitted. Attempting to export a declaration outside the module scope will result in a parse error.
```
tommyscholly: Can I have an example where you would want to hide exported values from the rest of the module they are defined in? I'm struggling to see why this would be desired behavior.

MagmaBurnsV: The first codeblock in the doc shows an example where an exported function references a scoped upvalue for encapsulation.

```luau
do
	local counter = 0
	
	export function increment(): number
		counter += 1
		return counter
	end
end
-- counter and increment not visible here
```

It's less about hiding the exported bindings and more about hiding their dependencies.

tommyscholly: Hiding the dependencies makes sense, however, I do think it will be nicely covered by the classes RFC, plus any additional RFC extensions to add a `private` keyword.

SPY: This part seems quirky to me as well. It introduces weird ghost definitions and it is easy to miss with global assignment.
```
do
    export local f = 4
end

add(f + 5) -- valid syntax, f is global here

export local f = 5 -- syntax error, I cannot use f , but it changes scope anyway
```

I think if we want to have `export` at top level blocks we must allow them to escape the scope.
it works the same with other global syntax, globals themselfs
```
do
    local l = 23
    function globalFunc() return l end
end

globalFunc()
```

`globalFunc` refers locally scoped local, but as global entity still available outside of it

MagmaBurnsV: I don't see anything weird going on here since this is already how locals with global variable access behave like. Just like if `f` was a normal local variable, using it out of scope will give you nil and raise a lint/type error.

The only difference is you can't reuse the same name for an exported variable anywhere, which is already a given for anything you're exporting.

MagmaBurnsV: For the sake of keeping things moving, I'm just going to remove this and defer it to a future RFC since it's safe to readd later.

tommyscholly:
Clarifying intended attribute syntax, I assume we would expect:

```luau
@native
export function test() end
```

and not

```luau
export @native function test() end
```

SPY:
Yes, I would definitely expect 
```luau
@native
export function test() end
```

SPY (approved):
LGTM up to attributes remark

tommyscholly (approved):

PR #180, opened 2026-02-28
RFC: Add vector.sqrmagnitude function

Ramdoys:
[Rendered](https://github.com/Ramdoys/rfcs/blob/sqrmagnitude/docs/squaredmagnitude.md)

### DIFF
--- a/docs/squaredmagnitude.md
+++ b/docs/squaredmagnitude.md
@@ -1,4 +1,4 @@
-# Squared Magnitude
+# vector.sqrmagnitude
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


Ramdoys:
Why does Copilot do that...

docs/squaredmagnitude.md
```diff
@@ -0,0 +1,51 @@
...
+```luau
+local range = 100
+if vector.sqrmagnitude(vec1 - vec2) < range^2 then
+    --Within the range.
+end
+```
+
+Even outside of distance comparisons, it is much easier and more understandable to write `vector.sqrmagnitude(vec)` than `vector.dot(vec, vec)`.
+
+Second, performance. When computing the squared distance between two points, the alternative `vector.dot(vec1 - vec2, vec1 - vec2)` computes the subtraction twice. A single parameter `vector.sqrmagnitude(vec1 - vec2)` avoids this redundancy. The subtraction is computed once and operated on directly. In case of `vector.dot`, even if someone stored the difference in a local, `vector.sqrmagnitude` would still potentially be faster since it is handled as a single intrinsic rather than going through the dot product way.
```
tommyscholly, line 33: While true, this point really should just be filed as a potential optimization opportunity, rather than requiring an entirely new library function just to avoid the double subtraction.

PR #182, opened 2026-03-18
RFC: 64-bit integer type proposal improvements

vegorov-rbx:
This PR adds changes to the RFC following additional user feedback and early usage feedback.
It also fixes more wording issues and adds clarity to some changes that were not mentioned.

* goes over how negative integer literals are supported
* `min`/`max` are variadic to match the `math` library they mirror
* `band`/`bor`/`box`/`btest` are variadic to match the `bit32` library they mirror
* `gt`/`ge`/`ugt`/`uge` comparisons to not worry about transforming the order to fit into the small set of operations
* explicitly mention the change to `tostring`, `type` and `typeof`
* mention support in `rawequal`; similar to previous base library functions, it is expected to work on `any`
* remove `integer.tostring` since it duplicates `tostring`
* add `types.integer` constant for use in type functions
* `math.mininteger` and `math.maxinteger` moved to `integer.minsigned` and `integer.maxsigned`

### DIFF
--- a/docs/type-long-integer.md
+++ b/docs/type-long-integer.md
@@ -10,7 +10,7 @@ Current methods of representing 64-bit integers in Luau with the default environ
 
 In the case of a high-low pair implementation, each individual number takes 16 bytes (or 24) and needs to be handled together which can be confusing.
 
-In the case of vector implementations, ergonomics are improved by storing the entire integer in one value. However you have the same issues of implementation complexity and lack of integration with the type system.
+In the case of vector implementations, ergonomics are improved by storing the entire integer in one value. However, you have the same issues of implementation complexity and lack of integration with the type system.
 
 In both cases, performance is lacking compared to what could be provided by a native implementation.
 
@@ -28,7 +28,7 @@ Userdata remains the right mechanism for complex or embedder-specific numeric ty
 This will be implemented with a new type called `integer`.
 
 An additional character may be specified at the end of numeric literals `i` which will signify a 64-bit integer literal.
-64-bit integer literals will support separators, hex, and binary values:
+64-bit integer literals will support separators, hexadecimal, and binary values:
 
 ```luau
 local a = 123i
@@ -38,6 +38,12 @@ local d = 0b1000_1000i
 ```
 
 Integer literal numbers have to be exactly representable, overflow is a parsing error.
+
+Binary and hexadecimal literals can specify the full value including the sign bit:
+```luau
+local a = 0xFFFF_FFFF_FFFF_FFFFi -- -1i
+local b = 0b11111111_11111111_11111111_11111111_11111111_11111111_11111111_11111111i -- -1i
+```
 
 Operations performing a string formatting of an integer should format the integer as signed by default.
 
@@ -70,11 +76,35 @@ This behavior is chosen to closely match `math.tointeger` from Lua 5.4 and can h
 Converts a string representation of a number into an integer.
 
 The number inside the string has to be an integer.
-The string is allowed to have leading and trailing spaces and the number can be preceded by a sign.
+The string is allowed to have leading and trailing spaces and the number can be preceded by a '+' or '-' sign.
 If base is not specified, the number can have a `0x` or `0X` prefix to be converted in base 16, otherwise base 10 is used.
 
-When base is specified, it has to be in range between 2 and 36 inclusive.
+When base is specified, it has to be in the range between 2 and 36 inclusive.
 In base 10 and base 16, numbers are allowed to have a `0x` or `0X` prefix.
+
+```luau
+integer.fromstring("11", 2) -- 3i
+integer.fromstring("-11", 2) -- -3i
+
+integer.fromstring("-1", 10) -- -1i
+integer.fromstring("-11", 10) -- -11i
+
+integer.fromstring("-1", 16) -- -1i
+integer.fromstring("-11", 16) -- -17i
+
+integer.fromstring("-0x1", 10) -- -1i
+integer.fromstring("-0x11", 10) -- -17i
+```
+
+When base of the number is not 10, all integer bits can be specified, including the sign bit:
+
+```luau
+integer.fromstring("0x8000000000000000") -- 0x8000000000000000i
+integer.fromstring("0xFFFFFFFFFFFFFFFF") -- 0xFFFFFFFFFFFFFFFFi or -1i
+integer.fromstring("8000000000000000", 16) -- 0x8000000000000000i
+integer.fromstring("FFFFFFFFFFFFFFFF", 16) -- 0xFFFFFFFFFFFFFFFFi or -1i
+integer.fromstring("1777777777777777777777", 8) -- 0xFFFFFFFFFFFFFFFFi or -1i
+```
 
 Returns `nil` if the string doesn't contain a number.
 
@@ -82,9 +112,9 @@ This behavior is chosen to match `tonumber`, but excludes floating-point numbers
 
 `function integer.tonumber(n: integer): number`
 
-Converts an integer to a double.
+Converts an integer to a double number.
 
-If the value cannot be represented as a double exactly, round to nearest, ties to even rounding mode is used.
+Precision loss can occur if the integer is not exactly representable.
 
 `function integer.neg(a: integer): integer`
 
@@ -332,6 +362,14 @@ Integer value representing -2^63 (`-9_223_372_036_854_775_808i`)
 'o', 'u', 'x' and 'X' format specifiers will format an integer as an unsigned 64-bit integer number.
 
 Other format specifiers like 'e', 'f', 'g', etc will throw an error.
+
+`string.pack` and `string.unpack` are **not** extended to accept integer numbers.
+
+Today, signed and unsigned integer specifications in these functions work with double numbers.
+
+Allowing integers to be accepted would create a disparity between `string.pack` inputs and `string.unpack` results.
+
+New specifiers specific to integer numbers could be introduced in the future if there is a user need.
 
 ### Extensions to the type library (type functions)
 
--- a/docs/type-long-integer.md
+++ b/docs/type-long-integer.md
@@ -80,10 +80,6 @@ Returns `nil` if the string doesn't contain a number.
 
 This behavior is chosen to match `tonumber`, but excludes floating-point numbers.
 
-`function integer.tostring(n: integer): string`
-
-Converts an integer to a string representation in a signed form with no 'i' suffix.
-
 `function integer.tonumber(n: integer): number`
 
 Converts an integer to a double.
--- a/docs/type-long-integer.md
+++ b/docs/type-long-integer.md
@@ -320,6 +320,14 @@ Returns 64 if `n` is zero.
 
 Returns `n` with the order of the bytes swapped.
 
+`integer.maxsigned: integer`
+
+Integer value representing 2^63-1 (`9_223_372_036_854_775_807i`)
+
+`integer.minsigned: integer`
+
+Integer value representing -2^63 (`-9_223_372_036_854_775_808i`)
+
 ### Extensions to the base library
 
 `tostring` will convert an integer to a string representation in signed form with no 'i' suffix.
@@ -340,18 +348,6 @@ Reads an integer out of a buffer at the given offset.
 `function buffer.writeinteger(b: buffer, offset: number, value: integer): ()`
 
 Writes an integer to a buffer at the given offset.
-
-### Extensions to the math library
-
-For parity with Lua 5.3+, two additional constants are added to the math library.
-
-`math.maxinteger`
-
-Integer value representing 2^63-1 (`9_223_372_036_854_775_807i`)
-
-`math.mininteger`
-
-Integer value representing -2^63 (`-9_223_372_036_854_775_808i`)
 
 ### Extensions to the string library
 
--- a/docs/type-long-integer.md
+++ b/docs/type-long-integer.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-A builtin 'integer' type to represent 64 bit integer numbers.
+A builtin 'integer' type to represent 64-bit integer numbers.
 
 ## Motivation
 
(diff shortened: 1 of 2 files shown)


docs/type-long-integer.md
```diff
@@ -144,59 +146,83 @@ If `a` is -2^63 and `b` is -1, result is 0.
...
-Performs a bitwise xor (exclusive or) of `a` and `b`.
+`function integer.btest(...integer): boolean`
+
+Performs a bitwise AND of all arguments and returns `true` iff the result is not 0.
+If there are no arguments, returns `true`.
 
 `function integer.lt(a: integer, b: integer): boolean`
 
-Compares signed less than (<) comparison of `a` and `b`.
+Performes a signed less than (<) comparison of `a` and `b`.
```
ahae-roblox: Performs?

docs/type-long-integer.md
```diff
@@ -144,59 +146,83 @@ If `a` is -2^63 and `b` is -1, result is 0.
...
 
 `function integer.le(a: integer, b: integer): boolean`
 
-Compares signed less than or equal (<=) comparison of `a` and `b`.
+Performes a signed less than or equal (<=) comparison of `a` and `b`.
```
ahae-roblox: See above ("Performs"; also additional instances below)

docs/type-long-integer.md
```diff
@@ -144,59 +146,83 @@ If `a` is -2^63 and `b` is -1, result is 0.
...
+Performes a signed greater than (>) comparison of `a` and `b`.
+
+`function integer.ge(a: integer, b: integer): boolean`
+
+Performes a signed greater than or equal (>=) comparison of `a` and `b`.
 
 `function integer.ult(a: integer, b: integer): boolean`
 
-Compares unsigned less than (<) comparison of `a` and `b`.
+Performes a unsigned less than (<) comparison of `a` and `b`.
```
ahae-roblox: less-than? (Also other occurrences below)

aatxe (approved):

AsynchronousAI:
Hi, I don’t know if this is the place to mention but a popcount function would also be helpful. 

Not a requirement tho because it can still be manually implemented to be fast with countlz and countrz and a pop_lsb but having it be O(1) with CPU instructions would be extremely useful.

vegorov-rbx:
Popcnt does make sense and it was considered.
The reason it was excluded is that it makes sense for `bit32` library as well, so it could be a separate proposal to add to both.

AsynchronousAI:
> Popcnt does make sense and it was considered. The reason it was excluded is that it makes sense for `bit32` library as well, so it could be a separate proposal to add to both.

One more thing, will a `buffer.writei64` function be added?

docs/type-long-integer.md
ahae-roblox, line 352: Do we want to support string.pack()/string.unpack()?

docs/type-long-integer.md
```diff
@@ -39,37 +39,39 @@ local d = 0b1000_1000i
...
 
-Functions for creating and manipulating this type will exist in a new library called 'integer`.
+Negative integer literals are allowed when unary `-` is applied to the literal directly in the AST: `-123i`, `-0b1000i`, but not `-(123i)`.
+
+Functions for creating and manipulating this type will exist in a new library called `integer`.
 
 ### Library
 
-Unless otherwise specified, all operations interpret integer values as signed two's complement 64 bit numbers.
+Unless otherwise specified, all operations interpret integer values as signed two's complement 64-bit numbers.
```
MagmaBurnsV, line 63: Since there are no functions to convert to/from unsigned integers and literals are parsed as signed, how are you even supposed to represent unsigned numbers? If someone wanted to add two unsigned integer literals together, they would have to write them in signed form and then format the result back to unsigned.

Perhaps this warrants a `u` suffix for unsigned literals? At the very least, there needs to be some way to convert from unsigned number/string and back.

docs/type-long-integer.md
```diff
@@ -39,37 +39,39 @@ local d = 0b1000_1000i
...
 
-Number inside the string has to be an integer.
-String is allowed to have leading and trailing spaces and number can be preceded by a sign.
-If base is not specified, number can have a `0x` or `0X` prefix to be converted in base 16, otherwise base 10 is used.
+The number inside the string has to be an integer.
+The string is allowed to have leading and trailing spaces and the number can be preceded by a sign.
+If base is not specified, the number can have a `0x` or `0X` prefix to be converted in base 16, otherwise base 10 is used.
 
 When base is specified, it has to be in range between 2 and 36 inclusive.
 In base 10 and base 16, numbers are allowed to have a `0x` or `0X` prefix.
```
MagmaBurnsV: How does `integer.fromstring` work with different bases if it's signed? Since Lua 5.1 defines `tonumber` to only allow unsigned integers when using different bases, it's unclear what something like `integer.fromstring("-1", 2)` would give.

docs/type-long-integer.md
```diff
@@ -301,9 +327,15 @@ Integer value representing -2^63 (`-9_223_372_036_854_775_808i`)
 
 `string.format` function is updated to support integer arguments.
 
-'d', 'i' and '*' format specifiers will format an integer as a signed 64 bit integer number.
+'d', 'i' and '*' format specifiers will format an integer as a signed 64-bit integer number.
+
+'o', 'u', 'x' and 'X' format specifiers will format an integer as an unsigned 64-bit integer number.
```
MagmaBurnsV, line 358: I find it weird that hexadecimal integer literals are parsed as signed while formatting into a hexadecimal string is unsigned.

This is made more odd considering integer literals are supposedly parsed based on their numerical value instead of their binary representation, so `0xfffffffffffffffbi` isn't `-5` but an out-of-range literal, while `string.format("%x", -5i)` would output `"fffffffffffffffb"`.

MagmaBurnsV: Maybe hexadecimal and binary integer literals should be forbidden to avoid this footgun?

docs/type-long-integer.md
MagmaBurnsV: It was mentioned somewhere else before, but these are slightly pointless since they are integers instead of numbers, so backported code won't be compatible, e.g. `x <= math.maxinteger`. It would make more sense to put these in the integer library as they don't provide actual parity.

docs/type-long-integer.md
```diff
@@ -39,37 +39,39 @@ local d = 0b1000_1000i
...
 Integer values will have a built-in equality comparison, but will not have any other operators or metamethods defined.
 
 Integers are never automatically converted to numbers or strings, and vice-versa.
 Passing an integer to a function expecting a number (or string) will result in a type error.
 
-While there are no arithmentic operators defined for integers in this RFC, it is still worth noting that operations between `number` and `integer` are also not supported.
+While there are no arithmetic operators defined for integers in this RFC, it is still worth noting that operations between `number` and `integer` are also not supported.
 
-Functions for creating and manipulating this type will exist in a new library called 'integer`.
+Negative integer literals are allowed when unary `-` is applied to the literal directly in the AST: `-123i`, `-0b1000i`, but not `-(123i)`.
```
MagmaBurnsV, line 57: This still seems like it would cause many problems down the line.

How is whitespace treated, e.g. is `- 1i` allowed? What about in the case of `5 - 1i`? Is that invalid since it parses as two consecutive expressions, or does it parse as `sub(5, -1i)` which throws a runtime error?

Either way, this introduces whitespace significance to operators, which I can't imagine being popular with users.

WheretIB: `- 1i` is allowed.
No problems with `5 - 1i`, just like `5 - 1` it parses as a binary operation.
No special parsing handling was required in the implementation.

docs/type-long-integer.md
vegorov-rbx: Binary and hexadecimal literals can include all 64 bits, including the sign bit.

vegorov-rbx: We can add more examples for clarity on this.

docs/type-long-integer.md
vegorov-rbx: Answered in the earlier comment, 0xfffffffffffffffbi is valid.

docs/type-long-integer.md
vegorov-rbx: Lua does place them with an integer tag value as far as I know.
But I'll defer to @vrn-sn to see if this addition doesn't work nice in Luau.

docs/type-long-integer.md
vegorov-rbx: `integer.fromstring("-1", 2)` will be `-1i`.
I can add an explicit mention (maybe not about the particular example, but which interpretation parsing takes for the full range of values)

docs/type-long-integer.md
tommyscholly: nit: Should be performs, same with the lines below.

vegorov-rbx: Probably a stale review? This was fixed in 1f6dabd4bd1feb7a872a68198b9f6c33b043928a

docs/type-long-integer.md
MagmaBurnsV: That doesn't make sense since the numeric value would change. For example, the literal `0x8AC7230489E80000` currently represents the value `10000000000000000000` while `0x8AC7230489E80000i` would be `-8446744073709551616i`.

Right now, it reads that suffixing any number literal will try to convert its value to a signed integer, not that it will reinterpret it.

docs/type-long-integer.md
MagmaBurnsV: That's weird, since it appears Luau wraps around with `tonumber("-1", 2)` and gives `9223372036854776000`. This is different from Lua 5.1, so is that just an implementation bug?

docs/type-long-integer.md
MagmaBurnsV: Lua 5.3 integers are interned with the number type and have operator support. What I meant was adding/comparing these won't work since Luau integers are a distinct type that have no operators.

docs/type-long-integer.md
vegorov-rbx: It does make sense and works.

vegorov-rbx: We will add examples so it is explicitly mentioned.

docs/type-long-integer.md
vegorov-rbx: That's not an implementation bug, that's a choice Luau made.

docs/type-long-integer.md
MagmaBurnsV: Would `5 -1i` be different from `5 - 1i`? Also, how would scientific notation like `10e-1i` parse?

vegorov-rbx: No, there is no difference between `5 -1i` and `5 - 1i` just as there is no difference between `5 -1` and `5 - 1`.

vegorov-rbx: Scientific notation is not allowed.

docs/type-long-integer.md
vegorov-rbx: Added a section about it, it doesn't seem backwards compatible to accept integers in those functions.

vrn-sn: @MagmaBurnsV, I think it's a good point. Adding them to remain consistent with Lua 5.3+ feels hollow if the underlying semantics don't match nicely. It's probably more confusing to include them in `math`—how do we feel about `integer.max` and `integer.min` instead?

Another alternative is to just leave them out if the primary benefit was compatibility. I don't think that's true though, I think they're reasonably valuable and easy to add.

MagmaBurnsV: `integer.maxsigned` and `integer.minsigned` would be better names as max/min are already library functions. Unsigned max constant could also be considered, although you can easily represent it with `-1i`.

docs/type-long-integer.md
```diff
@@ -39,74 +39,102 @@ local d = 0b1000_1000i
 
 Integer literal numbers have to be exactly representable, overflow is a parsing error.
 
-Operations performing a string formatting of an integer should format the number as signed by default.
+Binary and hexadecimal literals can specify the full value including the sign bit:
+```luau
+local a = 0xFFFF_FFFF_FFFF_FFFFi -- -1i
+local b = 0b11111111_11111111_11111111_11111111_11111111_11111111_11111111_11111111i -- -1i
```
MagmaBurnsV, line 45: Just to clarify, what would `-0xffffffffffffffffi` be? Because depending on how that is parsed, it could either be `1` or `-9223372036854775807`.

Inversely, is `-0x8000000000000000i` invalid from overflow, or does it wrap back around to `0x8000000000000000i` due to two's complement?

vegorov-rbx: `-0xffffffffffffffffi` is `1i`, `-0x8000000000000000i` is `0x8000000000000000i`.

docs/type-long-integer.md
vrn-sn: I'm happy going in that direction. We can revisit if `integer.maxunsigned` is worth adding in a future amendment, but I don't think it's important to address now (and is backwards compatible with this RFC nonetheless).

@vegorov-rbx, if you are happy with `integer.maxsigned` and `integer.minsigned`, we can amend this section in the RFC. We should also update this wording, as it is now simply part of the `integer` library, and the primary motivation is no longer perfect parity with Lua 5.3+ but simply usability.

> ### Extensions to the math library
>
> For parity with Lua 5.3+, two additional constants are added to the math library.

vegorov-rbx: Moved and renamed.

vrn-sn:
Wanted to mention that as part of this PR, maybe it's worth also amending this section of the math constants RFC (math-constants.md):

> Integer types are not supported in Luau, so math.maxinteger and math.mininteger wouldn’t make much sense, but adding other commonly used constants would be an easy value-add for users of the language.

It's worth keeping up-to-date since Luau does now support integers, along with support for the integer library constants instead.

Maybe: "Luau supports similar integer constants through its `integer` library, but other commonly used math constants are an easy value-add for users of the language."

vegorov-rbx:
> Maybe: "Luau supports similar integer constants through its `integer` library, but other commonly used math constants are an easy value-add for users of the language."

Applied.

PR #183, opened 2026-03-25
RFC: N-Dimensional Vectors

AsynchronousAI:
This RFC proposes using vectors with more than 3 or 4 components instead of tables or multiple vectors.

Alongside this comes a `function vector.dim(v: vector): number` function to query how many components a vector has.

### DIFF
--- a/docs/n-dimensional-vectors.md
+++ b/docs/n-dimensional-vectors.md
@@ -1,0 +1,89 @@
+# N-Dimensional Vectors
+
+## Summary
+
+Update the vector type and library to support vectors with any number of components, using numeric indexing beyond four dimensions.
+
+## Motivation
+
+Luau's `vector` type is currently restricted to 3 or 4 components. While this covers most spatial use cases, development with Luau demands native support for higher-dimensional vectors for reduced GC load and accelerated scalar operations.
+
+Currently, if you need more than 4 dimensions, multiple vectors or tables must be used:
+
+```lua
+-- Adding all components of two 6D vectors
+local a1, a2 = vector.create(1,2,3), vector.create(4,5,6)
+local b1, b2 = vector.create(6,5,4), vector.create(3,2,1)
+print(a1 + b1, a2 + b2)
+
+-- Adding all components with tables
+local a = {1,2,3,4,5,6}
+local b = {6,5,4,3,2,1}
+
+local c = {}
+for i = 1, #a do
+  c[i] = a[i] + b[i]
+end
+
+print(c)
+```
+
+This approach has several drawbacks:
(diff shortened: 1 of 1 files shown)


gaymeowing:
This doesn't work as vectors that are 4 wide are restricted behind a FFlag because they increase the size of a TValue so they can fully fit on the stack.

Although I do think there is value in making vectors more generalized, as I've commonly ran into situations where I'm using vectors that don't necessarily have an `x` or `y`. Like for instance an RGB color represented as a vector; it'd be nice to be able to create a vector alias that allows for accessing the components using `r`,`g`, and `b`.
But that's a ways away, as currently the compiler doesn't actually use the type solver. Plus the solver isn't stable enough to where that could be a possibility, although it is close currently.

AsynchronousAI:
> This doesn't work as vectors that are 4 wide are restricted behind a FFlag because they increase the size of a TValue so they can fully fit on the stack.

Potentially for larger vectors a reference/pointer could be used similarly to how buffers are.
This would add overhead but should still be much faster than other solutions.

Bottersnike:
One concern I can think of with the "auto-switch between 3/4 size and 'large' size" approach is how it would interact with NCG. `vector` operations are able to make use of SIMD, but if the size of a vector can no longer be statically determined those optimisations are going to struggle. I could see the possibility for `foo (a: vec, b: vec) if (a.dim == 4 and b.dim == 4) { SIMD path } else { slow path }` but that's not especially ideal or pretty. Someone who's more familiar with the precise inner-workings of NCG might be able to provide a solution to that problem though.

docs/n-dimensional-vectors.md
```diff
@@ -0,0 +1,89 @@
...
+```lua
+local v = vector.create(1, 2, 3, 4, 5, 6)
+print(v.x)   -- 1
+print(v.y)   -- 2
+print(v[5])  -- 5
+print(v[6])  -- 6
+```
+
+* Indices 1–4 are equivalent to `.x`, `.y`, `.z`, `.w`, so `v[1] == v.x` is always true.
+* Accessing an out-of-bounds index returns `nil`.
```
whoamimaybeme, line 68: Consider the following code:
```lua
local newvector = vector.create(1, 2)
print(newvector.z)
```
Currently, this would print 0. Does this line mean that z would now be equal to nil?

vegorov-rbx:
Thank you for the suggestion, but we do not plan to make such a change to the vector type.

SPY:
I would suggest to research extension to `buffer` library to provide more vector-like functions like `buffer.addF32(a, b, len, aOff = 0, bOff = 0)`

PR #184, opened 2026-03-28
Revert "Introduction of 'const' keyword"

loadstring1:
This reverts commit 3ebffa2575ef0ae87184d645a7783a8631a66a15. (https://github.com/luau-lang/rfcs/pull/166)

I am proposing the removal of the const keyword from the Luau core. While the intent behind it is clear, its implementation introduces unnecessary syntax bloat that contradicts Luau’s goal of being a lightweight, high-performance scripting language.

Luau’s compiler is already highly efficient at identifying constants through Constant Folding and Dead Code Elimination. Adding a formal const keyword provides marginal runtime benefits while increasing the complexity of the VM’s bytecode and the parser.

False sense of immutability in a language where tables are the primary data structure, const is often misleading. It prevents variable reassignment but does not prevent table mutation. This creates a "shallow" safety net that confuses developers coming from languages with deeper immutability.

Adding more keywords moves Luau further away from the simplicity of the Lua ecosystem. One of Luau’s greatest strengths is its low barrier to entry; adding "syntactic sugar" that serves no functional purpose beyond what table.freeze() or type-checking already provides makes the language feel cluttered.

Every new keyword increases the surface area for bugs in the linter, autocomplete engines, and the parser itself. The cost of maintaining this feature outweighs the stylistic preference of having an explicit constant marker.

Zrzut ekranu 2026-03-28 202408

It is also worth noting the significant community pushback regarding this addition. The original proposal received a high volume of negative feedback (over 26 downvotes), signaling that the vast majority of the developer base perceives this feature as unnecessary complexity rather than a helpful tool.

Pushing through a syntax change of this magnitude—despite clear community opposition—risks alienating the developers who value Luau for its streamlined, "no-nonsense" approach. Reverting this would show a commitment to building a language that reflects the actual needs and preferences of its users.


pyxfluff:
Why are we complaining about new language features we have been asking about for years? Most major other languages already have this and Lua does too (with weirder syntax). If you think it's bloat, just don't use it. It's that simple.

Very nice and thoughtful to have AI write this issue too. Likewise I propose this issue get closed.

loadstring1:
> Why are we complaining about new language features we have been asking about for years? Most major other languages already have this and Lua does too (with weirder syntax). If you think it's bloat, just don't use it. It's that simple.

Just because other languages have this feature doesn't mean Luau should. This is a language designed for game development, not JavaScript. We should avoid unnecessary bloat, which is why I’m requesting its removal. I’m not trying to force this change feel free to reject my PR if you consider my arguments invalid.

> Very nice and thoughtful to have AI write this issue too. Likewise I propose this issue get closed.

I can't tell if you're being sarcastic, but if the way I phrased things is an issue, I can open a new PR and describe everything in my own words without any AI assistance.

StoryBakeryDev:
I also think it might be better not to do it, but since it's already been implemented, it probably can't be reverted.
personally, there are times when I define things like this:

```lua
local CONST
if cond1 then
    CONST = 1
elseif cond2 then
    CONST = 2
else
    CONST = 3
end
```
But if `const` is added and it becomes the general convention for defining constants:
```lua
const CONST
if cond1 then
    CONST = 1
elseif cond2 then
    CONST = 2
else
    CONST = 3
end
```
this will naturally throw an error.

well, there are various alternatives, but using an if-then-else expression is not an option.

Roblox's Lua style guide also advises against using chained if-then-else expressions

[https://roblox.github.io/lua-style-guide/#if-then-else-expressions](https://roblox.github.io/lua-style-guide/#if-then-else-expressions)

> While if expressions do support elseif, it should be used sparingly. If your set of conditions is complicated enough to need several elseifs, then it may be difficult to read as a single expression. When using an if expression that includes elseif clauses is preferred, put the elseif (condition) then on a new line just like then and else.

Then, the remaining options would be to create a local variable, break the style rule and use a chained if-then-else expression, or create a function.

```lua
local CONST_BUT_LOCAL
if cond1 then
CONST_BUT_LOCAL = 1
elseif cond2 then
CONST_BUT_LOCAL = 2
else
CONST_BUT_LOCAL = 3
end
const CONST = CONST_BUT_LOCAL
```

`const CONST = getConst()`

Well, that looks a bit silly.

I always welcome performance improvements too, but I feel like updates should be rolled out more carefully.

Other troublesome cases, if `const` becomes the default convention for values that are only assigned once:
- values decided inside a loop
- values chosen after `pcall` / validation / fallback logic
- values that require statements before assignment in each branch
- multiple values chosen together in one conditional block

```lua
const PATH
if useCache then
    warmUpCache()
    PATH = getCachedPath()
else
    logFallback()
    PATH = computePath()
end

const IS_EDIT_MODE = false
pcall(function()
    IS_EDIT_MODE = game:GetService("RunService"):IsEditMode()
end)

const VALUE
if cond1 then
    local x = computeA()
    local y = normalize(x)
    VALUE = finalizeA(x, y)
elseif cond2 then
    local x = computeB()
    VALUE = finalizeB(x)
else
    VALUE = DEFAULT_VALUE
end
```

then again, mixing local and const to define constants feels like it would just make things more ambiguous...
even if we argue that values determined by branching logic shouldn't be set as CONSTANTs, I'm not really sure where the exact boundary of a 'constant' lies

jLn0n:
const keyword can introduce more optimizations regarding constant-folding and inlining, so no

antiskid4life ezmoney ezlife

Bottersnike:
> adding "syntactic sugar" that serves no functional purpose beyond what table.freeze() or type-checking already provides makes the language feel cluttered.

This isn't what `const` does. `table.freeze` and `const` address fundamentally different problems. If we consider

```lua
const foo = table.freeze({ bar = 5 })
foo = { bar = 6}  -- Const disallows this
foo.bar = 6  -- Frozen tables disallow this
```

For the "problem" raised just a few minutes ago, I'd counter by asking how many of your constant magic numbers are actually dynamic? Beyond say a simple `const IS_DEVELOPMENT = RunService:IsStudio()`.

The style guide linked is very old, and I'd argue if you _really_ wanted something like is proposed there one could write:

```lua
const FOO = (
    if a then 1
    elseif b then 2
    else 3
)
```

which contrary to the suggestion is perfectly readable.

> Other troublesome cases, if const becomes the default convention for values that are only assigned once

This is what has happened in JS land, and I'd argue it's a net positive. It makes marking a variable as "I plan to reassign to this" an active choice, which leads to accidental reassignment (typos, forgetting it was meant to be constant, etc) becoming a warning. This is, of course, purely a personal choice though, and one can just use `local` exclusively if they don't like having to consider if variables should be reassigned.

P.S. (It's worth noting too that there was plenty of off-platform discussion in ROSS about this feature prior to its merge. We're getting "`const`-like" variables regardless from the incoming `export` syntax, so it makes sense to have a non-exported `const` too. `const` + `table.freeze` applied on a variable at the same time also brings the opportunity for some pretty cool optimisations)

MagmaBurnsV:
I'm just going to chime in to mention my thoughts on this, and while I obviously don't support reverting a feature that's already in production, I want to highlight on why I believe merging the const RFC received so much backlash.

The decision to make static exports exhibit const-ness was entirely preferential on the team's part, and I am tired of hearing people parrot the merge reason that adding const was somehow a "dependency" for static exports. This is not true, as there is a fine distinction between "required for this to work" and "nice to have along with," with const being the latter case. If anything, const should have been added *after* static exports were finalized, as implementing it now basically damns the current plans for static exports in place with little hope for amendment.

I also want to mention that nothing about treating exported variables as const would be conducive to better semantics or optimizations, with it in fact probably complicating semantics in regard to aliased shorthand exports and mutually dependent functions. As far as I can see, this was an entirely stylistic choice, which makes the decision for adding const along for the ride more infuriating since it was clearly motivated by personal preference instead of necessity.

I have proposed alternative designs for static exports that omitted the need for adding const-ness to the language, with the team categorically dismissing them for bogus reasons. Now that const is in production, it seems like the current semantics for static exports are also set in motion to be accepted, which is very disappointing. There are many footguns present in the current design which stem directly from treating exported variables as const, and they could have been avoided if our comments were considered. I am seeing a concerning pattern of ignoring community input with recent sweeping language changes here.

andyfriesen:
Hello!

We created this RFC process specifically to gather your input and we're very grateful for it, but we will not be reverting this language feature.

PR #185, opened 2026-03-30
RFC: Table Comprehension

InfraredGodYT:
[Rendered](https://github.com/InfraredGodYT/rfcs/blob/master/docs/table-comprehension.md)

This RFC proposes the feature of table comprehension, similar to languages like Python & MoonScript. This is my first RFC so it's probably pretty rough.

### DIFF
--- a/docs/table-comprehension.md
+++ b/docs/table-comprehension.md
@@ -47,8 +47,8 @@ However, this approach has the following drawbacks:
 - Increases the likelihood of indexing or insertion mistakes
 - Lacks a concise expression form for common transformation patterns
 
-Additionally, these patterns are difficult for the compiler to recognize as a single transformation, which limits opportunities for optimization.
-A table comprehension syntax allows these operations to be expressed as a single construct. While it does not inherently change runtime behavior, it provides the compiler with a clearer representation of the programmer’s intent, which may enable future optimizations such as preallocation or loop specialization.
+Additionally, these patterns are expressed using general-purpose loops, which may make it harder for the compiler to consistently identify them as table transformations.
+A table comprehension syntax allows these operations to be expressed as a single construct. While it does not inherently change runtime behavior, it provides a more explicit representation of the programmer’s intent, which may enable future optimizations such as preallocation or loop specialization.
 
 ## Design
 
--- a/docs/table-comprehension.md
+++ b/docs/table-comprehension.md
@@ -172,6 +172,8 @@ end
 
 The `namelist in explist` portion follows the same semantics as Luau `for` loops, supporting all valid iterator forms including `ipairs`, `pairs`, and custom iterators.
 
+### Grammar
+
 ## Drawbacks
 
 Table comprehension introduces new syntax to the language, increasing overall language complexity. While the feature is designed to be minimal and consistent with existing `for` loop semantics, it adds another way to express iteration and table construction, which may impact readability for developers unfamiliar with the syntax.
--- a/docs/table-comprehension.md
+++ b/docs/table-comprehension.md
@@ -1,17 +1,190 @@
-
 # Table Comprehension
 
 ## Summary
 
-Add table comprehension syntax for constructing arrays and maps using inline iteration and optional filtering expressions.
-
----
+Table comprehension is a specific type of syntax which allows you to construct tables using inline iteration and optional filtering expressions.
 
 ## Motivation
 
-Constructing tables using loops is a very common pattern in Luau. However, this often results in verbose and mutation-heavy code.
+Constructing tables using loops is a very common practice in Luau, particularly for transforming and filtering collections. However, the current approach is verbose, even for simple transformations.
 
-For example:
+Examples:
+```lua
+-- Arrays
+local numbers = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
+local doubled = {}
+for i, v in ipairs(numbers) do
+    doubled[i] = v * 2
+end
+
+local evens = {}
+for _, v in ipairs(numbers) do
+    if v % 2 == 0 then
+        table.insert(evens, v)
+    end
+end
+
+-- Maps
+local playersByName = {}
+for _, player in ipairs(game.Players:GetPlayers()) do
+    playersByName[player.Name] = player
+end
+
+local alivePlayers = {}
+for _, player in ipairs(game.Players:GetPlayers()) do
+    if player.Health > 0 then
+        alivePlayers[player.UserId] = player
+    end
+end
+```
+
+Patterns like these are especially common in Roblox development, especially in cases such as building lookup tables and filtering instances by state.
+
+However, this approach has the following drawbacks:
+- Requires explicit mutation of tables (`table.insert`, index assignment)
+- Introduces boilerplate for simple transformations
+- Obscures intent by separating iteration, filtering, and assignment
+- Increases the likelihood of indexing or insertion mistakes
+- Lacks a concise expression form for common transformation patterns
+
+Additionally, these patterns are difficult for the compiler to recognize as a single transformation, which limits opportunities for optimization.
+A table comprehension syntax allows these operations to be expressed as a single construct. While it does not inherently change runtime behavior, it provides the compiler with a clearer representation of the programmer’s intent, which may enable future optimizations such as preallocation or loop specialization.
+
+## Design
+
+### Syntax
+
+Two forms of table comprehension are proposed.
+
+#### Array-style
+`{expr for namelist in explist [if condition]}`
+
+Examples:
+
+```lua
+local doubled = {v * 2 for i, v in ipairs(numbers)}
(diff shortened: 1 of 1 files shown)


InfraredGodYT:
Marked it as a draft because it will probably require several changes before it's ready.

Cooldude2606:
Just an early comment for consideration as you flesh it out. 

What is the reasoning for why comprehension should not be combined with standard lua fields? It is possible that it could follow the same logic as if the comprehension produces duplicate keys.

Would be nice to see an extended alternative to "helper functions" of including them in the standard lib. For example having `table.map` would enable fast calling and iteration on the C side. Which could see better performance than a lua only helper function, although likely slower than a for loop or comprehension given `table.foreach` is deprecated.

vegorov-rbx:
Fastcall functions can not call user code, yield, fail, or reallocate stack, so if table.map was a library function it would not get special treatment on the C side.

Bottersnike:
Does this RFC address multiple loops and conditions within a single comprehension? `[j for i in array for j in i]` in Python is a very useful way to flatten a 2D array into a 1D one.

If not, something to consider is what order the `for`s would evaluate. Python for example does `[expression] [ordered as if you were writing normal for loops] [conditions]` which can be quite confusing to read.

With Luau in its current state I personally would prefer comprehensions over `table.map` simply due to the cumbersome nature of writing lambdas right now; `table.map(foo, function (x) return x + 1 end)` is quite a mouthful, and the actual "what's going on" (`x + 1`) is pretty nested in there.

Something else to consider is other types of `for` loops. For example, (again using python as an example) you can write `[i + 1 for i in range(10)]` in python, to get an array of `1..10`. Luau doesn't use a magic `range` iterator, instead doing `for i=1,10 do`, so it would be interesting to be able to write `{i+1 for i=1,10}` or similar (to in this case get `2..11`).

Something else to _also_ consider is allowing mixing `for` and `if` statements, rather than exclusively `for`s followed by `if`s. For example, Python allows

```py
>>> a = [[1, 2, 3], None, [4, 5, 6]]
>>> [
...  i
...  for j in a
...  if j
...  for i in j
... ]
[1, 2, 3, 4, 5, 6]
```

These suggestions all depend on how "complex" you want to allow comprehensions to be. Python's can be very complex, and therefore very powerful, but the trade-off comes from potential readability. Formatting them properly aids a lot with this, but that complexity may still be undesirable. As is, too, there's nothing in the proposal that would prevent future expansion to any of the above, so it may be preferable to have first a minimal implementation, followed by more complex extensions later.

Edit: Just as a reference, given I've mentioned Python a lot in this comment, https://docs.python.org/3/reference/expressions.html#grammar-token-python-grammar-comprehension and https://docs.python.org/3/reference/expressions.html#grammar-token-python-grammar-dict_comprehension are the explained grammar for how Python handles comprehensions.

InfraredGodYT:
> Just an early comment for consideration as you flesh it out.
> 
> What is the reasoning for why comprehension should not be combined with standard lua fields? It is possible that it could follow the same logic as if the comprehension produces duplicate keys.
> 
> Would be nice to see an extended alternative to "helper functions" of including them in the standard lib. For example having `table.map` would enable fast calling and iteration on the C side. Which could see better performance than a lua only helper function, although likely slower than a for loop or comprehension given `table.foreach` is deprecated.

I added standard library helpers as an alternative in the RFC.

InfraredGodYT:
> Does this RFC address multiple loops and conditions within a single comprehension? `[j for i in array for j in i]` in Python is a very useful way to flatten a 2D array into a 1D one.
> 
> If not, something to consider is what order the `for`s would evaluate. Python for example does `[expression] [ordered as if you were writing normal for loops] [conditions]` which can be quite confusing to read.
> 
> With Luau in its current state I personally would prefer comprehensions over `table.map` simply due to the cumbersome nature of writing lambdas right now; `table.map(foo, function (x) return x + 1 end)` is quite a mouthful, and the actual "what's going on" (`x + 1`) is pretty nested in there.
> 
> Something else to consider is other types of `for` loops. For example, (again using python as an example) you can write `[i + 1 for i in range(10)]` in python, to get an array of `1..10`. Luau doesn't use a magic `range` iterator, instead doing `for i=1,10 do`, so it would be interesting to be able to write `{i+1 for i=1,10}` or similar (to in this case get `2..11`).
> 
> Something else to _also_ consider is allowing mixing `for` and `if` statements, rather than exclusively `for`s followed by `if`s. For example, Python allows
> 
> ```python
> >>> a = [[1, 2, 3], None, [4, 5, 6]]
> >>> [
> ...  i
> ...  for j in a
> ...  if j
> ...  for i in j
> ... ]
> [1, 2, 3, 4, 5, 6]
> ```
> 
> These suggestions all depend on how "complex" you want to allow comprehensions to be. Python's can be very complex, and therefore very powerful, but the trade-off comes from potential readability. Formatting them properly aids a lot with this, but that complexity may still be undesirable. As is, too, there's nothing in the proposal that would prevent future expansion to any of the above, so it may be preferable to have first a minimal implementation, followed by more complex extensions later.
> 
> Edit: Just as a reference, given I've mentioned Python a lot in this comment, https://docs.python.org/3/reference/expressions.html#grammar-token-python-grammar-comprehension and https://docs.python.org/3/reference/expressions.html#grammar-token-python-grammar-dict_comprehension are the explained grammar for how Python handles comprehensions.

This RFC currently only targets a minimal form of table comprehension. At the moment, the proposal only supports a single `for` clause and an optional trailing `if` condition. More complex forms like multiple `for` clauses or interleaved conditions aren't included for now to keep the feature simple and readable.

Do you think starting with this minimal subset makes sense?

Bottersnike:
> Do you think starting with this minimal subset makes sense?

I'm honestly unsure! As-is, it doesn't restrict their addition in the future, which is really good.

I really couldn't say if it makes more sense to have one uber-fleshed-out RFC or one that's a starting point and a second that expands it. The benefit of just starting with a simple RFC is it's potentially more likely to get merged, but the risk of hoping to get a second merged too is if that second _doesn't_ you can end up with a half-baked feature.

InfraredGodYT:
> I really couldn't say if it makes more sense to have one uber-fleshed-out RFC or one that's a starting point and a second that expands it. The benefit of just starting with a simple RFC is it's potentially more likely to get merged, but the risk of hoping to get a second merged too is if that second _doesn't_ you can end up with a half-baked feature.

That makes sense. I think it’s safer to establish a simple baseline first and then explore more complex extensions separately if there’s interest.

For now, I’m planning to keep the RFC minimal to keep the feature easy to understand and reason about. That said, I’m open to expanding the scope if there’s strong feedback that the minimal version would be too limited.

deviaze:
I always found the `[ x for y in whatever ]` syntax quite confusing once you got into nested comprehensions, so I recommend a syntax that aligns with regular Luau's if expressions (in Python, you have `x if y else z` but in Luau you have `if y then x else z`) and preserves readability with nested comprehensions:

```luau
-- simple
local luau_files: { string } = { 
    for _, filename in fs.listdir("./some/path") do
        if string.match(filename, "%.luau$") then
            filename
        else -- like with regular if expressions outside table comprehensions, every if expression must have an else clause
            continue
}
```

This makes nested comprehensions much more readable (even if they're more verbose)

```luau
-- nested
local results = {} :: { [string]: { string } }
local elements_with_meow = {
    for name, list in results do
        for _, element in list do
            if string.match(element, "meow") then
                element
            else
                continue     
}
```

I actually wrote a whole table comprehensions RFC on this a few years ago (that also clearly describes the semantics of `break` and `continue` within comprehensions) but I can't seem to find it anymore :(

As a nit, I recommend removing the uses of `pairs` and `ipairs` in the draft since they're \*basically\* deprecated in favor of generalized iteration (no iterator function needed) in Luau.

InfraredGodYT:
I think that kind of design could be worth exploring, but it would likely be better suited as a separate RFC rather than being combined into this one. For this RFC, I’m intentionally keeping the design minimal, while leaving room for more complex forms like nested comprehensions to be explored as future extensions if there’s interest.

I'll also remove the use of `pairs` and `ipairs` in the next commit.

Brawldude2:
What does it do other than bloating syntax tree and ruining the intuitive design principle. It doesn't take 2 lines to do the same thing without additional syntax

InfraredGodYT:
> What does it do other than bloating syntax tree and ruining the intuitive design principle. It doesn't take 2 lines to do the same thing without additional syntax

While this RFC doesn't introduce new capabilities, the goal of it is to provide a more concise way to represent common table transformation patterns. These transformations can be written within a few lines today, but usually involve boilerplate that can obscure the intent of the code (e.g. table initialization, mutation, and conditional insertion). Using less than 2 lines for said common transformations will make the code unnecessarily more difficult to read, and this proposal focuses on making those patterns easier to express inline rather than expanding what the language can do.

InfraredGodYT:
Pretty sure this RFC is a lost cause so I'm closing it.

PR #186, opened 2026-03-31
RFC: Export Type Barrel

StoryBakery-1:
[Rendered](https://github.com/StoryBakery/luau-rfcs/blob/syntax-type-export-barrel/docs/syntax-type-export-barrel.md)

This RFC proposes a type-only barrel form for forwarding exported types from one module through another:

```luau
-- Child.luau
export type A<T, U> = ...
export type B = ...
export type C = ...

-- Interface.luau
export type * from "@self/Child"

-- Consumer.luau
local Interace = require("./Interface")

local a: Interace.A<any, any>
local b: Interace.B
local c: Interace.C
````

### DIFF
--- a/docs/syntax-type-export-barrel.md
+++ b/docs/syntax-type-export-barrel.md
@@ -1,0 +1,273 @@
+# Export Type Barrel
+
+## Summary
+
+Add a top-level declaration that re-exports all names from another module’s exported type namespace.
+
+```luau
+export type * from "@self/Child"
+```
+
+This declaration makes every exported type from the target module available as an exported type of the current module, 
+without creating a runtime binding and without affecting the module's returned value.
+
+## Motivation
+
+Today, exported types are accessed through the alias of a directly required module:
+
+```luau
+local Child = require("../pkgs/Interface/Child")
+
+local a: Child.A
+local b: Child.B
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Crazyblox** [comment]
This might not be what you are looking for, but I wanted to explore syntax for this on the side of the requirer rather than the export.

```luau
-- All export types: Assign to requirer top level.
type require("path/to/module")

-- All export types: Assign to requirer top level with option to reassign type names, preventing clashing.
type SpecificType = string
type require("path/to/module") {RenamedSpecificType = SpecificType}

-- Single export type: Assign to requirer top level.
type require("path/to/module").SpecificType

-- Single export type: Assign to requirer top level with option to reassign, preventing clashing.
type SpecificType = string
type RenamedSpecificType = require("path/to/module").SpecificType
```

TenebrisNoctua:
Couldn't this be achieved in a more simplistic manner as:

```luau
local module = require("path/to/module")
export type module -- All types from the module can now be exported in this simplistic syntax.
```

You could still do:

```luau
local module = require("path/to/module")
export type A = module.A
```

For better control. This would use the existing type resolving system with `require()`, and may be less expensive to implement.

PR #187, opened 2026-04-04
RFC: function `coroutine.finally(thread, callback)`

jkelaty-rbx:
status: implementing

[Rendered](https://github.com/jkelaty-rbx/rfcs/blob/coroutine-finally/docs/function-coroutine-finally.md)


docs/function-coroutine-finally.md
```diff
@@ -0,0 +1,199 @@
...
+coroutine.finally(co: thread, callback: (boolean, any) -> ())
+```
+
+### Callback invocation
+
+The callback is called with two arguments `(ok, err)`:
+
+- **Normal return:** `callback(true, nil)`. The coroutine's function returned. Return values are not forwarded to the callback; they are available to the caller through `coroutine.resume`'s own return values.
+- **Unhandled error:** `callback(false, err)`. The coroutine terminated with an error. The error value is passed as the second argument **when available.**
+- **External close:** `callback(false, "coroutine was closed")`. `coroutine.close` was called on the coroutine. The coroutine did not complete its work. A default error message is provided so callbacks can identify cancellation without external bookkeeping.
```
MagmaBurnsV, line 29: Would it be possible to instead return a status string detailing if the thread finished normally/failed/was cancelled, instead of a boolean? Or is this information lost after the thread is dead?

Because right now it's essentially impossible to differentiate between a closed thread and a thread that errored with the message `"coroutine was closed"`.

docs/function-coroutine-finally.md
```diff
@@ -0,0 +1,199 @@
...
+
+### When callbacks fire
+
+Callbacks fire synchronously after the coroutine has transitioned into its terminal state but before `coroutine.resume` or `coroutine.close` returns to its caller. The thread that called resume or close executes the callbacks. If `coroutine.finally` is called on a coroutine that is already dead, the callback fires immediately before `coroutine.finally` returns (see below).
+
+After all callbacks have fired, they are removed and will not fire again. Each callback fires exactly once, or not at all if the coroutine yields indefinitely.
+
+### Already-dead coroutines
+
+If `coroutine.finally` is called on a coroutine that is already dead, the callback fires immediately on the thread that called `coroutine.finally`. The callback receives `(true, nil)` if the coroutine finished normally, or `(false, nil)` if it errored or was closed. The original error value is no longer available because the coroutine has already been reset.
```
MagmaBurnsV, line 45: It might be better to just error if the thread is already dead, since this behavior might lead to subtle bugs where you didn't expect the callback to run on the same thread.

You can still get this functionality back by checking the coroutine status and running the function inline, after all.

docs/function-coroutine-finally.md
```diff
@@ -0,0 +1,199 @@
...
+- **Unhandled error:** `callback(false, err)`. The coroutine terminated with an error. The error value is passed as the second argument **when available.**
+- **External close:** `callback(false, "coroutine was closed")`. `coroutine.close` was called on the coroutine. The coroutine did not complete its work. A default error message is provided so callbacks can identify cancellation without external bookkeeping.
+
+`coroutine.finally` is a control-flow abstraction, not a data-flow abstraction. The boolean is the only reliable signal. The second argument is an error diagnostic: for unhandled errors and external close it is available when the callback fires inline during `coroutine.resume` or `coroutine.close`, but is not guaranteed in all cases (see "Already-dead coroutines" below). Callers that need the coroutine's return values should use `coroutine.resume`'s own returns.
+
+### Multiple callbacks
+
+Multiple callbacks can be registered on the same coroutine. They are called in LIFO order (last registered, first called). Each callback is implicitly called via `pcall`; if a callback errors, the error is discarded and the remaining callbacks still execute.
```
MagmaBurnsV, line 35: Can the callbacks yield? I assume not since other callbacks have to also run synchronously.

Cooldude2606:
Overall I find this well written and a useful language level feature. However, silencing errors would be a deal breaker for me because it hinders not only debugging but also the detection of issues. For example, if most finally callbacks perform clean up actions then their failure may not introduce logical errors but rather memory leaks; which would be challenging to identify then trace back to the callback.

Propagating the error to the caller of resume or close would be preferable, which can be pcalled at that point if desired.

Although an alternative of replacing the status and message provided to proceeding callbackss could be a consideration. This would be similar to xpcall recalling the handler with its own error if the handler errors. But this would remove the ability to inspect the callstack through a custom xpcall error handler.

docs/function-coroutine-finally.md
```diff
@@ -0,0 +1,199 @@
...
+
+```luau
+coroutine.finally(co: thread, callback: (boolean, any) -> ())
+```
+
+### Callback invocation
+
+The callback is called with two arguments `(ok, err)`:
+
+- **Normal return:** `callback(true, nil)`. The coroutine's function returned. Return values are not forwarded to the callback; they are available to the caller through `coroutine.resume`'s own return values.
```
bradsharp, line 27: If finally provided the values that are still on the stack of the given thread it would allow you to implement await.

```luau
local function await(thread)
    local result = nil
    coroutine.finally(thread, function (...)
        result = table.pack(...)
    end)
    if not result then
        -- yield until the other thread has finished
        local current = coroutine.running()
        coroutine.finally(thread, function ()
            coroutine.resume(current)
        end)
        coroutine.yield()
    end
    local success = result[1]
    if not success then
        error(result[2])
    end
    return table.unpack(result, 2)
end
```

I don't think this needs to be perfect - the guarantee can be "whatever is left on the stack". A library can handle cases where the stack is later cleared by caching the result.

docs/function-coroutine-finally.md
```diff
@@ -0,0 +1,199 @@
...
+
+`coroutine.finally(co, callback)` registers a callback on coroutine `co` that will be called when `co` reaches a terminal state.
+
+```luau
+coroutine.finally(co: thread, callback: (boolean, any) -> ())
+```
+
+### Callback invocation
+
+The callback is called with two arguments `(ok, err)`:
```
bradsharp, line 25: Is the callback allowed to yield? My initial reaction is no and we could always allow it later if we think it's okay.

alicesaidhi:
i think it'd be more ideal instead of returning a boolean, it'd return a string literal between "error", "finished", "cancelled"

gaymeowing:
> i think it'd be more ideal instead of returning a boolean, it'd return a string literal between "error", "finished", "cancelled"

Would be much better than giving a `"coroutine was closed"` string in the same spot the error argument is, as the error could be `"coroutine was closed"`. This could be an issue for things like modding apis in games, or games that allow for using luau to do custom logic similar to Stormworks.

docs/function-coroutine-finally.md
```diff
@@ -0,0 +1,201 @@
...
+    end
+end)
+
+coroutine.resume(co)
+-- Both callbacks fire (LIFO: B runs first, then A).
+```
+
+## Drawbacks
+
+This adds per-coroutine state (a callback list stored in a weak-keyed registry table), which slightly increases memory usage for coroutines that register finally callbacks. Coroutines without registered callbacks pay no cost beyond the registry table lookup on terminal events.
```
vegorov-rbx, line 179: This is an implementation detail, but weak-keyed table in the registry will not work when callbacks capture the thread.
Luau doesn't implement ephemeron tables so weak key still contains a strong value to that key.

In the actual implementation I will suggest storing a table inside lua_State.

docs/function-coroutine-finally.md
vegorov-rbx: My suggestion is to include in the proposal that callbacks are allowed to yield.

Callbacks are not metamethods, so the restriction doesn't apply.
I see no technical/implementation reason to not allow it.
It will make the language library more flexible.

docs/function-coroutine-finally.md
vegorov-rbx: I suggest that yes, callbacks can yield.

docs/function-coroutine-finally.md
```diff
@@ -0,0 +1,201 @@
...
+- **Unhandled error:** `callback(false, err)`. The coroutine terminated with an error. The error value is passed as the second argument **when available.**
+- **External close:** `callback(false, "coroutine was closed")`. `coroutine.close` was called on the coroutine. The coroutine did not complete its work. A default error message is provided so callbacks can identify cancellation without external bookkeeping.
+
+`coroutine.finally` is a control-flow abstraction, not a data-flow abstraction. The boolean is the only reliable signal. The second argument is an error diagnostic: for unhandled errors and external close it is available when the callback fires inline during `coroutine.resume` or `coroutine.close`, but is not guaranteed in all cases (see "Already-dead coroutines" below). Callers that need the coroutine's return values should use `coroutine.resume`'s own returns.
+
+### Multiple callbacks
+
+Multiple callbacks can be registered on the same coroutine. They are called in LIFO order (last registered, first called). Each callback is implicitly called via `pcall`; if a callback errors, the error is discarded and the remaining callbacks still execute.
```
vegorov-rbx, line 35: I will push back on this, discarding errors is not a good design decision.
I propose that errors will error
- `coroutine.resume` will return false with the produced error
- `coroutine.close` will return false with the produced error
This has similarity with xpcall erroring on error handling, resulting in 'error in error handling'.

docs/function-coroutine-finally.md
vegorov-rbx: I agree with Brad, it will be useful to provide results to coroutine callbacks and while thinking of the implementation path, I don't see a technical reason not to.

docs/function-coroutine-finally.md
vegorov-rbx: Agree with MagmaBurnsV here, erroring here is much better.

It makes the whole system have less exceptions:
- you don't get states where results or errors have become unavailable
- callback is not called on a separate control-flow path, different environment from resume/close location might cause divergent behavior of the callback (on top of divergence caused by error not being recoverable)
- natively managed coroutine could have been reused, so patterns installing callbacks late have an extra footgun of installing on running coroutines that are logically different from the one before resume call on it

Coroutine is a lower level primitive, so I would not expect Promise semantics from it, but a promise can be built on top.

docs/function-coroutine-finally.md
vegorov-rbx: Might be a good idea, while a small departure from 'pcall' API, coroutine API is already richer with coroutine.status, so having "completed"/"cancelled"/"errored"

It does have a drawback of a bit harder distinction than `if success then else end` when you don't care about the 'fail' state details.
Also might need bikeshdding on what words are passed in.

docs/function-coroutine-finally.md
```diff
@@ -0,0 +1,201 @@
...
+
+Structured concurrency requires two fundamental operations on a unit of work: **cancel** and **join**. `coroutine.close` provides cancel. `coroutine.finally` provides join. Together they form the minimal pair needed to build higher-level primitives (nurseries, task groups, cancellation trees) entirely in userspace, anchored to the `thread` type that the VM already uses as its unit of execution.
+
+Because the hook is registered on the coroutine directly, not on a wrapper object, it enables interoperability by default. A task scheduler can register a `finally` callback to track completion while a separate logging library registers its own callback on the same coroutine for diagnostics, with no coordination between the two.
+
+## Design
+
+`coroutine.finally(co, callback)` registers a callback on coroutine `co` that will be called when `co` reaches a terminal state.
```
vegorov-rbx, line 17: Can we add a note that makes it error if called on VM main thread?
In many environments you can't even access it, so it's a small thing, but might save some C API user from confusion.

docs/function-coroutine-finally.md
vegorov-rbx: Updated the RFC to provide a string status.

docs/function-coroutine-finally.md
vegorov-rbx: Updated the RFC to error on dead coroutines.

docs/function-coroutine-finally.md
vegorov-rbx: Updated the RFC to allow the callbacks to yield.

docs/function-coroutine-finally.md
vegorov-rbx: Updated the RFC so that completed coroutine results are provided to each callback.
And as soon as callbacks begin to execute, it no longer matters what happens to the completed thread, it can even be reset and reused.

docs/function-coroutine-finally.md
vegorov-rbx: Updated the RFC to allow callback yielding.

docs/function-coroutine-finally.md
vegorov-rbx: Storage has been changed in the implementation to avoid such issues while requiring a bit more thread memory as a compromise.

docs/function-coroutine-finally.md
vegorov-rbx: Updated RFC no longer discards callback errors.
Callbacks can still choose to use `pcall`/`xpcall` to self-ignore their own mistakes.

vegorov-rbx:
> However, silencing errors would be a deal breaker for me because it hinders not only debugging but also the detection of issues.

RFC has been updated to no longer silence callback errors.

> i think it'd be more ideal instead of returning a boolean, it'd return a string literal between "error", "finished", "cancelled"

Changed from boolean status to a string.

docs/function-coroutine-finally.md
```diff
@@ -0,0 +1,314 @@
...
+
+Coroutine result or error is discarded if a callback errors.
+Callback erroring also prevents other callbacks from running.
+An alternative would be to aggregate errors or propagate the original one of the completed coroutine, at the cost of a more complex error model.
+
+If a coroutine is abandoned (becomes unreachable without being closed or run to completion), its finally callbacks are collected by the GC and never fire.
+
+## Prior Art
+
+Python has [`asyncio`](hhttps://docs.python.org/3/library/asyncio-task.html) where it's possible to attach completion callbacks to a scheduled task.
```
PhoenixWhitefire, line 280:
```suggestion
Python has [`asyncio`](https://docs.python.org/3/library/asyncio-task.html) where it's possible to attach completion callbacks to a scheduled task.
```

PR #189, opened 2026-04-04
Signed and Unsigned Integer Subtype Refinements

Crazyblox:
[Rendered](https://github.com/Crazyblox/rfcs/blob/rfc-integer-refinement/docs/type-integer-refinement.md)

### DIFF
--- a/docs/type-integer-refinement.md
+++ b/docs/type-integer-refinement.md
@@ -1,0 +1,81 @@
+# Signed and Unsigned Integer Subtype Refinements
+
+## Summary
+This proposal introduces `signed` and `unsigned` as subtype refinements of the existing `integer` primitive type. By establishing a type hierarchy where `signed <: integer` and `unsigned <: integer`, the Luau type checker can distinguish between integers intended for signed operations (e.g., `integer.lt`) and those intended for unsigned operations (e.g., `integer.ult`). This allows the `integer` library to enforce type safety at the type-checking level, preventing the accidental use of signed integers in unsigned contexts and vice versa.
+
+## Motivation
+With the introduction of the `integer` primitive and the `integer` library, Luau provides powerful tools for low-level integer manipulation. However, a critical gap exists: there is currently no way to communicate or validate whether an integer is signed or unsigned within the type system.
+
+Currently, a user could pass a negative integer to `integer.ult` (Unsigned Less Than), and the type checker would not flag this as an error because both arguments are simply typed as `integer`. This shifts the burden of validation entirely to the developer or results in unexpected runtime behavior.
+
+By introducing subtype refinements, we can:
+1. **Prevent Logic Errors:** Catch cases where a signed value is passed to an unsigned function before the code ever runs.
+2. **Improve Documentation:** Type signatures for library functions become self-documenting (e.g., seeing `unsigned` immediately tells the developer the expected range).
+3. **Reduce Runtime Overhead:** By validating signedness at the type-checking stage, we reduce the need for expensive runtime assertions within the `integer` library.
+
+## Design
+We propose the introduction of two new type refinements: `signed` and `unsigned`. These are not separate primitive types, but rather specialized versions of the `integer` type.
+
+### Type Hierarchy
+The relationship is defined as:
+- `signed` as subtype of `integer`
+- `unsigned` as subtype of `integer`
+
+### Integration with the Integer Library
+The `integer` library functions will be updated to accept these refined types. To ensure a smooth transition and avoid unnecessary casting for general-purpose integers, functions will accept either the refined subtype OR the base `integer` type. When the base `integer` type is provided, the type checker will implicitly refine it to the required subtype for that operation.
+
+**Example Signatures:**
+- `integer.lt(a: signed, b: signed): boolean`
+- `integer.ult(a: unsigned, b: unsigned): boolean`
+
+### Behavior and Examples
+
+#### 1. Implicit Refinement
+If a variable is typed as the general `integer` type, it can be used in either signed or unsigned functions. The type checker treats this as a valid "promotion" to the required refinement.
+
+```luau
+local x: integer = 10i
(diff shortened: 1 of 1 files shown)


docs/type-integer-refinement.md
```diff
@@ -0,0 +1,81 @@
...
+local function processUnsigned(val: unsigned)
+    print(integer.ult(val, 100i))
+end
+
+local a: integer = 10i
+local b: signed = -10i
+
+processUnsigned(a) -- Valid
+processUnsigned(b) -- Type Error: Expected unsigned, got signed
+```
```
MagmaBurnsV, line 78: Is `a` supposed to be annotated as unsigned? Because it doesn't make sense to allow integer in a place that expects unsigned if unsigned <: integer.

Crazyblox: Should be resolved, along with an additional example for clarity.

PhoenixWhitefire:
Not sure about the naming; `signed`/`unsigned` is not very descriptive to me. Maybe it could be `(un)signedinteger` instead. Consider noting the alternative names in the `Alternatives` section with a reason for why it was not chosen 

EDIT: Resolved (maybe I should've made this a Review Comment...)

Bottersnike:
How is
```lua
-- Accepts 'integer' and its subtypes
local function processInteger(val: integer)
    -- Type errors if val was passed as 'signed' integer subtype 
    print(integer.ult(val, 100i)) -- 'integer' refines to 'unsigned'
end

local a: integer = 10i
local b: integer = -10i

processUnsigned(a)    -- Type Error: Expected unsigned, got integer
processInteger(b)     -- Valid ('integer' refines to 'unsigned')
```
intended to function? Both `a` and `b` have been cast to `integer`, so we lost any knowledge regarding if they were signed or unsigned. Inside `processInteger`, `val` is now also just an `integer` (`signed | unsigned`) but the key problem we face is there _is_ no way to narrow a bag of 64 bits to "signed" or "unsigned" once we've thrown away that metadata. `ult` would fail there because it required `val` to be `unsigned` but got `signed | unsigned`, but we have no way for the function to do `Type errors if val was passed as 'signed' integer subtype` as described, so we'd need to explicitly do `integer.ult(val :: unsigned, 100i)`.

Something we'd also want to consider is how do constant integers infer? If we have `ult(5i, 10i)`, is that a type error? If we have `-5i` we can guaranteeably type that as `signed`, and if we have a positive integer outside of signed range, that's `unsigned`, but integers between `0` and `2^63-1` can be represented with both. Do we concede that we would always have to write `ult((5i) :: unsigned, (10i) :: unsigned)` or would special-case logic be required for inference of integer constants used as literals in a function call?

---

Reading further down in the RFC actually there's

> If a variable is typed as the general integer type, it can be used in either signed or unsigned functions. The type checker treats this as a valid "promotion" to the required refinement.

How would we represent this? `integer = signed & unsigned`? Purely conceptually, the intersection of signed and unsigned would be any number `[0, 2^63-1]`, but notably _not_ `[-2^63, -1]` (signed only) nor `[2^64, 2^64-1]` (unsigned only). Is that inaccuracy something that would be considered just waved away?

I do think waving away that conceptual inaccuracy would resolve the issues suggested above, with a literal would always infer as `integer`. It still leaves integers in a position where a single use of `integer` instead of `signed` or `unsigned` suddenly loses all signed-ness information, though I think that battle was lost when integers were added in their current state.

It would be interesting to see some discussion of inference in the RFC in general. Inferring a negative constant as always `signed`, for example, would be very useful.

---

Just one final edit: This is quite a wall of text, but I'm very much in support of adding proper signed and unsigned types!

docs/type-integer-refinement.md
```diff
@@ -0,0 +1,89 @@
...
+
+integer.lt(s, s)  -- Valid
+integer.ult(s, s) -- Type Error: Expected unsigned, got signed
+
+integer.ult(u, u) -- Valid
+integer.lt(u, u)  -- Type Error: Expected signed, got unsigned
+```
+
+#### 3. Function Overloads/Refinement
+Functions can be written to accept the base `integer` type and refine it internally, or specify the refinement to constrain the input.
```
Bottersnike, line 59: I think it's important to note that
> and refine it internally

isn't possible. There's no way at runtime to ever discern between a `signed` or `unsigned` integer, and therefore once the types no longer carry that information they likewise have no reasonable checks that could be used as part of refinemt.

Crazyblox: Providing the equivalent of `:: signed` or `:: unsigned` within the relevant integer library functions would be sufficient to avoid runtime behaviour, which the RFC is stated multiple times to not involve.

Bottersnike: I was meaning more in as much as any code that takes an `integer` can't safely know _to_ cast it to one or the other. I'm not suggesting types should leak to runtime, but rather that there's no runtime behaviour that could be used to determine if/how an integer should be refined. A quick `::signed` or `::unsigned`, in the same way you can do

```c
int64_t foo = -1;
uint64_t bar = (uint64_t)foo;
```

In C is fine, but I'm not sure if it's worth mentioning given that's not something that can be "safely" done, except when intentionally breaking rules (bit maths on negative numbers, etc.). Would `:: signed` and `:: unsigned` not work fine here?

InfraredGodYT:
> [Rendered](https://github.com/Crazyblox/rfcs/blob/rfc-integer-refinement/docs/type-integer-refinement.md)

I'm not sure how I feel about Luau supporting integers, but since that's already been merged, there will definitely be cases where you need to differentiate between signed and unsigned integers. I'd be baffled if this didn't get merged after integers are fully implemented.

Crazyblox:
This PR has been closed; I no longer believe what I proposed to be a valid approach.

The luau team has created an [integer type discussion megathread](https://github.com/luau-lang/luau/discussions/2574) largely surrounding potential use-cases, please share them if you have any.

PR #193, opened 2026-04-08
We will not be implementing shared self types.

andyfriesen:
I got a prototype of this working, but it doesn't actually work all that well in practice.  The resulting system is very brittle and tricky to work with.  Trickier, in fact, than the pattern that developers are already writing today.


deviaze:
What will happen with the existing, hard-to-read type hovers users see today with metatable based oop classes?

TenebrisNoctua:
Since the merge of the classes RFC, shared-self-types may have almost completely lost its purpose. This should be reviewed and possibly merged, in my opinion.

vegorov-rbx (approved):

PR #196, opened 2026-04-12
Parameter names in types.newfunction

DavidLoveLuau:
**RFC: Parameter names in `types.newfunction`**

This RFC proposes adding the ability to specify parameter **names** when dynamically constructing function types using `types.newfunction()`

This will significantly improve readability and developer experience when working with dynamically generated function types in User-Defined Type Functions

**RFC file:** [`docs/types-newfunction-named-parameters.md`](https://github.com/DavidLoveLuau/rfcs/blob/types-newfunction-named-parameters/docs/types-newfunction-named-parameters.md)


itsmath:
Also, as time goes on since I made my RFC on the same topic, I'm starting to think that this is only going to get accepted if there's a nicer solution that doesn't break backwards compatibility (you say this solution doesn't break it, but it does, specifically because of the `:parameters()` function).

So I'm thinking that we'll probably need to do something closer to my original idea using SoA instead of AoS

DavidLoveLuau:
> Also, as time goes on since I made my RFC on the same topic, I'm starting to think that this is only going to get accepted if there's a nicer solution that doesn't break backwards compatibility (you say this solution doesn't break it, but it does, specifically because of the `:parameters()` function).
> 
> So I'm thinking that we'll probably need to do something closer to my original idea using SoA instead of AoS

Thanks for the feedback!

I agree that backwards compatibility is important. While my current proposal tries to extend the existing `setparameters()` / `:parameters()` API with minimal disruption, I also see the potential issues you mentioned.

Your original idea using SoA style instead of AoS actually sounds quite clean and might indeed be a nicer long-term solution. I really like that direction as well.

Honestly, I would be happy if **either** of our approaches gets accepted

PR #197, opened 2026-04-17
Add RFC: match expression syntax

IllusionAC:
Adds a new RFC proposing a Luau match expression for ordered pattern-based branching over one or more values.

Supports single-value and multi-value matching
Supports wildcard _ patterns
Supports expression-list results, including multiple return values
Includes semantics, type-checking expectations, restrictions, drawbacks, alternatives, and future work

Notes
This RFC is added as a new document: docs/syntax-match-expression.md.

### DIFF
--- a/docs/syntax-match-expression.md
+++ b/docs/syntax-match-expression.md
@@ -1,0 +1,324 @@
+# Match Expression
+
+## Summary
+
+This RFC proposes a `match` expression for Luau.
+
+The goal is to make multi-branch value selection clearer when logic depends on one or more values and branch order matters.
+
+The construct supports:
+
+- matching against one or more values
+- positional patterns for multi-value matching
+- wildcard `_` patterns
+- expression-list results (including multiple return values)
+
+This is intended as a structured alternative to repeated `if` / `elseif` comparisons in expression position.
+
+## Motivation
+
+Luau already supports `if` expressions:
+
+```lua
+local result = if a == 1 then
+	5
+elseif b == 2 then
+	6
+else
+	7
+```
+
+For single conditions this is fine. For multi-value branching, conditions become repetitive and less readable:
+
+```lua
+local result
+if a == 1 and b == 0 then
+	result = 5
+elseif a == 0 and b == 2 then
+	result = 6
+else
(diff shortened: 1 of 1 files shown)


deviaze:
This proposed syntax isn't any improvement over`if-else` expressions and would only make Luau *less* readable. Did you see [the previous RFC](https://github.com/luau-lang/rfcs/pull/100), which approached `match` expressions in a more useful way?

IllusionAC:
Hey, i did read **your** comment on the previous RFC. If you read mine, you'd see that I propose more things while only adding two keyword (match ... with ... end, written the same as if ... then ... end or for ... do ... end etc)
Secondly, if you read mine correctly, if-else is just reusing existing syntax. We could just use if else instead of creating a third word like "case" or others.
I'm not very fond of the "=>" syntax, and you are missing tuple destructuring.
These are the main reason I decided to make a new post instead of reusing the old one, that hasn't been updated in a year.

Cooldude2606:
I prefer the reuse of "if then else" used here. But as a recent comment on the other PR suggests, we likely only need a single keyword to be added, dropping "with" or could reuse "in":

```lua
local value = match a, b in
    if 1, _ then 1
    elseif _, 1 then 2
    else 3
end
```

A more major comment, in agreement with deviaze, is that in its current state this RFC isnt more than if-then-else expressions that support tupple setting. I would at least like to see some of the other aspects of the other RFC brought here, namely: group matching ("a" or "b") and negation (not "a") which arent discussed here at all.

IllusionAC:
I think replacing `with` with `in` is a good idea. For the fact that yes, it looks like a fancy **_if-then-else_**, maybe then we can update the current ternary expression implementation that is `local a = if b then c else d` by adding what pattern matching would normally allow. This would still be pattern-matching without creating a "match" keyword.

TechEverythingTikTok:
It's a really good idea if it actually is fast.. For example, vLua which uses FiOne behind the hood does some smart little optimization, out of 38 instructions it halves it every time until it finds the correct instruction, but a dedicated switch that is instant would always be superior.

PR #200, opened 2026-05-07
nominal interface inheritance

skberkeley:


docs/classes-interface-inheritance.md
```diff
@@ -0,0 +1,243 @@
...
+	
+	function meow(self): string
+		return string.rep("meow\n", meowMult)
+	end
+end
+```
+
+## Motivation
+
+Other languages with support for object-oriented programming support inheritance in some way. We (the Luau team) have even written libraries which would benefit from having classes with nominal inheritance. Interfaces also make it easier to write more modular code, and can provide another guidepost for programmers coming from other languages.
```
alexmccord, line 34:
```suggestion
Object-oriented programming languages support inheritance in some way. There are libraries which would benefit from having classes with nominal inheritance. Interfaces also make it easier to write more modular code, and can provide another guidepost for programmers coming from other languages.
```

docs/classes-interface-inheritance.md
```diff
@@ -0,0 +1,243 @@
+# Nominal Interface Inheritance for Classes
+
+## Summary
+
+```luau
+-- All fields are public by default
```
alexmccord, line 6: Interface types don't have access modifiers, do they? So there's no default. It is _just_ `public`. That's the whole point behind interfaces.

docs/classes-interface-inheritance.md
```diff
@@ -0,0 +1,243 @@
...
+interface Animal = { 
+	species: string, 
+	__tostring: (self) -> string
+}
+
+interface Reptile : Animal {
+	species: string,
+	__tostring: (self) -> string,
+	sunbathe: (self) -> ()
+}
```
alexmccord, line 16: So, this is strictly a little off. There's a difference between `__tostring: (self) -> string` and `read __tostring: (self) -> string`. Which is such an annoying footgun. In _just_ two declarations, this mistake was made three times, more if I counted everything below.

`class Cat: Animal` should fail to typecheck because `Cat` defines a `function __tostring(self) -> string` which is different from the one in `Animal`: is it a field that holds `__tostring` or a `function` named `__tostring`? This is indistinguishable. You could try to pattern match on the type being a function type with the first argument being `self`, but that doesn't work in the general case. Rename the `__tostring` to something else random, and remove the `self` argument, now it's indistinguishable from a callback or a function that takes no arguments, or variations thereof where the function takes any number of arguments. The counterexample:

```luau
interface A = {
  f: () -> (),
}

class Foo: A
  public f: () -> ()
end

class Bar: A
  public function f() end
end

const function f(x: A, message: string)
  x.f = function() print(message) end
  return x
end

const function noop() end
f(Foo { f = noop }, "hello, world!").f() -- fine
f(Bar {}, "kaput") -- not fine.
```

If you have `x : A`, the type checker allows you to write `x.f = function() end`. That's sound if `x is Foo` because it's just a mutable field, unsound if `x is Bar` because it's a function which is immutable.

docs/classes-interface-inheritance.md
```diff
@@ -0,0 +1,243 @@
...
+```
+
+## Motivation
+
+Other languages with support for object-oriented programming support inheritance in some way. We (the Luau team) have even written libraries which would benefit from having classes with nominal inheritance. Interfaces also make it easier to write more modular code, and can provide another guidepost for programmers coming from other languages.
+
+## Design
+
+### Syntax
+Interfaces are introduced with the new `interface` keyword. Like classes, they can only be declared at the topmost scope, and `export interface X = ...` is permitted.
```
alexmccord, line 39: This is an argument for why I don't want `export <name>` syntax. If we added interfaces later, `export interface Foo = { ... }` would not be backward compatible.

docs/classes-interface-inheritance.md
```diff
@@ -0,0 +1,243 @@
...
+Like classes, defining two classes with the same name in the same module is not allowed.
+
+The right hand side of an interface declaration is almost syntactically identical to a table type, with the exception that there are no read/write access modifiers. Its contents are the fields and methods that the interface declares.
```
alexmccord, line 42: `read`/`write` access modifiers are actually valid in interfaces. This sentence is only true for at least as long as classes don't currently split read/write types, but you can imagine a lazy initialization where the write type does not allow the field to be set to `nil`, or with getters/setters.

```luau
type State = { x: number, y: string }

class Example
  public read state: State?
  public write state: State
  
  public function init_state(self)
    self.state = expensive_computation()
  end
end
```

docs/classes-interface-inheritance.md
```diff
@@ -0,0 +1,243 @@
...
+
+```luau
+interface Inter1 = {}
+
+class.isinstance(Inter1, Inter1) -- true
+
+interface Inter2 : Inter1 = {}
+
+class.isinstance(Inter2, Inter1) -- true
+class.isinstance(Inter1, Inter2) -- false
```
TenebrisNoctua, line 66: This method looks confusing to me, interfaces aren't class instances? Why shouldn't we have a new method inside the class library or a property inside interface objects instead?

docs/classes-interface-inheritance.md
```diff
@@ -0,0 +1,243 @@
...
+
+## Design
+
+### Syntax
+Interfaces are introduced with the new `interface` keyword. Like classes, they can only be declared at the topmost scope, and `export interface X = ...` is permitted.
+Like classes, defining two classes with the same name in the same module is not allowed.
+
+The right hand side of an interface declaration is almost syntactically identical to a table type, with the exception that there are no read/write access modifiers. Its contents are the fields and methods that the interface declares.
+
+To declare that a class implements a particular interface, the class name is suffixed with `: <Interface name>` (see above for a concrete example of this). Interfaces can inherit from other interfaces as well via the same syntax. We do not consider multiple inheritance in this RFC, so attempting to define a class implementing multiple interfaces is a syntax error. Additionally, we consider only interface inheritance in this RFC, so `class Cls1 : Cls2` is also a syntax error.
```
alexmccord, line 44: > Additionally, we consider only interface inheritance in this RFC, so `class Cls1 : Cls2` is also a syntax error.

A few issues with this one.

Firstly, since you also define `class.isinstance(x, A)` for some object `x` and an interface `A`, this turned the feature from something that existed only in the metalanguage to something that be realized in the object language.

Consider:

```luau
const M = require("./mod")

class Foo: M.A
end
```

This means it is no longer enough to make this a syntax error and requires the runtime to throw an error during the instantiation of `Foo`. Unless you restrict the syntax so heavily that you can't even have `class Foo: A` where `A` is an alias to some interface, e.g.:

```luau
const M = require("./mod")
const A = M.A

class Foo: A -- still a syntax error because `A` is not `interface`, it's just an ordinary lvalue.
end
```

Secondly, if you do this restriction in the parser, this would mean you can't set up an abstraction layer where you hide your classes and export just the interface, _and_ implement those classes in different files. Not to mention that if the library exports an interface, I expect to be able to implement my class against their interface.

```
src/
  foo/
    init.luau
    interface.luau
    class1.luau
    class2.luau

--> ./src/foo/init.luau
return require("./interface")

--> ./src/foo/interface.luau
export interface A = { ... }

--> ./src/foo/class1.luau
const Interface = require("./interface")

class Class1: Interface.A
  -- ...
end

--> ./src/foo/class2.luau
const Interface = require("./interface")

class Class2: Interface.A
  -- ...
end
```

TenebrisNoctua:
What this RFC proposes doesn't really make sense to me. Why introduce a new type and syntax to the language when we can just extend the existing class syntax instead to allow for implementation inheritance?

Looking at the language ecosystem and community, implementation inheritance feels the most right to me. It would be the simplest idea of inheritance for people to adapt while also not extending the VM to include a new type again.

Also, interfaces seem to be capable of "inheriting" one another, and it feels like a weird workaround to have interfaces that can inherit, but not classes.

Also, has the idea of 'abstract classes' considered? They could provide equivalent functionality as interfaces, yet we wouldn't be creating a new syntax.

Lastly, the proposed `self` type idea also adds to the complexity element of this proposed syntax. This syntax needs it for the reasons described in this RFC, but I wonder if we would need this inside implementation inheritance.

TenebrisNoctua:
The RFC also states that it does not consider multiple interface inheritance, which is personally *the only reason* why I would use interfaces. 

Implementation inheritance poses the diamond problem, if allowed to have multi-inheritance. If this RFC were to allow multiple interface inheritance, this may have posed as a strong argument against implementation inheritance. And yet, this hasn't been made possible in this RFC.

So what's the point? Do we want to take the long and weird way around just to maybe have multiple interface inheritance later? Or will we not have them at all?

docs/classes-interface-inheritance.md
alexmccord, line 44: > We do not consider multiple inheritance in this RFC, so attempting to define a class implementing multiple interfaces is a syntax error.

This restriction probably shouldn't be here, to be honest. Changing the VM _and_ the type system _and_ the compiler _and_ the codegen is laborious and painful, compared to just having it from the get-go. Classes are going to need to implement more than just one interface, period. `interface Stream: Reader & Writer` is the canonical example.

Bottersnike:
I'm not a huge fan of the `:` syntax for inheritance. It feels like it closes the door entirely on ever having implementation inheritance (which imo is important if classes are to truly replace metatable-based code).

```luau
class Cat : Animal extends FooBar
```

reads really poorly, and

```luau
class Cat : Animal, FooBar
```

gets confusing if `Animal` is an interface but `FooBar` is a parent implementation.

The `:` syntax could work if an interface is not anything special, and instead is just a normal class (with potentially something like `abstract function (...)` syntax used to mark members as abstract, and therefore the whole class as not constructable), but that's not the design outlined in this RFC right now.

Rob07erto:
This feels to me like it doesn't address what we need to:
Current luau code allows for similar type of behavior already but for some funky syntax:

--!strict
type Person = {
	name: string,
	age: number,
}

local person = {
	name = "John Doe",
	age = 0,
	greet = function(self: Person)
		print("Hello, my name is " .. self.name)
	end,
} 
person = person :: typeof(person) & Person


function name(person: Person)
	print(person.name)
end

person:greet(person)
name(person)

What I think we should be doing first is simplify this line
person = person :: typeof(person) & Person

My proposal is:
type Person = {
	name: string,
	age: number,
}

// Saying person -> Person is constrained to the type Person.
local person -> Person  = {
	name = "John Doe",
	age = 0,
	greet = function(self: Person)
		print("Hello, my name is " .. self.name)
	end,
} 

function name(person: Person)
	print(person.name)
end

person:greet(person)
name(person)

This keeps things simpler and allows for the same behavior.

Rob07erto:
Second of all I also believe for classes Id rather than be like an ECMA Script which is a syntaxic sugar for prop typical inheritance (table prototyping through metatables) I believe this would better in the end for better flexibility and intractability. 

I feel in order for classes to work, we need to define encapsulated types  so we get encapsulation at a comp time level (types that define public, private, and if we want too protected), shared self type (duh), and the ability to say we implement a type first. Then we produce a base proto object which automically overloads the __call method, and eq method so we can do let myClass: MyClass = MyClass() and if myclass1 == myclass2. Then we develope the class syntaxic sugar that simplified the creation of these.

skberkeley:
Hi folks, thank you all for your feedback and comments, and apologies for any concerns this RFC may have caused! I was intending to collect feedback internally on how we want inheritance to work, and this was the wrong medium for that. We'll open another RFC once we have a design that we have more confidence in!

PR #201, opened 2026-05-08
RFC: Dedicated INC/DEC VM opcodes for optimized local mutation

YarikSuperpro:
([Rendered](https://github.com/YarikSuperpro/rfcs/blob/inc-dec-opcodes/increment-decrement-opcodes.md))
Introduces an RFC proposing dedicated `INC` and `DEC` VM opcodes for local numeric increment/decrement patterns such as `x = x + 1`.

The proposal focuses on reducing VM overhead, improving hot loop execution efficiency, and allowing the VM/compiler to represent mutation intent directly instead of relying on generic arithmetic instructions like `ADDK`/`SUBK`.

No new language syntax or semantics are introduced.

### DIFF
--- a/increment-decrement-opcodes.md
+++ b/increment-decrement-opcodes.md
@@ -96,8 +96,8 @@ x = x - 1
 ```
 may lower into:
 ```luau
-INC Rx
-DEC Rx
+INC Rn
+DEC Rn
 ```
 instead of:
 ```luau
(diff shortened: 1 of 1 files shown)


aatxe:
We don't take RFCs for VM opcodes. These are implementation details, not part of the actual Luau language.

PR #202, opened 2026-05-08
Add 24-bit Integer Buffer Accessors (u24 / i24)

YarikSuperpro:
([Rendered](https://github.com/YarikSuperpro/rfcs/blob/int-24-buffer/docs/int-24-buffers.md))
This proposal adds `u24` and `i24` read/write support to the `buffer` library, enabling native 24-bit integer serialization for compact data representations such as RGB colors, small IDs, and packed binary formats.

The goal is to reduce the need for manual bit packing or multi-byte composition in common serialization-heavy code paths, while improving readability and reducing boilerplate in performance-sensitive systems.

No changes to language syntax or VM behavior are introduced; this is a pure extension of the buffer API.


PhoenixWhitefire (commented):
Your RFC is supposed to be in the `docs/` directory, please move it there.

vegorov-rbx:
Duplicate of #32

Also note that for custom widths, we already introduced readbits and writebits.

PR #203, opened 2026-05-09
Integral number singletons

PhoenixWhitefire:
[Rendered.](https://github.com/PhoenixWhitefire/rfcs/blob/integral-number-singletons/docs/number-singletons-heterogeneous-arrays.md)

### DIFF
--- a/docs/integral-number-singletons.md
+++ b/docs/integral-number-singletons.md
@@ -49,6 +49,8 @@ as the union of `Items` is not narrowed from `{ string } | { integer }`.
 
 Certain runtimes may expose platform-level APIs that use enumerations that are represented as `number`s.
 
+Programmers may want to annotate that a table only has a certain limited set of valid number indices.
+
 ## Design
 
 Integral numeric literals valid in the value language (`0`, `0xF2`, `0b11001`, etc) will be made valid for use in the type language, similar to literals
--- a/docs/integral-number-singletons.md
+++ b/docs/integral-number-singletons.md
@@ -1,0 +1,89 @@
+
+# Integral `number` singletons
+
+## Summary
+Allow for singletons of the `number` primitive, similar to `boolean` (`true`/`false`).
+
+## Motivation
+Programmers may want to use numeric constants as enumerations. Consider:
+
+```luau
+type Deserialized = {
+    Items: { integer }
+}
+
+local function convertFromV1ToV2(_: { string }): { integer }
+    -- ...
+end
+
+local function deserialize(Serialized): Deserialized
+    local deserialized = {}
+
+    if Serialized.Version == 1 then
+        deserialized.Items = convertFromV1ToV2(Serialized.Items)
+    elseif Serialized.Version == 2 then
+        deserialized.Items = Serialized.Items
+    else
+        error("unexpected version")
+    end
+
+    return deserialized
+end
+```
+
+As it stands, this code cannot be easily typechecked as the type of `Serialized` would have to be narrowed based on the numeric singleton `.Version`.
+The desired type of `Serialized` would be:
+
+```luau
+type SerializedStructure = {
+    Version: 1,
+    Items: { string }
+} | {
+    Version: 2,
+    Items: { integer }
+}
+```
+
+, however, this is currently inexpressible. `Version` would have to be left annotated as `number`, which prevents `deserialize` from typechecking
+as the union of `Items` is not narrowed from `{ string } | { integer }`.
+
+Certain runtimes may expose platform-level APIs that use enumerations that are represented as `number`s.
+
+## Design
+
+Integral numeric literals valid in the value language (`0`, `0xF2`, `0b11001`, etc) will be made valid for use in the type language, similar to literals
+such as `nil` or string literals like `"hello"`. They will be inferred as either their singleton or their supertype (`number`) with the same rules as
+string singletons.
+Literals with a decimal component are disallowed and raise a syntax error.
+
+Unions of the singletons will be narrowable through refinement similar to strings:
+
+```luau
+local x: 1 | 0b10 | 0x3 = getX()
+
+if x == 1 or x == 2 then
+    local y: 1 | 2 = x -- No type error
+    local z: 3     = x -- Type error
+end
+```
+
+They will be valid in the table indexer position:
+
+```luau
+type T = {
+    [1 | 2]: string
+}
+
+local t: T = {...}
+print(t[1]) -- Ok, 1 <: 1 | 2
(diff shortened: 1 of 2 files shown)


docs/integral-number-singletons.md
```diff
@@ -0,0 +1,91 @@
...
+type T = {
+    [1 | 2]: string
+}
+
+local t: T = {...}
+print(t[1]) -- Ok, 1 <: 1 | 2
+print(t[3]) -- Type error, 3 </: 1 | 2
+```
+
+Heterogeneous arrays are out of the scope of this RFC.
```
alexmccord: *ahem*
```luau
type T = { [1]: A } & { [2]: B }
```

PhoenixWhitefire: Oh dear

PhoenixWhitefire: I'm not sure what you want here, tables can only have one indexer, so that's `[1 & 2]: A & B` if I'm not mistaken, aka `[never]: never`?

alexmccord: Tables can have multiple indexers, and multiple indexers can be merged into one indexer whose result type is a dependent type. This is just `{ p: A } & { q: B }`, or `{ [string]: number } & { [number]: string }`. These are well-formed types that actually are meaningful. The one I wrote is also meaningful. Properties are just an indexer whose result type is a dependent type, whose type is dependent on what key was used to access it, so `{ [1]: A } & { [2]: B }` is equivalent to `{A, B}` as a heterogeneous array type.

PhoenixWhitefire: That's interesting, but when I try to intersect `{ [string]: number } & { [number]: string }` I get `never` as the indexer ([Playground](https://play.luau.org/#N4IgbiBcBMA0IDMqgFYEsDuB7AdgLwDoAbAVwEMSoQBaagAgBcALNAZzoCcBTARxLW7sydAA7cGDAJ50yAWwBGaBmiwl2UkV1YMyDLnVaTtXWQB0c5olgDGZIjMh0cJBVw4B+c5Zt26CEjjWyrh0LAAUAJTmdHRoCDKMTFwWODExbhxYHGGmIGh0ACa4DDJEVhhOWnoFdLZlaDgA5rlRqXTJBdF0XcIAvHTQXV0IWbF0-QCMsHQArAAMhVhdMVZ1teMJANR0AMx0tAlsTmhEXR1DbX3HpzjnKeERIPBkHGhcRMTklJA09MxHRS0TiwJW4fAE+g0VV0+l0pRu3jWrBMZBETCy+n6CDsyK8OFWvn8gWCqQeXTiBhRaIxiWSy3aHEy2Vy+SKOBKdnKlWMNTqRAazRArXSOE6bS6yNkqPR3A2DA4JC4eJWPnsV0GbRGHDGk2m80W9IJ9msG2E2x2Z1FyspUupsqxOKVtyt9zQkSeIClDU+5E+FCoULoABUNsAugBtZyuDgAXUc2leTXMAF86AAyOhhtrhhMCuN0eRYLBELhkFLJvFGugAD0cQfMBwABoXi6XUhnc01G4jfNJ+tXwxMYw36M2iyWy938aq6HgNgPckkyksQMOcE3bmA3FOqxhHABrLiSLAIAA8QYAfHjAwBVUMRzuNfNR+RuFN4puPnczsgHo8n08byvHAQGTZ4qC9HAfTIP1KHgVhkBALARGUWQ0DwXQVBwAAZLgtyIKA4BAawsFkEQTjcAAlFEOH3BDIHlRVk2TIA))

```luau
type T = {
  [number]: string
} & {
  [string]: boolean
}

local x: T
-- `boolean & string`
local y = x[1]
-- `boolean`
local z = x["hello"]
-- `never`
local w: keyof<T>

type U = {
  [string]: number
}

-- `string`
local a: keyof<U>
```

I'm not sure what's going on with `z` but `w` would seem to indicate that it's intersecting the indexers, like I would expect.

My intent is that `{ [1]: T }` is producing a table type with an *indexer* of a singleton type, not a property. From what I can tell, tables may only have one indexer (attempting to write multiple directly is a syntax error). The indexer `[1 | 2]: T` is supposed to be a more restricted `[number]: T`. There's no dependent type because there is no way to have `{ [1]: T, [2]: U }` with this RFC, I thought it would be too much.

alexmccord: Yeah, that current behavior is just silly.

Bottersnike: I think having `{[1]: A} & {[2]: B}` result in anything _other_ than `{[1]: A, [2]: B}` would be pretty confusing, as not matching `{["foo"]: A} & {["bar"]: B}` semantics would be weird.

Honestly, why _not_ consider them fields, and get heterogeneous arrays for free?

PhoenixWhitefire: I was thinking about this, it did feel like 2 separate ideas but yeah, I think I'll add it to the RFC.

PhoenixWhitefire: What kind of annoys me is this whole thing with indexers and properties that tables have, it feels sort of like something that should be the same thing has been split in two?

Anyway, I've gone ahead and replaced the indexer part with "they'll be like properties" to support heterogeneous arrays.

alexmccord: I don't think this RFC needs heterogeneous array types, that can be split into another one. I was gesturing that once you have intersection types _and_ integral number singletons, you already have heterogeneous array types (with convoluted syntax)

PhoenixWhitefire: It can only work if it's treated like a property and not an indexer because the intersection is different, so if this is going to define that `{ [1]: T }` is a property and not an indexer then that's just heterogeneous arrays already

docs/number-singletons-heterogeneous-arrays.md
```diff
@@ -0,0 +1,113 @@
...
+
+type V = T & U --> { [1]: boolean, [2]: buffer }
+```
+
+## Drawbacks
+* Increases complexity of the type system
+
+## Alternatives
+* Allow for decimals
+* Singletons for `integer`s
```
gaymeowing, line 113: Why can't integer support be included within this RFC as well? As it'd just be allowing the `i` suffix at the end of the singleton to indicate its an integer. With roughly the same syntax as `number` singletons as proposed.

PhoenixWhitefire: `number` is generally much more widely used (default indexer for array-like tables as well), `integer` singletons could be a future RFC.

jLn0n:
typescript has this, why not luau too

PR #191, opened 2026-04-06
Classes!!

andyfriesen:
[Rendered](https://github.com/luau-lang/rfcs/blob/master/docs/syntax-classes.md).

PR #204, opened 2026-05-11
Scout RFC: `is` keyword.

alexmccord:
[Rendered](https://github.com/alexmccord/rfcs/blob/kw-is-cool/docs/syntax-is-keyword.md).

### DIFF
--- a/docs/syntax-is-keyword.md
+++ b/docs/syntax-is-keyword.md
@@ -281,11 +281,16 @@ type Pred = (L: lua_State, x: unknown, polarity: boolean) -> boolean
 ```
 
 The `polarity: boolean` parameter is purely for the host to have an opportunity
-to coax the C/C++ compiler to generate a short-circuiting logic in the case that
-the host has authored a predicate as a series of logical conjunctions for when
-`polarity == false`. This `polarity` is load-bearing, and requires the lawful
-`pred(L, x, true) == not pred(L, x, false)`, which is trivial by writing
-`polarity == (x and y)`.
+to coax the C/C++ compiler to generate a short-circuiting logic for when
+`polarity == false` in the case that the host has authored a predicate as a
+series of logical conjunctions. This `polarity` is load-bearing, and the host is
+required to write a lawful predicate function satisfying:
+
+```
+pred(L, x, true) == not pred(L, x, false)
+```
+
+This is trivial by writing `polarity == (x and y)`.
 
 The typename registry resolver has the type:
 
@@ -477,3 +482,8 @@ TODO.
    module to fail to initialize, as opposed to matching the behavior of `x is
    SomeLocal` where `SomeLocal` is `nil` or some non-class which throws an error
    only when the expression `x is SomeLocal` is being executed.
+
+4. Instead of giving the `polarity` to the predicate functions, the host always
+   write a predicate that assumes `polarity == true` and the VM negates the
+   result on their behalf. This removes the `polarity` parameter from the
+   calling convention.
--- a/docs/syntax-is-keyword.md
+++ b/docs/syntax-is-keyword.md
@@ -458,17 +458,19 @@ The predicate function has the type:
 type Pred = (L: lua_State, x: unknown, polarity: boolean) -> boolean
 ```
 
+All predicate functions must satisfy the following law:
+
+- De Morgan: `pred(L, x, true) == not pred(L, x, false)`
+
+This is trivially discharged by `polarity == property` for some `property`,
+including logical conjunctions `propA and propB and propC`.
+
 The `polarity: boolean` parameter is purely for the host to have an opportunity
 to coax the C/C++ compiler to generate a short-circuiting logic for when
 `polarity == false` in the case that the host has authored a predicate as a
-series of logical conjunctions. This `polarity` is load-bearing, and the host is
-required to write a lawful predicate function satisfying:
-
-```
-pred(L, x, true) == not pred(L, x, false)
-```
-
-This is trivial by writing `polarity == (x and y)`.
+series of logical conjunctions, e.g. `not (a and b and c)` is equivalent to the
+condition `not a or not b or not c`, which short-circuits on the first satisfied
+disjunct.
 
 The typename registry resolver has the type:
 
--- a/docs/syntax-is-keyword.md
+++ b/docs/syntax-is-keyword.md
@@ -152,8 +152,8 @@ disambiguation mechanism.
 
 ### Value namespace vs typename namespace
 
-One of the suggestion in the classes RFC was to allow `class.isinstance` to work
-with certain primitives that have a built-in global library, e.g.
+One suggestion in the `class` RFC was to allow `class.isinstance` to work with
+certain primitives that have a built-in global library, e.g.
 
 ```luau
 function is_string(x: unknown)
@@ -551,15 +551,27 @@ type guards anyhow, e.g. `typeof(x) == "nill"` will silently do nothing and
 always returns `false` (unless by chance its `__type` is `nill`...). Dynamically
 typed programming languages are already full of this class of bugs, e.g. field
 projections, mistyped locals resolves to a global, etc. The type system can be
-used to rescue users from typos, but the status quo remain no worse than before.
+used to rescue users from typos, but the status quo remains no worse than
+before.
 
-This also requires the host to populate the `typename` registry to participate
-in the `is` keyword with all possible types from their environment. A solution
-that could alleviate this pain is to provide a hook for when the `typename` is
-not found in the registry, so that populating the registry can be done on-demand
-and keep the startup time and memory cost as small as possible. Nevertheless,
-this is one more thing that the host now has to do _if_ they want to cooperate
-with the `is` keyword.
+This also requires the host to populate the `typename` registry so types from
+their environment can participate in the `is` keyword with all possible types
+from their environment. A solution that could alleviate this pain is to provide
+a hook for when the `typename` is not found in the registry, so that populating
+the registry can be done on-demand and keep the startup time and memory cost as
+small as possible. Nevertheless, this is one more thing that the host now has to
+do _if_ they want to cooperate with the `is` keyword.
+
+We also can't integrate host-owned `userdata` with `__type` to cooperate with
+the `is` keyword by default, since you might have nontrivial predicates e.g.
+`part is BasePart`. If `part` has some `__type = "Part"`, then this predicate
+immediately fails. Ditto that `__type` does not necessarily need to be a fully
+qualified typename, e.g. `Enum.LuauTypeCheckMode.Strict` does not contain the
+qualified prefix path `Enum`. It's also possible that certain typenames are
+inherently structural beyond the `userdata` itself, e.g. `Character` might be a
+`Model` that contains a child instance named `Head`, some `Humanoid`, etc. So
+the current RFC design is generalized to support that at the cost of
+boilerplate.
 
 ## Alternatives
 
@@ -586,3 +598,7 @@ with the `is` keyword.
    write a predicate that assumes `polarity == true` and the VM negates the
    result on their behalf. This removes the `polarity` parameter from the
    calling convention.
+
+5. Instead of a registry, `userdata` could have `__is` metamethod for `userdata`
+   to participate in (locked in the same way `__namecall` and `__type` is), but
+   that loses out on various optimization opportunities, since `__is` is opaque.
--- a/docs/syntax-is-keyword.md
+++ b/docs/syntax-is-keyword.md
@@ -140,16 +140,15 @@ the typename registry, under the following constraints:
 
 1. It must not overwrite any built-in `typename`s.
 2. It cannot overwrite any `typename`s that have already been registered.
-3. It cannot register any new `typename`s once execution starts.
-4. No registered `typename`s can be retracted from the registry.
-5. The `typename` registry lives in the `global_State`, so no module-specific
+3. No registered `typename`s can be retracted from the registry.
+4. The `typename` registry lives in the `global_State`, so no module-specific
    host-defined `typename`s exist.
 
 The expectation is that `typename`s are globally stable and consistent, and no
-new `typename`s can show up or be invalidated at any arbitrary point in time. If
-the host environment has two different types of the same name, that's a design
-issue and the responsibility does not rest with us. Qualified paths are
-available as a disambiguation mechanism.
+`typename`s can be invalidated at any arbitrary point in time. If the host
+environment has two different types of the same name, that's a design issue and
+the responsibility does not rest with us. Qualified paths are available as a
+disambiguation mechanism.
 
 ### Value namespace vs typename namespace
 
--- a/docs/syntax-is-keyword.md
+++ b/docs/syntax-is-keyword.md
@@ -21,8 +21,8 @@
     - [Partial evaluation](#partial-evaluation)
     - [Prefix-sharing](#prefix-sharing)
     - [Decision tree](#decision-tree)
-  - [Drawbacks](#drawbacks)
-  - [Alternatives](#alternatives)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


Ukendio:
I personally want a more powerful feature to introspect types on runtime than this, it seems boring to have yet another way of doing something that can already be done ostensibly + integrating with the class feature. I know the luau team is conflicted with typeof but I really liked it and wish was there was a function that gave more expansive information that is readable on runtime.

What I had in mind was something like this and would aim to solve the same problem while integrating with the existing syntax and has similar ergonomics to user defined type functions
```lua
local ti = typeinfo(x)
if ti.type == "class" and t1.name == "Part" then 
  
end
if ti.type == "table" then 
  for k, v in ti.members do 

  end
end
if ti.type == "function" then 
  for k, v in ti.generics then 

  end
  for k, v in ti.parameters then 
  
  end
end
```

TenebrisNoctua:
I feel like this RFC could use additional examples in the motivation section. The rest of the RFC is fine, but it could benefit from additional reasons as to why people would rather have `is` than `typeof(...) and :isA()`.

MagmaBurnsV:
I don't know how I feel about adding new syntax with as many surprising restrictions/gotchas as this. I feel even more uncomfortable with the proposed evaluation semantics.

First, `is` poses the exact same syntax complications that the `as` keyword did 6 years ago. Disregarding the increased compiler complexity from the arbitrary lookahead required to make this work, the subtle differences with `x is T` and `x is (T)` destroys intuition with how expressions are treated in the language, because the rule "you can put explicit parentheses around any associativity and it's the same thing" breaks. Other oddities like `x is not T` and `x is T :: U` make the rules even harder to learn because they aren't really consistent.

Second, the typename resolution rules are unhygienic. Making an exception to the "identifiers fallback to the global namespace if not local" rule is bad enough, but I also want to mention that this prevents users from naming their locals *anything* that collides with a typename, lest they permanently break the expression:

```luau
local boolean = true
local result = x is boolean -- impossible to do now, not even forward-declaring an alias can fix this
```

All of these ugly warts combined convinces me that this is not worth the added syntax sugar.

docs/syntax-is-keyword.md
```diff
@@ -0,0 +1,602 @@
...
+is `(not x) is boolean`, ditto `-v is Vector3` is `(-v) is Vector3` and so on.
+
+### Built-in `typename`s
+
+In a barebone environment, the default set of typenames are all the following
+built-in Luau VM primitives:
+
+- `boolean`,
+- `buffer`,
+- `function`,
```
MagmaBurnsV, line 258: I don't think it would be possible to have `x is function` since a naked `function` looks like an incomplete function block. Again, theoretically solvable with lots of insane lookahead, but is it really worth it?

alexmccord: It sure is possible. `local b = x is function` is not a valid parse. Like I said in the grammar, if the `is` keyword is followed by 1. `nil`, 2. `function`, 3. `not`, or 4. `NAME`, then it's an `is` expression. No existing syntax occupies this. It can be parsed in LL(1).

alexmccord:
> First, `is` poses the exact same syntax complications that the `as` keyword did 6 years ago.

It doesn't! This RFC explains this at [this section](https://github.com/luau-lang/rfcs/pull/204/files#diff-d1113c5a3dd6052f7604e7b628207a56bda58e82c8df864c7ca3f5f96820ad16R416-R419), in particular.

> Other oddities like `x is not T` and `x is T :: U` make the rules even harder to learn because they aren't really consistent.

This RFC covers this. I already said that these two cannot be used together, and you have to explicitly write parentheses around the expression you want to cast or perform a type test, which is what happens if `is` and `::` has the same precedence, and both are neither left nor right associative. So parentheses are required. I'll rephrase this in the RFC to a little bit clearer.

> Second, the typename resolution rules are unhygienic. [sic]
>
> ```luau
> local boolean = true
> local result = x is boolean -- impossible to do now, not even forward-declaring an alias can fix this
> ```

This is true, but...

1. Why would you store a value in a local named `boolean`? It is absolutely not descriptive enough.
2. Renaming is always available.
3. The type system will warn you that `boolean` is not a subtype of `class`, even in nonstrict mode.
4. You already can shadow globals quite often. Does that make globals unhygienic? Yes. Yet when you want to use the globals, you end up renaming your locals, as opposed to aliasing the globals. This tells you a lot.

I'll add this into the RFC as a wart, but counter it with it not being all that different from globals. Global typenames are no different from global libraries from this point of view.

docs/syntax-is-keyword.md
MagmaBurnsV: That seems equally as bad since `function() end` is no longer an unconditional atom in the grammar. Yes, the operation `x is function() end` is already "incorrect," but that doesn't justify changing the rules. Anyone who would look at that expression for the first time would think the right operand is a function value, not that the parser clobbers everything.

My point is that this isn't just one grammar exception in isolation, but one in five exceptions that make reasoning with the syntax difficult because it's not following the established rules of the language. This is really problematic syntax for a low-value feature.

MagmaBurnsV:
Would a less magical global `is(x, kind)` function be an alternative the team would be willing to take? It would work basically the same way, where if you pass in a class/interface value to it, it will do an `isinstance` check, and if you pass in a string, it routes to a predicate registry defined by the host. Since it's an intrinsic function, type refinement and optimizations are still possible.

This is pretty much exactly what the RFC is proposing, but without the awkward syntax and namespacing rules. I find it surprising that this option isn't even mentioned in the alternatives.

docs/syntax-is-keyword.md
```diff
@@ -0,0 +1,599 @@
...
+to coax the C/C++ compiler to generate a short-circuiting logic for when
+`polarity == false` in the case that the host has authored a predicate as a
+series of logical conjunctions. This `polarity` is load-bearing, and the host is
+required to write a lawful predicate function satisfying:
+
+```
+pred(L, x, true) == not pred(L, x, false)
+```
+
+This is trivial by writing `polarity == (x and y)`.
```
Cooldude2606: Just a small point of clarity. It was not immediately obvious to me that this is a representation of the host writing a predicate. For example it was not clear what `y` was and I had assumed `x` was the argument provided in the previous line. I would suggest being a little more verbose: "this is trivial to implement for a host by writing predicates in the form: `polarity == (propA and propB)`

alexmccord: Will rephrase.

docs/syntax-is-keyword.md
```diff
@@ -0,0 +1,713 @@
...
+
+6. Instead of the `is` keyword, add a new built-in function `is(x, t)` where
+   `is` has the signature `(x: unknown, t: string | class) -> boolean`. The
+   tradeoff that's implicit in this is:
+
+   - occupies a fastcall slot
+   - requires pattern matching on `is(x, t)` for type refinements to trigger
+   - requires pessimistic codegen if the global scope was monkeypatched around
+   - optimization potentials are lost, e.g. partial evaluation, prefix sharing,
+     and decision tree is no longer possible.
```
MagmaBurnsV, line 683: How so? When supplied a string literal in a pure environment like `is(x, "Part")`, that would give the same amount of information at compile time that the keyword form would. The only difference is it must fallback to a runtime lookup in unidiomatic cases, which is perfectly fine.

Reading the section on optimizations, it seems like a major stake the team has in this RFC is in purely speculative optimizations, instead of the advertised motivation of unifying runtime type-checking conventions. This is a common pattern I'm seeing with controversial RFCs, where features meant for ergonomics and utility are retrofitted for optimization-first at the cost of usability. This does not a good feature make.

Things like generating decision trees is cool, but like you've said, that's an NP-hard algorithm. Not even optimizations in GCC and LLVM usually have that kind of complexity, with the closest I can think of being register allocation. It is unrealistic that this kind of optimization will ever be implemented, so it's shallow to even mention it, let alone reject an alternative over it.

alexmccord: > How so? When supplied a string literal in a pure environment like is(x, "Part"), that would give the same amount of information at compile time that the keyword form would. The only difference is it must fallback to a runtime lookup in unidiomatic cases, which is perfectly fine.

The "pure environment" is doing a lot of heavy lifting here. You and I are in full agreement on this, in principle these two are the same thing. If the language didn't have `getfenv`/`setfenv`/et al, this problem wouldn't exist and `is(x, "Part")` can be partially evaluated in the same exact way as in this RFC. But that's _conditional_ on the "pure environment" which you postulated here.

> it seems like a major stake the team has in this RFC

I am not on the team anymore. This is all in my free time.

> Reading the section on optimizations, it seems like a major stake [...] is in purely speculative optimizations, instead of the advertised motivation of unifying runtime type-checking conventions.

You made up a story to fit the narrative in your head with no basis of what actually entailed in reality. Let me give you the receipts so you can reconstruct the `is` keyword timeline yourself.

- I saw [this comment](https://github.com/luau-lang/rfcs/pull/191#discussion_r3041570882). My first instinct was to construct a counterexample to see if it falls apart, which it did. This means `instanceof` function wasn't quite generalized enough to work the way they wanted it to.
- I also saw a bunch of other comments on Roblox OSS discord that wanted some expression of the form `value is Class` or `value instanceof MyClass`.
  1. https://discord.com/channels/385151591524597761/906369439262461992/1490861974475768139
  2. https://discord.com/channels/385151591524597761/906369439262461992/1491141109940748308
  3. https://discord.com/channels/385151591524597761/906369439262461992/1495947965104263278
  4. https://discord.com/channels/385151591524597761/906369439262461992/1493600457409695834
- The argument that @deviaze made where `is` keyword would feel more first-class was compelling enough. I agree with this.
  - Several people have said they wanted to be able to write `x is number`.
  - I also noted the [use cases with Lute's API](https://discord.com/channels/385151591524597761/906369439262461992/1495946055668334623) e.g. `x is std.fs.File`.
- I made a realization that [`is` keyword is backward compatible](https://github.com/luau-lang/rfcs/pull/191#discussion_r3068644898).

  Since `is` keyword was now realistically possible, this was enough for me to start thinking about what the design would look like that satisfied the use cases with primitives, host-defined types, _and_ classes.

  This RFC contains the design.
- While thinking about this and writing this RFC, I also _noticed_ the optimization ideas that `is` unlocks by turning it into an expression.
  - I typed the first one in. Immediately noticed an optimization after typing that first one.
  - So I typed the second one in. Immediately noticed a possible optimization in the case that reordering is allowed.
  - So I typed that in. Noticed that finding an optimal reordering is the same as a decision tree.
  - That's why I said "[as a _side benefit_](https://github.com/luau-lang/rfcs/pull/204/files#diff-d1113c5a3dd6052f7604e7b628207a56bda58e82c8df864c7ca3f5f96820ad16R43-R45)."

Regarding motivations, there can be more than one motivation. If you go back to the motivation section, you can see "and as a side benefit, can be partially evaluated at compile time." These optimization _ideas_ (partial evaluation, prefix-sharing, and decision tree) that I was referring to are "by compile time."

Let's get into the partial evaluation.

1. Partial evaluation of `is(x, "Part")` is not possible because Luau's environment is impure.

   Therefore, to do this, you would need N different specialization of `is(x, t)` for some constant string `t`, that would take up N fastcall slots.

   That would fill those slots fast, and doesn't work with host-defined types.
2. The VM now has to dispatch through a hashmap from some interned string to a C function.
3. On cache-miss, parse the string to split on `.` for qualified typenames and resolve it, then cache both keys to the same identity and predicate.
4. If the predicate is expensive, negation of such tests are slower.

   One use case is `x is Character`.
   - Check if `x` is a subtype of `Model`,
   - Check if `x` has a child named `Humanoid` of type `Humanoid`,
   - Check if `x` has a child named `HumanoidRootPart` that is a subtype of `BasePart`,
   - Check if `x`'s `PrimaryPart` is assigned to the same `HumanoidRootPart`.
   - Negation of any of the conjunct should just... short-circuit and return `false`.

   That's just one instance of a nontrivial predicate function for one host to do. Another host in a completely different environment might want this escape hatch.

Now, say we postulate that partial evaluation was solved. What about prefix sharing? If any arbitrary computation could happen between two calls of `is`, you can't fuse two conditions at compile time at all.

Decision tree presupposes prefix sharing. If you can't do prefix sharing, you can't make a decision tree. Compile-time prefix sharing optimization presupposes a pure environment where no computation invalidates the global `is`.

And this is not even taking inlining into account. If you have a function that case-splits on the same value by type, and your function is also case-splitting on the same value by type, then you have a smaller cost model than the fastcall variant. This specific optimization is called case-of-case elimination, and that can be inlined even if the function itself as a whole is too expensive to be inlined.

> This is a common pattern I'm seeing with controversial RFCs, where features meant for ergonomics and utility are retrofitted for optimization-first at the cost of usability. This does not a good feature make.

This is purely rhetoric and completely unnecessary. The optimization-first rhetoric is false, as per the timeline I have given above.

> Things like generating decision trees is cool, but like you've said, that's an NP-hard algorithm.

Which is why I linked to Maranget's paper on the heuristics that generates a close-to-optimal decision tree. GHC, Rust, OCaml, and even C# (unsurprisingly) all use Maranget's heuristics _or some flavor_ of it.

Check this out: http://jeremykun.com/2017/12/29/np-hard-does-not-mean-hard/

> Not even optimizations in GCC and LLVM usually have that kind of complexity, with the closest I can think of being register allocation.

This is categorically false. GCC and LLVM are on the orders of 15 million lines or more, and they do run into NP-hard problems. NP-hard problems are literally par for the course in compilers, because (as per the blogpost) expressions have _shared_ subexpressions and so you end up with a DAG, and you have logic that follows the expression shape. That _immediately_ puts your problem into NP-hard. GCC, LLVM, GHC, Rust, OCaml, and so many others, all have many optimizations in this complexity class.

1. As you noted, register allocation is NP-hard iff you have register spilling, which turns this into a graph coloring problem. Otherwise interval graphs are enough, and this is in P. GCC and LLVM do register spilling, so their register allocation is NP-hard.
2. Instruction selection. This is NP-hard if you have subexpressions that shares the same registers.
3. Instruction scheduling.
4. Optimal code layout. For basic blocks, you generally want them to be close by to maximize I-cache locality. The same is also true for functions.
5. Optimal inlining decisions. In its most aggressive form, this is just a [knapsack problem](https://en.wikipedia.org/wiki/Knapsack_problem). This makes sense, since inlining decisions can go in both directions.

There are many more, but the point is made.

If Luau's type system could be used to guide codegen, then our codegen is in effect also NP-hard because the type system is also NP-hard (normalization, constraint solving, etc).

> It is unrealistic that this kind of optimization will ever be implemented, so it's shallow to even mention it, let alone reject an alternative over it.

Again, this is purely rhetoric.

"It is unrealistic that this kind of optimization will ever be implemented" is a predictive claim wrapped under the assertive language guise. The engineers who work on this language are the ones who can tell me that this is unrealistic.

As for "it's shallow to even mention it," this was uncouth and completely unnecessary. Just because you are frustrated with RFCs does not mean you can take it out on me. Read up on [adversarial collaboration](https://en.wikipedia.org/wiki/Adversarial_collaboration) for the next time. It argues substance over rhetoric.

Regarding "reject," I didn't use that word. It is merely a "tradeoff" that this alternative makes. Not everything in alternatives are rejections, they're just different design ideas and each alternative have their own conclusions.

Additionally, in your previous comment, you said you found it surprising that this wasn't in the alternatives. The second I add it as an alternative, it is somehow offensive.

MagmaBurnsV: I am sorry, I was not aware that you were no longer on the team. I was under the impression that this was something set course to being implemented, which was why I felt the need to apply special scrutiny in my criticisms. Clearly, I read the room wrong 😅.

For clarification, my cynical interpretation of the team's motivations largely stemmed from the way integers and exports/const were handled (e.g. integers only being useful in a handful of high-performance cases, and exports being const for optimizations which got walked back on). I thought this was a part of an ongoing string of events justifying my suspicions, but since it clearly wasn't, my view was unfounded.

And for your point on NP-hard compiler optimizations... I was wrong 🤦‍♂️. This is a good example of the Dunning-Kruger effect, because I apparently have very surface-level knowledge of how optimizing compilers work and underestimated how much rigorous graph theory actual compilers employed. On another note, I'm now questioning the integrity of my university's compiler course since we covered very little NP-hard algorithms...

Anyways, I didn't mean for any of my comments to come off as aggressive or to phrase things rhetorically. I truly just misunderstood what was going on and overestimated my knowledge on things, and I intend to learn from my mistakes here. Again, sorry for the trouble.

PR #205, opened 2026-05-11
RFC: Support cyclic requires.

andyfriesen:
[Rendered](https://github.com/luau-lang/rfcs/blob/cyclic-requires/docs/support-for-cyclic-requires.md)

We augment the `require()` function, the module interface, and the type checker to allow modules to cyclically require one another in many cases.

### DIFF
--- a/docs/support-for-cyclic-requires.md
+++ b/docs/support-for-cyclic-requires.md
@@ -16,7 +16,7 @@ This problem becomes much more difficult to deal with when classes are added to 
 
 Without cyclic requires, the following program cannot be evaluated.
 
-```
+```luau
 -- A.luau
 
 local B = require("./B")
@@ -76,7 +76,7 @@ Step 6 covers an important edge case: In this design, the `require` function som
 
 The new metatable `CyclicDependencyError` can roughly be defined as follows:
 
-```
+```luau
 local CyclicDependencyError = {
     __index = function(self, prop)
         error(`Cannot access the exported field {prop} because it has a cyclic dependency on its requiring module`)
@@ -89,7 +89,7 @@ local CyclicDependencyError = {
 
 In the absence of `export`, a script must be updated to support cyclic requires by making a small edit: Instead of creating an export table directly with `{}`, the script should accept it from `...` like so:
 
-```
+```luau
 local exports = ...
 
 function exports.foo() end
@@ -112,7 +112,7 @@ If necessary, a module could be crafted to work with or without support for cycl
 
 #### Reentrant Accesses
 
-```
+```luau
 --- A.luau
 
 local B = require("B")
@@ -152,7 +152,7 @@ The order of operations in this program is:
 
 #### Improper Reentrant Mutation
 
-```
+```luau
 --- A.luau
 
 local B = require("B")
@@ -202,7 +202,7 @@ The restrictions on how cyclic imports can be used are subtle\!  If two mutually
 
 For instance, the following code will fail:
 
-```
+```luau
 --- A.luau
 
 local B = require("B")
--- a/docs/support-for-cyclic-requires.md
+++ b/docs/support-for-cyclic-requires.md
@@ -1,0 +1,221 @@
+# Support Cyclic Imports
+
+## Summary
+
+Now that Luau is adding classes to the language, it's much more important that we afford some way modules to cyclically import one another.
+
+This RFC proposes that `require()` be augmented to pass an export table into the module.  This allows the runtime to close the loop and allow many cyclic import scenarios to work as desired.
+
+## Motivation
+
+Luau has always restricted `require()` cycles.  If the runtime encounters a cycle while evaluating a require, it raises an error and stops attempting to load the code.
+
+Prior to our addition of classes as a builtin language feature, this was rarely a big deal because it was always possible to move functions and type definitions into different source files to break any cycles that might arise.  Luau also permits `require()` to be called within a function body.
+
+This problem becomes much more difficult to deal with when classes are added to the mix because classes are always defined at the top level and must always be entirely defined within a single module.
+
+Without cyclic requires, the following program cannot be evaluated.
+
+```
+-- A.luau
+
+local B = require("./B")
+
+class A
+    public children: {B.B}
+
+    function add_child(self)
+        table.insert(self.children, B.B {})
+    end
+end
+
+-- B.luau
+
+local A = require("./A")
+
+class B
+    public parent: A.A
+end
+```
+
+The developer is left to choose between two bad options:
+
+1. They could introduce extra modules that just define interface types for `A` and `B`, or
+2. Move both classes into the same script
+
+Option 1 is laborious and sacrifices the fidelity of the type system.  Option 2 potentially means that the developer's entire program must be specified in a single script\!
+
+## Runtime Design
+
+For modules that return tables, we can solve this issue by having `require` tie the knot: When it encounters a cyclic import, `require` will instead return an empty table that will later be populated with the export surface of the module.  As long as the requesting module doesn't access it at the topmost global scope, that table will eventually be populated and everything will work out.  The system will temporarily attach a metatable to surface these issues and produce a clear error message.
+
+There are subtle edge cases to consider here:
+
+1. If a module fails to access a property from another module because of a cycle, Luau needs to clearly communicate what happened.
+2. If a module acquires a reference to an incomplete module due to a cycle, it should not be able to mutate that module\!
+3. Today, many modules return something other than a table.  It is okay if these modules do not support participation in cycles, but they still need to work as-written.
+4. This proposal requires modules to be adjusted to work with cyclic imports.  Modules that have not been adjusted need to work exactly as-written.
+5. When a non cycle-supporting module appears in a cycle, Luau still needs to communicate the problem to developers clearly.
+
+### Algorithm
+
+`require()` will be adjusted to do the following:
+
+1. First, augment the module of the current script's export table with a new `CyclicDependencyError` metatable.  This metatable prohibits reads and writes to the table by raising an exception with a clear error message.
+2. Look up the requested module in the cache to see if it has already been loaded or begun loading
+3. If a module is in the cache, return it immediately. Otherwise,
+4. Populate the cache with an empty table.
+5. Pass this new table to the target script as its sole argument and evaluate it.  This table can be accessed within the script via `...` at the top level.
+6. Once the module has been evaluated and returned a value, test to see if that value is the same as the table that was passed in.  If they are not the same, set `CyclicDependencyError` as the metatable on the original export table. (the one that wound up not being used)  The table will also be frozen for good measure.
+7. Replace the module cache result with the result of the module
+8. Strip `CyclicDependencyError` from the current script's export table.
+
+This approach handles many cases, but has an important limitation:  A module that participates in a cycle can freely access imported symbols within function bodies, but not at the top level.  This is because those imported symbols cannot be guaranteed to have been evaluated yet.
+
+Step 6 covers an important edge case: In this design, the `require` function sometimes speculatively returns a table with the expectation that it will eventually become the export surface of the requested module.  If it is not, then we have a problem: We have already provided that table to other requesting modules\!  Luckily, this can only happen when we encounter a cycle between modules that do not accept the export table, so all we need to do is to mark that speculative export table as something that cannot be used.
+
+The new metatable `CyclicDependencyError` can roughly be defined as follows:
+
+```
+local CyclicDependencyError = {
+    __index = function(self, prop)
+        error(`Cannot access the exported field {prop} because it has a cyclic dependency on its requiring module`)
+    end,
+    __newindex = function(self, prop, value)
+        error(`Cannot set the exported field {prop} because it has a cyclic dependency on its requiring module`)
+    end
+}
+```
+
+In the absence of `export`, a script must be updated to support cyclic requires by making a small edit: Instead of creating an export table directly with `{}`, the script should accept it from `...` like so:
+
+```
+local exports = ...
+
+function exports.foo() end
+exports.MY_CONSTANT = true
+
+return exports
+```
+
+If necessary, the script could instead adopt a compatibility shim so that it works in older Luau environments that do not implement this RFC: `local exports = ... or {}`
+
+The new `export` keyword will be updated to handle this automatically.
+
+This algorithm satisfies a bunch of important properties:
+
+While existing code will not support cyclic `require()` calls, it will continue to work as-written.  Modules that return non-table values will also continue to work exactly as expected.
+
+If necessary, a module could be crafted to work with or without support for cycles by instead starting with `local exports = ... or {}`.
+
+### Examples
+
+#### Reentrant Accesses
+
+```
+--- A.luau
+
+local B = require("B")
+
+export class Tree
+    children: {B.Node}
+
+    function append(self, prototype: B.Node)
+        -- In this example, we suppose that the tree needs to
+        -- insert a clone of the passed argument.
+        table.insert(self.children, B.Node(prototype))
+    end
+end
+
+--- B.luau
+
+local A = require("A")
+
+-- create a global tree for some reason
+local t = A.Tree{children={}}
+
+export class Node
+
+end
+
+-- main.luau
+
+require("A")
+```
+
+The order of operations in this program is:
+
+1. `main.luau` starts importing `A.luau`
+2. `A.luau` starts importing `B.luau`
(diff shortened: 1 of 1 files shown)


docs/support-for-cyclic-requires.md
```diff
@@ -0,0 +1,221 @@
...
+5. When a non cycle-supporting module appears in a cycle, Luau still needs to communicate the problem to developers clearly.
+
+### Algorithm
+
+`require()` will be adjusted to do the following:
+
+1. First, augment the module of the current script's export table with a new `CyclicDependencyError` metatable.  This metatable prohibits reads and writes to the table by raising an exception with a clear error message.
+2. Look up the requested module in the cache to see if it has already been loaded or begun loading
+3. If a module is in the cache, return it immediately. Otherwise,
+4. Populate the cache with an empty table.
```
hgoldstein: Might be better to indicate this is really one step with two alternatives.

docs/support-for-cyclic-requires.md
```diff
@@ -0,0 +1,221 @@
...
+5. Pass this new table to the target script as its sole argument and evaluate it.  This table can be accessed within the script via `...` at the top level.
```
hgoldstein: Are there any backwards compatibility issues here?

docs/support-for-cyclic-requires.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Type System Design
+
+The user-facing behaviour of the type inference engine should be unchanged as a result of this RFC, but the internal structure of the type checker is going to need significant changes.
+
+Today, typechecking is driven by a class called `Frontend`.  It accepts a set of modules that need checking, builds a DAG from that, and checks modules one after another.
+
+We will augment this class to instead work on one strongly-connected component\* at a time.  All modules within an SCC use the same arena and are typechecked together in a single pass through the solver.
+
+\* A "strongly connected component" is a set of modules that all mutually `require()` one another.
```
hgoldstein: Wording nit: if I didn't know what an SCC was (or forgot because it's been a bit since my graph theory class), I might assume this only meant _direct_ mutually recursive requires, e.g.:
```luau
--- A.luau
local B = require('B')
--- B.luau
local A = require('A')
```
... and not just _any_ set of modules where every one depends on every other via _some_ chain of requires.

docs/support-for-cyclic-requires.md
```diff
@@ -0,0 +1,221 @@
...
+class ClassAOne extends B.ClassOne ... end
+class ClassATwo ... end
+
+--- B.luau
+
+local A = require("A")
+
+class ClassBOne ... end
+class ClassBTwo extends A.ClassATwo ... end
+```
```
hgoldstein, line 222: Food for thought, I think this would get tricker with a (sketched out) `import` syntax:
```
--- A.luau

import ClassBOne from B

export class ClassAOne extends ClassBOne ... end
export class ClassATwo ... end

--- B.luau

import ClassATwo from A

class ClassBOne ... end
class ClassBTwo extends ClassATwo ... end
```

... if we're considering this a serious drawback, this seems like the real danger.

docs/support-for-cyclic-requires.md
```diff
@@ -0,0 +1,221 @@
...
+
+`CyclicDependencyError` saves us here.  We use it to freeze the shape of `B` at step 2\.  It remains frozen until step 5\.  We therefore raise an error in step 4\.
+
+## Type System Design
+
+The user-facing behaviour of the type inference engine should be unchanged as a result of this RFC, but the internal structure of the type checker is going to need significant changes.
+
+Today, typechecking is driven by a class called `Frontend`.  It accepts a set of modules that need checking, builds a DAG from that, and checks modules one after another.
+
+We will augment this class to instead work on one strongly-connected component\* at a time.  All modules within an SCC use the same arena and are typechecked together in a single pass through the solver.
```
alexmccord, line 192: You could also keep them separate, and instead dispatch one constraint solver in the same component, and dispatch the next one if that one is stuck, and so on, until either all of them completes or all of them are stuck.

This would avoid a constraint from module A making its way into some random metavariable's bounds in module B.

MagmaBurnsV:
The ceremony required to make modules that don't use static exports participate in cycles seems silly to me. As the RFC mentions, cycles are already rare for projects prior to classes, and since classes go hand-in-hand with static exports, I see no sense in adding an escape hatch for this edge case.

Just support cycles only for modules that use static exports. That way you don't even have to worry about if a module returns a different value since it's impossible.

docs/support-for-cyclic-requires.md
```diff
@@ -0,0 +1,221 @@
...
+    __newindex = function(self, prop, value)
+        error(`Cannot set the exported field {prop} because it has a cyclic dependency on its requiring module`)
+    end
+}
+```
+
+In the absence of `export`, a script must be updated to support cyclic requires by making a small edit: Instead of creating an export table directly with `{}`, the script should accept it from `...` like so:
+
+```luau
+local exports = ...
```
karl-police, line 96: So, this is ``GETVARARGS``. Now, the thing I know about it, is that it's often used within function bodies.

Doing ``local var = ...`` would mean that it's taking 1 element from the var arg alone.

If I do ``local a,b,c = ...`` it would take 3 and not more.

What would the design have to change here? Or what would this syntax mean if it is within function bodies?
Or does nothing change about ``GETVARARGS`` and it will stay ``1`` element, it's just to accept the ``export table``

karl-police:
Will there be a new kind of type?

I know there's modules that do the following
```luau
-- now we could require("Main.luau") for the varargs?
--- A.luau
return function (varargs)
   print(varargs.OurEnvironment) -- inheritance from upper module, I guess.
end
```
but I guess cyclic requires would solve this too, e.g. for Require Tracer?


```luau
-- Main.luau
local PortionA = require("A.luau")

local Main = {
  OurEnvironment = {}
}

PortionA(Main)
```

docs/support-for-cyclic-requires.md
karl-police: how would hot-reloading a module affect everything, since it talks about cache

docs/support-for-cyclic-requires.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Runtime Design
+
+For modules that return tables, we can solve this issue by having `require` tie the knot: When it encounters a cyclic import, `require` will instead return an empty table that will later be populated with the export surface of the module.  As long as the requesting module doesn't access it at the topmost global scope, that table will eventually be populated and everything will work out.  The system will temporarily attach a metatable to surface these issues and produce a clear error message.
+
+There are subtle edge cases to consider here:
+
+1. If a module fails to access a property from another module because of a cycle, Luau needs to clearly communicate what happened.
+2. If a module acquires a reference to an incomplete module due to a cycle, it should not be able to mutate that module\!
+3. Today, many modules return something other than a table.  It is okay if these modules do not support participation in cycles, but they still need to work as-written.
```
dcope-rbx, line 56: Could you provide more explanation as to why this is ok?
Is it the case that these types of modules don't need to support cycles at all, or that we just consider it a corner case that isn't important enough to handle?

docs/support-for-cyclic-requires.md
```diff
@@ -0,0 +1,221 @@
...
+The developer is left to choose between two bad options:
+
+1. They could introduce extra modules that just define interface types for `A` and `B`, or
+2. Move both classes into the same script
+
+Option 1 is laborious and sacrifices the fidelity of the type system.  Option 2 potentially means that the developer's entire program must be specified in a single script\!
+
+## Runtime Design
+
+For modules that return tables, we can solve this issue by having `require` tie the knot: When it encounters a cyclic import, `require` will instead return an empty table that will later be populated with the export surface of the module.  As long as the requesting module doesn't access it at the topmost global scope, that table will eventually be populated and everything will work out.  The system will temporarily attach a metatable to surface these issues and produce a clear error message.
```
dcope-rbx, line 50: I may not be following all the details here correctly, but would it be problematic if the ModuleScript tries to set a metatable on the export table? i.e. would it clobber the metatable that was already set by the system?

docs/support-for-cyclic-requires.md
dcope-rbx: I am guessing the answer is "yes", but have we thought about whether we could implement this without using `...` to pass the table to the ModuleScript. For example could we instead "patch" a temporary table returned to the requiring script with the contents of the table returned from the ModuleScript?

docs/support-for-cyclic-requires.md
andyfriesen: Does this materially change the problem?  Such a syntax would probably desugar like so:

```luau
local _unnamed = require("./B')
local ClassBOne = _unnamed.ClassBOne
type ClassBOne = _unnamed.ClassBOne
```

It certainly fails in exactly the same way as the example in the RFC, but I think that's acceptable.  You'd get a readable error message, after all.

docs/support-for-cyclic-requires.md
andyfriesen: This RFC describes an extension and a relaxation of the rules.  The proposed implementation should be great from an efficiency perspective, but it has some limitations.  This is one of them.

Given that we're moving to a world where most code just uses `export` rather than explicit `return` statements, it seems like an okay tradeoff if existing code can't take advantage of the new feature, but at least works exactly as well as it used to.

docs/support-for-cyclic-requires.md
andyfriesen: Yes!  This is a pretty bad situation to be in, however: In order to trigger this, you would have to set the metatable on some _other_ module different from the current one.  eg

```luau
local A = require("./A")
setmetatable(A, ...) -- !!!
```

This sounds like extremely bad practice to me!

I can look into what it would take to disallow this.

dcope-rbx: I was actually thinking of the case where a ModuleScript sets its own export table's metatable:
```lua
local exports = ...
setmetatable(exports, {})
```
My understanding is that this would then replace any metatable that the system had already set on the table passed to the ModuleScript.

docs/support-for-cyclic-requires.md
andyfriesen: We could probably add the export table to the global namespace directly and have it only be accessible to `export`.

I could add this to the RFC as an alternative.

docs/support-for-cyclic-requires.md
MagmaBurnsV: This is why I think cycles should only be supported between modules that use the static export system. This issue isn't possible with static exports since the export table isn't directly accessible until after the module has returned.

andyfriesen: Got it.

This isn't good programming style at all, but the adaptation is simple enough: Instead of setting and unsetting the metatable, we'll just save and restore whatever metatable it had.

andyfriesen:
> The ceremony required to make modules that don't use static exports participate in cycles seems silly to me. As the RFC mentions, cycles are already rare for projects prior to classes, and since classes go hand-in-hand with static exports, I see no sense in adding an escape hatch for this edge case.
> 
> Just support cycles only for modules that use static exports. That way you don't even have to worry about if a module returns a different value since it's impossible.

This is another possibility that we talked about, but the implementation of this design is very simple and doesn't really compromise the `export` experience in any way.

I'll add it to the RFC as an alternative.

andyfriesen:
> Wouldn't it be better if a separate keyword was used instead (i.e., `requireDeferred`, `lazyRequire`)? This allows `require` to stay pure instead of throwing new kinds of errors, or worrying about a workaround to capture tuples after returning the cyclic metatable.

This proposal doesn't propose any errors that would not be raised today.  It simply proposes a different mechanism for deciding when to raise that error.

> A second benefit to this is that it handles the case where `A` immediately requires `B` but `B` does not immediately require `A` (which I believe is an edge case that wasn't considered in this RFC).

I don't see how this would be a problem under this proposal.  If `B` requires `A` later, after it has finished evaluating, everything should work just fine.

docs/support-for-cyclic-requires.md
WheretIB: > This is why I think cycles should only be supported between modules that use the static export system. This issue isn't possible with static exports since the export table isn't directly accessible until after the module has returned.

I don't think the export table should be any kind of secret.
You can ask module B to require you and pass your export table back to you.

> My understanding is that this would then replace any metatable that the system had already set on the table passed to the ModuleScript.

You can lock the metatable so that attempt will fail.

docs/support-for-cyclic-requires.md
WheretIB: Transferring contents between tables means there is no unique module table identity preserved and different modules will observe different tables. Doesn't sounds great.

docs/support-for-cyclic-requires.md
MagmaBurnsV: > I don't think the export table should be any kind of secret.
> You can ask module B to require you and pass your export table back to you.

Since that's only possible after module A has frozen and returned the table, that's fine. It's only a problem if the module can access the export table *while* it's being populated, since that allows user code to clobber it like showcased.

My thought was that the metatable attached to the pre-populated table would have a `__metatable` that locks the requiring module from setting a new metatable, which is then stripped by the runtime when that table is passed to the required module to populate. Would this not work?

MagmaBurnsV:
@andyfriesen My concern with supporting non-static-export scenarios is that it means the system has to assume the worst case and do things like save metatables, double check returned values, etc. With the limitations of static exports, these cases aren't possible and the resolution steps can be simplified.

Also, I just think onboarding for learners is going to suck with these mechanics. Nobody is going to understand "you have to grab a proxy table from `...` within both modules, unless this is done internally with `export`" but it will be easier to explain that "only modules that use static exports can be cyclic."

docs/support-for-cyclic-requires.md
WheretIB: It will not be frozen, in A -> B -> A chain, B receives incomplete table of A that it can return to A for the same observation and modification.

WheretIB:
I don't see a point of bifurcating how modules can perform import and export and adding extra checks on what kind of module is being required. Feels largely unnecessary.
By treating evaluation semantics the same, it simplifies implementation.

andyfriesen:
> @andyfriesen My concern with supporting non-static-export scenarios is that it means the system has to assume the worst case and do things like save metatables, double check returned values, etc. With the limitations of static exports, these cases aren't possible and the resolution steps can be simplified.

The implementation is really simple.  It's not much of an imposition on the system.

> Also, I just think onboarding for learners is going to suck with these mechanics. Nobody is going to understand "you have to grab a proxy table from `...` within both modules, unless this is done internally with `export`" but it will be easier to explain that "only modules that use static exports can be cyclic."

I agree that we wouldn't bother to explain this to most users.  The main reason this RFC touches on it at all is because it technically is visible in Luau-land and because it could conceivably be a bit of a gotcha in certain obscure cases. (eg scripts that are run from the commandline under both Lua and Luau)

andyfriesen:
Hey everyone!

The conversation has settled on this one pretty quickly so, unless anything big comes up, I am going to merge this RFC on Monday.

As always, thanks for your feedback!

docs/support-for-cyclic-requires.md
```diff
@@ -0,0 +1,223 @@
...
+The new metatable `CyclicDependencyError` can roughly be defined as follows:
+
+```luau
+local CyclicDependencyError = {
+    __index = function(self, prop)
+        error(`Cannot access the exported field {prop} because it has a cyclic dependency on its requiring module`)
+    end,
+    __newindex = function(self, prop, value)
+        error(`Cannot set the exported field {prop} because it has a cyclic dependency on its requiring module`)
+    end
```
vegorov-rbx: Would be nice to have __metatable key as well to block changes to the metatable on an unfinished table.

docs/support-for-cyclic-requires.md
```diff
@@ -0,0 +1,223 @@
...
+2. Look up the requested module in the cache to see if it has already been loaded or begun loading
+3. If a module is already present in the cache, return it immediately. Otherwise,
+    a. Populate the cache with a fresh table.
+    b. Pass this new table to the target script as its sole argument and evaluate it.  This table can be accessed within the script via `...` at the top level.
+    c. Once the module has been evaluated and returned a value, test to see if that value is the same as the table that was passed in.  If they are not the same, set `CyclicDependencyError` as the metatable on the original export table. (the one that wound up not being used)  The table will also be frozen for good measure.
+    d. Replace the module cache result with the result of the module
+4. Restore the current export table's metatable.
+
+This approach handles many cases, but has an important limitation:  A module that participates in a cycle can freely access imported symbols within function bodies, but not at the top level.  This is because those imported symbols cannot be guaranteed to have been evaluated yet.
```
vegorov-rbx, line 73: Can probably be worded a little better to refer to the evaluation of the top level.
Since top level can call a function and will fail as well.

PR #208, opened 2026-05-24
Format Specification for String Interpolation

Kiiyoko:
I had this one on the backburner for a while. It's basically just an exact rehash of https://github.com/luau-lang/rfcs/pull/20 but uses `;` as the delimiter. I've taken some examples from it.
I'd really like to hear thoughts on this again because it keeps on popping up in my projects and feels trivial to add.
I'm leaving this as a draft pull request for now in case some people really do not like `;`. I feel impartial to it myself.

[Rendered](https://github.com/kiiyoko/luau-rfcs/blob/format-specification-string-interpolation/docs/syntax-string-interpolation-format-specification.md).

## Example
```luau
local balance = 100.2035
print(`You have ${balance;.2f} in your bank account.`) -- 'You have $100.20 in your bank account.' 
```


InfraredGodYT:
I can't help but feel like this syntax is still a bit awkward. I understand why other delimiters weren't chosen, but I'm not sure this approach would be preferred over current options.

```luau
local balance = 100.3921
print(`You have ${string.format("%.2f", balance)} in your bank account.`) 
```

Bottersnike:
`;` feels weird to me as a formatting separator. Even though it's not currently a legal token in this position in the language, `;` in general reads as "statement terminator" (given that's its use elsewhere) so using it as more of an infix operator feels odd.

vegorov-rbx:
Not that I'm a fan of it, but `;` is an alternative table separator in Lua/Luau: `{balance; 2.0}`

Wonder if Scala style is a good choice of the separator:
`` `You have ${balance%.2f} in your bank account.` `` / `` `You have ${balance %.2f} in your bank account.` ``.
It will match exactly the string that is passed instead of `%.*`

Bottersnike:
Would `%` not have the same problem? eg is `{foo%x}` (or `{foo%#x}`) an expression or asking for foo in hex

vegorov-rbx:
Oh yes 😅

Bottersnike:
It would be interesting to see what characters are even available to be used in this position. We've got `;`, and `$` is formally reserved, but I'm not sure what else we have. Off the top of my head maybe only `!`, `@`, `\`, `&` and `|`. Anything that _is_ picked then comes with the reverse problem where it could never be used for infix in the future.

Kiiyoko:
> Not that I'm a fan of it, but `;` is an alternative table separator in Lua/Luau: `{balance; 2.0}`

This is what I was getting at in the document. I agree it’s not pretty but it ‘works’. 

> ```luau
> local balance = 100.3921
> print(`You have ${string.format("%.2f", balance)} in your bank account.`) 
> ```
This might just end up being the best choice, considering that the Luau team seems to want https://github.com/luau-lang/rfcs/pull/170. I’ll likely close this if that ever gets merged. 

> It would be interesting to see what characters are even available to be used in this position.

In the document there’s a small list, I think the only ones I would ever consider are `|`, `&`, `~`, `!`, and `@`. The rest seem too ‘out there’ (such as `=` and `\`). `!` seems strange due to its prevalence as `not` in other languages, but I think it’d be the best choice in Luau. I have no clue what a `!` operator could be used for in the future. 

> Wonder if Scala style is a good choice of the separator: `` `You have ${balance%.2f} in your bank account.` `` / `` `You have ${balance %.2f} in your bank account.` ``. It will match exactly the string that is passed instead of `%.*`

At the very bottom of the document I off-handedly mentioned this. I really wish there was some way for this to work, because I think that it would be the best delimiter. There is the _evil_ option of doubling up on it like `You have ${balance%%.2f} in your...`. Frankly I wouldn’t even be entirely opposed because we have stuff like `::` and `<<T>>()` in the language right now. But, `%%` collides with how `%%` usually escapes `%` in `string.format`.

MagmaBurnsV:
The semicolon is fine with me. It is the only reasonable option that is both backwards and forwards compatible, and the only real argument against it is that it looks odd.

With regards to #170, I think the burden of supporting specifiers should be on that RFC instead of OP. Specifiers are a more ubiquitous feature than structured logging, and personally I don't like how the desugaring in that RFC works. If there should be any interest in it again, a more fleshed out version of "interpolated destructuring" should be made that properly incorporates or at least leaves room for specifiers.

jLn0n:
> Would `%` not have the same problem? eg is `{foo%x}` (or `{foo%#x}`) an expression or asking for foo in hex

problem with using `%` as delimiter is that it is used as modulo operator too though

PR #213, opened 2026-06-24
RFC: Vector constructor syntax

whoamimaybeme:
[Rendered](https://github.com/whoamimaybeme/rfcs/blob/master/docs/vector-syntax.md)

### DIFF
--- a/docs/vector-syntax.md
+++ b/docs/vector-syntax.md
@@ -44,6 +44,7 @@ print(vector.create(1, 2, 3) == <1, 2, 3>) -- true
 ```
 
 If a number is put next to a vector without a comma separating it, it will just compare the two numbers instead of constructing a vector.
+
 ```lua
 function Foo(...)
   for i, v in {...} do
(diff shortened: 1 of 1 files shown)


aatxe:
Your proposed syntax may be problematic for parsing since angle brackets are used as the syntax for binding type parameters. The RFC needs to address how we can still parse this syntax without arbitrary lookahead, but I don't see any mention of the existing syntax that uses angle brackets beyond saying "it won't be an issue."

whoamimaybeme:
> Your proposed syntax may be problematic for parsing since angle brackets are used as the syntax for binding type parameters. The RFC needs to address how we can still parse this syntax without arbitrary lookahead, but I don't see any mention of the existing syntax that uses angle brackets beyond saying "it won't be an issue."

Thank you for pointing that out! I'll work on that.

whoamimaybeme:
I hope that works. If there are any things wrong with it, please tell me! To be honest i dont use generics / type parameters that often

jackhexed:
It was well established in the past that this syntax can be easily parsed.

whoamimaybeme:
> It was well established in the past that this syntax can be easily parsed.

it was?

whoamimaybeme:
ive added some improvements to the rfc to hopefully make it better.

i realized that the whole differentiation between vectors and generics section from earlier probably wasnt needed anyways since parameter names cant start with numbers. i really hope i didnt create some other problem.

docs/vector-syntax.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Alternatives
+
+Do nothing; vectors can already be constructed using `vector.create` or any other runtime provided vector constructor.
+
+Use different syntax for constructing vectors instead of using angle brackets. However, other proposals would cause issues with other features in Luau or would be confusing to work with.
+- `(x, y, z)` is often used for tuples and functions, so it would not work.
+- `[x, y, z]` is used for table indexing, so writing out `newtable[[x, y, z]]` may confuse readers. People may also write `newtable[x, y, z]` expecting for it to index the vector.
+- `vector(x, y, z)` would give the `vector` library a `__call` metamethod, which requires more discussion over whether this is a good idea.
+- `|x, y, z|` isn't valid vector notation.
```
SPY, line 76: wdym by `isn't valid vector notation`. out of all alternative it will probably be less invasive because `|` is not used for `or` and `bitwise or` operators (but it will break potantial compatibility with Lua 5.3 though).

whoamimaybeme: i shouldve just wrote that it isnt commonly used as vector notation.

whoamimaybeme:
closing this for now. i think that the rfc needs a lot more polish in order to actually be a good rfc. the angle bracket syntax probably wouldnt work, and when this rfc comes back, the syntax would probably be `|x, y, z|`

and also i accidentally made this rfc on the master repository instead of a branch

jackhexed:
Why would angle bracket syntax not work?

whoamimaybeme:
> Why would angle bracket syntax not work?

there was a whole discussion in the roblox oss community server that it would cause infinite lookahead or something like that

for instance: `<1, 2, if 3 > somevalue then 4 else 5>`
+ this also might be hard to read

PR #214, opened 2026-06-25
Granular Imports

comicallynathan:
This RFC proposes syntax for allowing named imports

[Rendered view](https://github.com/comicallynathan/rfcs/blob/master/docs/granular-imports.md)

### DIFF
--- a/docs/granular-imports.md
+++ b/docs/granular-imports.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-Allow developers to directly import only necessary parts of libraries or modules as a first-class variable.
+Allow developers to directly import only necessary parts of libraries or modules as first-class variables.
 
 ## Motivation
 
--- a/docs/granular-imports.md
+++ b/docs/granular-imports.md
@@ -1,0 +1,71 @@
+# Granular Imports
+
+## Summary
+
+Allow developers to directly import only necessary parts of libraries or modules as a first-class variable.
+
+## Motivation
+
+- Less boilerplate for importing commonly used methods; repeated prefixing methods with module names (e.g. `Fusion.Computed()`, `Fusion.Observer()`) creates visual noise
+- People who work with React, Fusion, and other components-based toolings
+- Consistency with the incoming [`export`](https://github.com/luau-lang/rfcs/pull/42) keyword.
+- Reinforced with relative imports
+- Potentially enabling any optimization through static and read-only bindings
+
+## Design
+
+### Syntax
+
+Given that this is the module utilizing incoming RFC features
+
+```lua
+export class Dog
+  	private const Sound = "Bark";
+  
+	function New(sound: string): Animal
+		local self = Animal();
+		return self;
+	end
+
+	function Taunt(self: Animal): ()
+		print(self.Sound);
+	end
+end
+```
+
+Selectively import objects, methods, and other values from a module
+
+```luau
+import { Dog } from "./Animals";
+```
+
+Federalize the import
+
+```luau
+import * as Animals from "./Animals";
+const Dog = Animals.Dog;
+```
+
+Granular imports applied in practice
+
+```luau
+import { Computed, Observer, Tween, Spring } from "@ReplicatedStorage/Packages/Fusion";
+```
+
+## Drawbacks
+
+- Lua already has `require()` that can be done in one line. However, it is not too ergonomic, and it can cause visual strain with repeated mentions of the module name or the keyword.
+- This is a huge change that can go against Lua's design philosophy. Though, we must weigh utiliarian and identity.
+- Adding a new import keyword alongside `require()` may cause inconsistencies in the collaborative environments.
+
+## Alternatives
+
+## "Lua" approach
+
+In a more unique Lua approach while maintaining less boilerplate without the keyword, we can use the syntax:
+
+```
+const { Dog } = require("./Animals");
+```
+
+Though, this implies a limited opportunity for developers to be able to rename values to their namespace needs.
--- a/docs/granular-imports.md
+++ b/docs/granular-imports.md
@@ -6,7 +6,7 @@ Allow developers to directly import only necessary parts of libraries or modules
 
 ## Motivation
 
-- Less boilerplate for importing commonly used methods; repeated prefixing methods with module names (e.g. `Fusion.Computed()`, `Fusion.Observer()`) creates visual noise
+- Less boilerplate for importing commonly used methods; repeatedly prefixing methods with module names (e.g. `Fusion.Computed()`, `Fusion.Observer()`) creates visual noise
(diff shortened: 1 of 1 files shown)


soyuta:
I support this.
Destructuring syntax was marked as "not planned" for Luau, and Luau has also said that it does not want to introduce too many new keywords or syntax that differs too much from Lua
but honestly, after adding `export` and `class`, i feel like that principle has started to lose a lot of its meaning

i think it's worth taking another look at this
Without destructuring syntax for imports, I honestly haven't liked any of the solutions proposed so far
I think the TS/JS-style approach is the cleanest

comicallynathan:
> I support this. Destructuring syntax was marked as "not planned" for Luau, and Luau has also said that it does not want to introduce too many new keywords or syntax that differs too much from Lua but honestly, after adding `export` and `class`, i feel like that principle has started to lose a lot of its meaning
> 
> i think it's worth taking another look at this Without destructuring syntax for imports, I honestly haven't liked any of the solutions proposed so far I think the TS/JS-style approach is the cleanest

Destructuring syntax has been a hot topic throughout the existence of Luau. It has been requested two times, and they were denied twice. The reasoning behind them was the lack of predictability and concern for Luau’s identity.

It’s getting harder to maintain an identity once you’ve recognized the need for convenience and tooling. Hence, this is why such features like const, classes, and exports are entering the Luau workspace.

Also, leaving export on its own isn’t very predictable. Once you have export, you’ll need a polarizing keyword to leave it symmetrical. At this point, I would rather have a full ESM-style control on Luau. This addition would be pragmatic, and it leaves less repetition for eye-strain and helps writers track their code easier without the incompetency of repeatedly prefixing commonly used variables as stated in the document. It’s a big leap if we’re asking the question of human-readability.

comicallynathan:
This opens the discussion for a potential structure. Will import be static and top-level? It makes sense to separate require and import as they both serve totally different audiences. Imports may be used for importing external static typing without prefixing. I think that import would make sense as static, while require() would be dynamic. Import could serve as gathering critical utilities and components while requiring will handle dynamic objects like skills and assets.

jackhexed:
The proposed syntax is not backwards compatible or parsable. import { foo } from "bar" is valid syntax today.

jon9933:
This feels like a needless change. you call require() "visual strain" yet how is your proposed solution any less eye straining?

sammygrey:
I understand the premise of a different programming language having different syntax than what you are used to resulting in an "unlearning process" that takes some time to get used to. However, this problem does not exist for programmers who already are proficient in Luau and a change like this would create needless friction for existing developers.

soyuta:
> This feels like a needless change. you call require() "visual strain" yet how is your proposed solution any less eye straining?

> I understand the premise of a different programming language having different syntax than what you are used to resulting in an "unlearning process" that takes some time to get used to. However, this problem does not exist for programmers who already are proficient in Luau and a change like this would create needless friction for existing developers.

Discussions about adding `import` have been going on for quite a while
you can find one of the earlier discussions in #103 as well.

That RFC #103 also proposed an `import` syntax without relying on `require()`:

```luau
import myname, type mytype, myname2, type mytype2 from "mymodule"
```

However, I think adopting destructuring imports with a TypeScript/JavaScript-style syntax would be a better approach.
It would support namespace imports (`*`), type-only imports, and other familiar patterns. since Luau's syntax is already fairly close to TypeScript in many ways, I think this style would fit the language better.

Instead of writing:

```luau
const Module = require("./Module")

const A = Module.A
const AnotherB = Module.B
type U = Module.U
type V = Module.V
```

you could simply write:

```luau
import { A, B as AnotherB, type U, type V } from "./Module"
```

Or, if you want to import everything:

```luau
import * from "./Module"

print(A, B)

const u: U = "U"
const v: V = "V"
```

It's much more concise and significantly easier to read.

More than anything, now that the `export` keyword has been merged in #179, it feels inconsistent to have export without import.

Handling everything with `require()` alone doesn't feel sufficient anymore. 
Since `export` now allows [cyclic requires](https://rfcs.luau.org/support-for-cyclic-requires.html), I think Luau also needs an import system that supports them in the same way.

soyuta:
There is another practical reason why `import` would be useful.

In Luau, when writing modules, it is common for the value returned by a module to have the same name as the module itself.

```luau
src/
  Car/
    init.luau -- returns Car
    Bus.luau -- returns Bus
```

For example, if `Bus` uses `Car`, the code often ends up looking like this:

```luau
const Car = require("./")
const Passenger = require("../Entity/Passenger")

const Bus = {}
Bus.__index = Bus

export type Bus = setmetatable<{
    Passengers: {Passenger.Passenger}
}, typeof(Bus)> & Car.Car

function Bus.new(): Bus end
function Bus:GetNearestCar(): Car.Car end

return Bus
```

One issue here is that types such as `Car.Car` or `Passenger.Passenger` become fairly awkward to read. 
The repeated name is visually noisy, especially when this pattern appears across many modules.

A common workaround is to export the type under a generic name such as `Car.Object` or `Passenger.Object`, but that also has a downside. When reading type hints outside the module, the type may appear as just `Object`, which makes it harder to distinguish from other similarly named exported types.

Another workaround is to create local aliases:

```luau
type Car = Car.Car
type Passenger = Passenger.Passenger

export type Bus = setmetatable<{
    Passengers: {Passenger}
}, typeof(Bus)> & Car
```

This works, but it adds extra declarations purely to avoid repetitive type names. 
In small cases, that means adding extra lines and increasing the diff just to use a type once or twice. 
It also adds another step when tracing where a type originally came from, since the reader has to follow the local alias back to the required module.

The `import` syntax removes the need for separate local type aliases.

```luau
import { type Car } from "./"
import { type Passenger } from "../Entity/Passenger"
```

comicallynathan:
lowk can I be unprofessional

why are Roblox developers so anti-convenience

InfraredGodYT:
> lowk can I be unprofessional
> 
> why are Roblox developers so anti-convenience

It's best to keep things professional on a GitHub repo.

To answer your question, not all Roblox/Luau developers are "anti-convenience", and nobody is disliking this RFC because they don't like convenient syntax. The biggest issue here is the fact that your proposed syntax **isn't backwards compatible.**

```luau
import { foo } from "bar"
```

If you have a function called `import` today, `import { foo }` is technically valid syntax since **Luau function calls don't require parentheses.** Backwards compatibility is a very high priority for adding new features to Luau, and any RFC that doesn't follow this rule likely won't be merged.

It doesn't help that there's an older [import RFC](https://github.com/luau-lang/rfcs/pull/103) that solves the same problems that this one addresses while still being backwards compatible:

```lua
import myname, type mytype, myname2, type mytype2 from "mymodule"
```

I do believe that `import` should be in Luau, but #103 is strictly the better option at the moment.

PR #216, opened 2026-07-05
Alternative Class Syntax & Semantics

InfraredGodYT:
[Rendered](https://github.com/InfraredGodYT/rfcs/blob/patch-1/docs/syntax-classes.md)

This is an alternate draft of the existing Classes RFC intended to encourage discussion before native classes are implemented.

When the original RFC was accepted, I supported it because I was excited to see native classes come to Luau. After reading and participating in the subsequent constructor RFC discussion, however, I came to feel that constructors and encapsulation are fundamental enough to be designed alongside the core class syntax rather than introduced incrementally in later RFCs.

### DIFF
--- a/docs/syntax-classes.md
+++ b/docs/syntax-classes.md
@@ -1,4 +1,5 @@
 
+
 # Classes
 
 ## Summary
@@ -41,7 +42,7 @@ print(p:length()) -- 5.0
 ```
 This RFC proposes native class syntax with:
 - `public` & `private` fields
-- constructors
+- explicit constructors
 - implicit `self` for instance methods
 - static fields & methods
 - nominal typing
@@ -70,7 +71,7 @@ However, a class abstraction is more than a fixed-layout record with attached me
 
 Class definitions are a block construct. `export class X` is allowed. 
 
-Fields are introduced with the new `public` and `private` keywords. Members are public only when explicitly marked `public`.
+Fields are introduced with the new `public` and `private` keywords. Fields are public only when explicitly marked `public`.
 
 ```luau
 class Player
@@ -124,10 +125,12 @@ Field modifiers must appear in a canonical order to improve readability and simp
 2. `static`
 3. `read` (If introduced in the future)
 
-Function declarations do not use access modifiers. Because classes are closed declarations, methods cannot be defined or replaced outside the class body, making separate access modifiers unnecessary.
+Function declarations do not use access modifiers. All instance and static methods are publicly accessible, although they may only be declared within the class body and cannot be replaced afterward.
 
-#### Constructors
-Classes may define a constructor using the contextual keyword `constructor`.
+#### Construction
+Every class has a construction protocol.
+
+A class may declare an explicit constructor using the contextual keyword `constructor`.
 ```luau
 class Player
     public name: string
@@ -137,36 +140,68 @@ class Player
     end
 end
 ```
-Instances are constructed by calling the class.
+Instances are constructed by invoking the class value.
 ```luau
 local p = Player("Builderman")
 ```
+
+If a class does not declare an explicit constructor, the compiler synthesizes a default constructor that accepts exactly one mapping from instance fields to their initial values. Positional calls such as `Class(x, y)` are invalid for such classes.
+```luau
+class Point
+    public x: number
+    public y: number
+end
+
+local point = Point {
+    x = 3,
+    y = 4,
+}
+```
+The mapping may initialize fields accessible at the construction site. Fields omitted from the mapping use their declared default initializers, or `nil` when their types permit it. Values supplied by the field mapping override declared field defaults.
+
+Every field whose declared type does not include `nil` must be initialized either by the field mapping or by a default field initializer. Otherwise, construction is a type error.
+
+```luau
+class Player
+    public name: string
+    public level: number = 1
+    public nickname: string?
+end
+
+local player = Player {
+    name = "Builderman",
+}
+```
+If a class declares any private instance fields, its synthesized default constructor is private and may only be invoked from within the lexical body of that class. Otherwise, the synthesized constructor is public.
+```luau
+class Point
+    private x: number
+    private y: number
+
+    static function origin(): Point
+        return Point {
+            x = 0,
+            y = 0,
+        }
+    end
+end
+
+local origin = Point.origin() -- OK
+local point = Point { x = 1, y = 2 } -- error: default constructor is private
+```
+Declaring an explicit constructor replaces the synthesized field-mapping constructor.
+
 Additional rules include:
-- Only one constructor is allowed per class.
+- Only one explicit constructor is allowed per class.
 - Constructors do not return values.
-- Constructors cannot be invoked directly.
-- Default field initializers execute before the constructor body.
-- If no constructor is declared, a default zero-argument constructor is synthesized. Field default initializers still run. 
-- Fields without default initializers are initialized to `nil`. 
-- A field whose declared type does not include `nil` must be definitely assigned by either a field initializer or every control-flow path through the constructor. Failing to do so is a type error.
+- Constructors cannot be invoked or referenced directly.
+- Default field initializers execute before an explicit constructor body.
+- A field whose declared type does not include `nil` must be definitely assigned by a field initializer or on every control-flow path through an explicit constructor.
 
-For examples of a class with no constructor, this:
-```
-class Player  
-end
-```
-would be identical to this:
-```
-class Player  
-	constructor()  
-	end  
-end
-```
-
-A constructor is a special declaration used to initialize a newly allocated instance. Unlike a factory function, it is automatically invoked when a class object is called and cannot return a value.
+A constructor is a special declaration used to initialize a newly allocated instance. Unlike a factory function, it is automatically invoked when a class value is called.
 
 #### Access Control
-Fields are public only when explicitly marked `public`.  Private members are accessible only from within the lexical body of the enclosing class declaration.
+Fields are public only when explicitly marked `public`.  Private fields are accessible only from within the lexical body of the enclosing class declaration.
 ```luau
 class Player
     private coins: number = 0
@@ -188,7 +223,7 @@ end
 
 #### Static Members
 
-Class members may be declared `static`. Static members belong to the class object rather than class instances and static methods do not receive the implicit `self` binding.
+Class members may be declared `static`. Static members belong to the class value rather than class instances and static methods do not receive the implicit `self` binding.
 
 ```luau
 class Player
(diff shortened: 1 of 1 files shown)


Kampfkarren:
Would help if you clearly documented what is different about this to the original

InfraredGodYT:
> Would help if you clearly documented what is different about this to the original

The Compatibility section lists the differences from the original RFC. Would you like that section to go into more detail explaining why those differences exist?

Bottersnike:
I don't think putting inheritance in "out of scope for now" really makes sense if the aim here is a one-shot of classes. It's something that _is_ going to come up, and regardless of personal views on the correctness of using inheritance it's something people use a lot in their existing code that classes really ought to support if they are to replace use of metatable-based OOP.

As already evidenced, it's something that needs designed in from the get-go to ensure it works properly within classes, so I'd be really interested to at least see discussion about how it could work within this design.

InfraredGodYT:
> I don't think putting inheritance in "out of scope for now" really makes sense if the aim here is a one-shot of classes. It's something that _is_ going to come up, and regardless of personal views on the correctness of using inheritance it's something people use a lot in their existing code that classes really ought to support if they are to replace use of metatable-based OOP.

The aim of this RFC wasn’t a one-shot of classes or to say inheritance should never be in Luau, but to establish a stronger foundation for classes before tackling larger features. The goal here was to establish a more organized core object model that could still be amended in the future.

I left inheritance and composition out of scope because I didn’t have a design for them that I was confident in yet (which is also why I didn’t reserve `protected` and `super` yet). I was leaning more towards composition rather than inheritance, but I wanted more feedback before deciding if just one or both of them would be better to add.

deviaze:
I don't like that this version of the RFC implicitly passes `self` and therefore disallows 'static' functions on a class. 

`static` is a nonobvious term and would be a superficial keyword. A static function on a class should just be one that doesn't take in `self`, simplifying the mental model without adding another keyword.

MagmaBurnsV:
There seems to be a big tension in the design space here that I think needs addressing.

Obviously the current "table constructor" protocol is inspired by Rust's approach, where there is one constructor for a struct/class and all custom constructors are just static factory methods that wrap around the one true constructor.

There's some merit to this approach, as it guarantees initialization is atomic unlike typical OOP constructor setups. If you forget to initialize a field when constructing, then the typechecker will warn you about it, which is something not trivially doable with constructor inheritance. It also meshes nicely with a trait/interface system like the one Rust has.

However, the downside is that this doesn't compose at all with inheritance, since you can't just "extend" a static factory method. This is what the previous constructor RFC tried and failed to resolve. Atomic constructors are simply mutually exclusive to inheritance.

The irony is that people seem to *really* want inheritance, but to do so requires changing the constructor protocol to be nonatomic and unchecked (like this RFC proposes), which is unpopular. Likewise, we could keep the atomic constructor protocol, but then our only options for polymorphism are interfaces, which also seems unpopular.

InfraredGodYT:
> I don't like that this version of the RFC implicitly passes `self` and therefore disallows 'static' functions on a class.
> 
> `static` is a nonobvious term and would be a superficial keyword. A static function on a class should just be one that doesn't take in `self`, simplifying the mental model without adding another keyword.

I didn't write out the section yet at the time of this reply, but implicit `self` would not apply to static functions or fields, which would use the `static` keyword. Static functions wouldn't be disallowed.

I plan to amend the RFC with a section dedicated to the semantics of `static` to clear up any confusion surrounding it.

Bottersnike:
> However, the downside is that this doesn't compose at all with inheritance, since you can't just "extend" a static factory method. This is what the previous constructor RFC tried and failed to resolve. Atomic constructors are simply mutually exclusive to inheritance.

I think the bigger problem with `Bar({ foo = 1 })` isn't inheritance but rather just constructors in _general_. Factory functions are an option but aren't particularly ergonomic.

The Rust model works when you think about classes as just POD classes that have some methods on it but that's quite a specific model (akin to a C struct with some functions pointers on it) that doesn't particularly feel like "classes" in the ergonomic native way you'd expect from a high level construct.

soyuta:
I like this RFC's implicit `self`, similar to how methods are currently declared with `:` in Luau, and I also like the introduction of `private`.

While using `_camelCase` for private members has been a long-standing convention, I think it's also reasonable for the language itself to enforce access restrictions, just like it does with `const`.

However, I don't like constructing classes by calling the class object itself like a function. It would break a long-standing convention in the Luau ecosystem, where native objects have traditionally used `.create()` while user-defined objects have used `.new()`.

I'm also not sure about introducing a `constructor` keyword. It feels a bit unfamiliar in the context of Lua. Considering Lua's metamethod naming conventions, I think `__init` (as proposed in RFC #210) would be a better fit.

Or maybe it would be better to introduce a `constructor` modifier instead?

```luau
constructor function new()
end

constructor function create()
end
```

Kind of like `public` and `private`.

Constructors really seem to be the trickiest part of the class design.

Bottersnike:
One problem `.new()` and `.create()` end up with is you're now reserving those method names to never be usable. They also don't match the `function __init` that was defined, which is Very Strange behaviour (see: unintuitive). `Foo()` to call the constructor means no names are reserved, matches what all other languages use for classes, and matches the "somewhat magic" nature of a constructor (be that via `constructor() ...` or `function __init`.

It's a pattern that's quite easily doable right now with metamethods by defining `__call` as a metamethod on the table that would otherwise have `.new()`, but it's not a frequently employed pattern because it's honestly a pain to write out all that code.

docs/syntax-classes.md
```diff
@@ -83,9 +110,69 @@ Classes can define the following Luau metamethods.  They all work just like they
...
+        self.name = name
+    end
+end
+```
+Instances are constructed by calling the class.
+```luau
+local p = Player("Builderman")
+```
+Additional rules include:
+- Only one constructor is allowed per class.
```
dyslexicsteak: The concept of a constructor in object oriented languages hinges on overloading or some sort of dispatch to remain useful, this restriction significantly reduces usability. I see this maybe being relegated to acting as a self-type aggregate initialiser simply because you have to use it, but de facto factories handling the actual logic as they have much more flexible signatures. There would also probably be a split of some using tables and others using positionals here.

Plus, the fact the constructor is outside of access control (as far as I can tell) requires adding validation logic inside it as it's always accessible, so you'd end up either `pcall`ing it, with double validation logic, or with invalid objects. I'd recommend cutting the middleman and just using object/table init syntax of some sort rather than requiring bugprone assignment boilerplate.

docs/syntax-classes.md
```diff
@@ -83,9 +110,69 @@ Classes can define the following Luau metamethods.  They all work just like they
...
+- Constructors do not return values.
```
dyslexicsteak: The pointer-to-uninit signature of constructors coupled with this forces fallible construction to resort to exceptions to signal errors or resort to a sentinel class value, which I find both restrictive and not ideal.

docs/syntax-classes.md
```diff
@@ -83,9 +110,69 @@ Classes can define the following Luau metamethods.  They all work just like they
...
+- Constructors cannot be invoked directly.
```
dyslexicsteak: Not allowing higher-order functions like `pcall`, maybe the `task` library, or such to deal with constructors seems like another limitation that is unnecessary and actively sabotages some things users might want to do to deal with fallible constructors (as mentioned on return) and yielding constructors. This also sheds light on how a constructor is not really constructing anything at all, but simply functions as an initialisation hook, and so is inherently not freestanding.

docs/syntax-classes.md
```diff
@@ -244,47 +351,28 @@ Object oriented codebases tend to have far more cyclic dependencies between modu
...
+An explicit `self` can make instance methods more readable but creates boilerplate code. Type annotations for `self` also wouldn't be allowed.
+
+### Constructor syntax
+This RFC uses a dedicated `constructor` declaration instead of requiring constructors to be expressed as ordinary methods or metamethod-like hooks.  
+  
+A dedicated constructor has a few advantages:  
+  
+- It separates object initialization from normal instance behavior.  
+- It avoids making construction look like an ordinary method call.  
+- It avoids reserving conventional method names such as `new`.
```
dyslexicsteak, line 447: I haven't looked at the accepted proposal, but I greatly value factories over this because they do not have to conform to a certain restrictive signature. I don't believe objects in an unready state, such as between initial creation and initialisation, should ever be observed.

docs/syntax-classes.md
```diff
@@ -202,39 +291,57 @@ local globalCount = 0
...
+3. Invoke the class constructor, if one exists.
+4. Return the initialized instance.
+
 ### Out of scope for now
 
-#### Private fields, const fields
+#### Const/Static fields
 
-These are things we want to do, but integrating them with the existing structural type system is surprisingly tricky and will be tackled in separate RFCs.
+The keyword `static` will be reserved for future RFCs.
```
dyslexicsteak: I think classes without some kind of `static` mechanism are awkward, should be avoided, and that adding methods to the class object itself isn't too complex to discuss here and worth including to avoid this unfortunate intermediate state.

docs/syntax-classes.md
```diff
@@ -202,39 +291,57 @@ local globalCount = 0
 class Counter
     public count: number
 
-    function new()
-        local count = globalCount
+    constructor()
+        self.count = globalCount
         globalCount += 1
-        return Counter {count=count}
```
dyslexicsteak, line 208: I really don't mind this, and having private fields hide this would be a very simple way to create complete encapsulation.

docs/syntax-classes.md
InfraredGodYT: I intentionally limited classes to one `constructor` because Luau does not support function overloading. I also think that factory functions would be better expressed as static functions rather than making the constructor itself overloadable or return values.

docs/syntax-classes.md
InfraredGodYT: `pcall` would work like this:

```luau
local ok, player = pcall(Player, "Builderman")
```

but not like this:

```luau
local ok, player = pcall(Player.constructor, ...)
```

docs/syntax-classes.md
InfraredGodYT: I agree. If instance methods use an implicit `self` then class-level functions need explicit syntax, which is why I plan to add `static` members to the design of this RFC.

docs/syntax-classes.md
InfraredGodYT: I agree with this as well. The main distinction I’m trying to make here is that private fields make encapsulation part of the class declaration itself, rather than relying on module structure.

docs/syntax-classes.md
InfraredGodYT: I agree that factories are valuable, and I'm not attempting to remove that pattern. I'm separating it from initialization.

Under this design, `constructor` handles the initialization path for `Class(...)`, while alternate/fallible constructors would be static factory functions.

dyslexicsteak: Then why force people into boilerplate when there's a reasonable default?

InfraredGodYT: I don’t think the two solve the same problem.

A dedicated constructor defines the canonical initialization path for `Class(...)` and allows the language to reason about object initialization directly. By making construction part of the language rather than an ordinary factory function, the compiler can statically verify that every non-optional field is definitely assigned before an instance is returned to user code. 

This avoids exposing partially initialized objects while still allowing static factory functions for alternate or fallible construction.

HarmosCreations:
Can the `self` definition be shared through out the methods this time?
if not we could try and replace it with `this` or anything else that does that.

also is there a way to introduce an access-control that is private to write but is public to read.
this will save us methods that only return private fields.

Example:

```luau

class Player
    private coins: number = 0

    function Coins()
        return self.coins
    end
end

-- replace with this

class Player
    protect coins: number = 0
end

```

Neverless this whole thing is quite a big change for luau.

PR #219, opened 2026-07-07
Add labeled break and continue statements

SPY:
[Rendered](https://github.com/SPY/rfcs/blob/syntax-labeled-break-continue/docs/syntax-labeled-break-continue.md).

### DIFF
--- a/docs/syntax-labeled-break-continue.md
+++ b/docs/syntax-labeled-break-continue.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-Introduce a labeled form of loop and `do` block statements, and extend `break` and `continue` to optionally reference a label. The label is written as a colon-suffix on the loop's `do` keyword — `for i = 1, 10 do: outer` — making it structurally bound to the block it names. This allows structured non-local exits from nested loops without workarounds such as flag variables or immediately-invoked function expressions.
+Introduce a labeled form of loop, and extend `break` and `continue` to optionally reference a label. The label is written as a colon-suffix on the loop's `do` keyword — `for i = 1, 10 do: outer` — making it structurally bound to the loop it names. This allows structured non-local exits from nested loops without workarounds such as flag variables or immediately-invoked function expressions.
 
 ## Motivation
 
@@ -51,10 +51,9 @@ Labeled `break` and `continue` address all of these problems with a straightforw
 
 ### Syntax
 
-A label is introduced by appending `:` and a name directly to the `do` keyword that opens a loop body or standalone block:
+A label is introduced by appending `:` and a name directly to the `do` keyword that opens a loop body:
 
 ```
-doblock  ::= 'do' ':' Name block 'end'
 forstat  ::= 'for' ... 'do' ':' Name block 'end'
 whilestat ::= 'while' exp 'do' ':' Name block 'end'
 repeatstat ::= 'repeat' ':' Name block 'until' exp
@@ -78,8 +77,6 @@ When no label is given, `break` and `continue` behave exactly as today: they ref
 ### Scoping rules
 
 A label is in scope from the `do:` (or `repeat:`) token to the end of the enclosing block. It is visible in all nested blocks and functions — but only for the purpose of `break`/`continue`; a label cannot be referenced in any other context, and it is never captured as an upvalue.
-
-A label on a `do` block may be targeted by `break` but not `continue`, because `do` blocks do not iterate. Attempting to `continue` a `do` block is a compile-time error.
 
 Label names must be unique within their immediately enclosing block. Redeclaring a label name at the same scope level is a compile-time error. Shadowing an outer label with an inner label of the same name is permitted, following the same rules as variable shadowing.
 
@@ -126,19 +123,6 @@ for _, row in grid do: outer
         end
         process(cell)
     end
-end
-```
-
-**Breaking a `do` block to skip a section of code:**
-
-```lua
-do: validate
-    if not config.enabled then break validate end
-    if config.value < 0 then
-        warn("negative value")
-        break validate
-    end
-    applyConfig(config)
 end
 ```
 
@@ -200,8 +184,6 @@ end
 **Contextual keyword extension.** `continue` is already a contextual keyword. `break label` and `continue label` require the parser to recognize that the identifier following `break`/`continue` on the same line is a label reference. This is straightforward but must be carefully specified to avoid edge cases at line boundaries.
 
 **Increased language surface area.** Labels are a feature programmers must learn. Deeply nested loops with non-obvious label names can reduce readability. Poor use of labeled breaks can produce code nearly as hard to follow as `goto`.
-
-**`do` block labeling asymmetry.** Allowing `break` but not `continue` on `do` blocks introduces an asymmetry that users must remember.
 
 ## Alternatives
 
--- a/docs/syntax-labeled-break-continue.md
+++ b/docs/syntax-labeled-break-continue.md
@@ -1,0 +1,194 @@
+# Labeled break and continue
+
+## Summary
+
+Introduce optional labels on loop statements (`for`, `while`, `repeat`) and `do` blocks, and extend `break` and `continue` to optionally reference a label. This allows structured non-local exits from nested loops without workarounds such as flag variables or immediately-invoked function expressions.
+
+## Motivation
+
+It is common in Luau code to iterate over nested data structures and want to abort or skip iteration at an outer level once some condition is met. Today, Luau provides no direct mechanism for this. The available workarounds each have significant problems.
+
+**Flag variables** add noise and can obscure intent:
+
+```lua
+local found = false
+for _, chunk in chunks do
+    for _, entity in chunk.entities do
+        if entity:matches(target) then
+            found = true
+            break
+        end
+    end
+    if found then break end
+end
+```
+
+**Immediately-invoked function expressions (IIFEs)** are commonly used in Luau codebases:
+
+```lua
+local result = (function()
+    for _, chunk in chunks do
+        for _, entity in chunk.entities do
+            if entity:matches(target) then
+                return entity
+            end
+        end
+    end
+    return nil
+end)()
+```
+
+This pattern has several drawbacks:
+
+- It reuses `return` with different semantics than the enclosing function, which can confuse readers.
+- Luau's type checker must reason through the IIFE boundary, which complicates type narrowing and return type inference.
+- It is stylistically alien — the construct exists solely to work around a language limitation rather than to express intent.
+- If compiled without optimization(or inlining doesn't happen), it introduces a closure allocation and call overhead. The IIFE takes zero arguments and captures locals as upvalues. Luau's inlining cost model deprioritizes such functions, so the call is unlikely to be inlined. Even when inlining does occur, the lambda's prototype remains in the bytecode — occupying space and contributing to parse time.
+
+Labeled `break` and `continue` address all of these problems with a straightforward, well-precedented mechanism found in Java, Kotlin, JavaScript, Rust, and other languages. It is the natural complement to the `continue` statement already present in Luau.
+
+## Design
+
+### Syntax
+
+A label may be placed immediately before a `for`, `while`, `repeat`, or `do` statement. The label uses double-colon delimiters, matching the label token form used in Lua 5.2:
+
+```
+labeledstat ::= '::' Name '::' loopstat
+loopstat    ::= forstat | whilestat | repeatstat | dostat
+```
+
+The label name is a plain identifier and follows normal Luau identifier rules. Label names share no namespace with local variables, upvalues, or global variables; a label named `i` does not shadow or conflict with a variable named `i`.
+
+`break` and `continue` are extended to accept an optional label name:
+
+```
+breakstat    ::= break [Name]
+continuestat ::= continue [Name]
+```
+
+Note that the label *reference* in `break` and `continue` uses a bare name — only the *declaration* uses `::name::`. This keeps the call-site syntax concise while making label declarations visually prominent and syntactically unambiguous.
+
+When no label is given, `break` and `continue` behave exactly as today: they refer to the innermost enclosing loop.
+
+When a label is given, `break label` exits the statement identified by `label`, and `continue label` skips to the next iteration of the loop identified by `label`.
+
+### Scoping rules
+
+A label is in scope from the point of its declaration to the end of the enclosing block. It is visible in all nested blocks and functions — but only for the purpose of `break`/`continue`; a label cannot be referenced in any other context, and it is never captured as an upvalue.
+
+A label that names a `do` block may be targeted by `break` but not `continue`, because `do` blocks do not iterate. Attempting to `continue` a `do` block is a compile-time error.
+
+Label names must be unique within their immediately enclosing block. Redeclaring a label name at the same scope level is a compile-time error. Shadowing an outer label with an inner label of the same name is permitted, following the same rules as variable shadowing.
+
+It is a compile-time error to `break` or `continue` a label that is not in scope.
+
+### Grammar disambiguation
+
+Because label declarations begin with `::`, there is no ambiguity with any existing Luau syntax. The `::` token is not valid at the start of a statement expression, so the parser can commit to the label rule as soon as it sees `::` in statement position.
+
+Label references in `break label` and `continue label` use a bare identifier. Since `break` and `continue` are keywords that currently take no arguments, any identifier following them on the same line is unambiguously a label reference. The parser handles this with a single-token lookahead after `break`/`continue`: if the next token is an identifier, it is consumed as the label name.
+
+### Interaction with `continue`
+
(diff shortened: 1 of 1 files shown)


Dekkonot:
My preference would be to have the label follow the `do` I think, since it makes it more obvious that it's attached to that scope rather than just the loop. It also solves some of the ambiguity problems of less ugly syntax (my apologies to anyone who is in love with `::label::`) because the position is already a syntax error with something as simple as a colon:

```luau
for i = 1, 10 do: label
  if i == 8 then
    continue label
  end
end
```

While I'm not strictly proposing this syntax (it gets weird due to e.g. `do:label("foo"):lower()`), it gives us more options than making it a prefix to the keywords.

`repeat` has the same property, which is that a lot of symbols after it are just outright syntax errors right now.

bradsharp:
While this would be a nice feature I am not sure about the syntax to to add a label. Could we lean into attributes here with something like `@label <name>`? Alternately, could this be a number where `break 1` is equivalent to `break` and `break 2` would break the outer loop?

nnullcolumn:
> While this would be a nice feature I am not sure about the syntax to to add a label. Could we lean into attributes here with something like `@label <name>`? Alternately, could this be a number where `break 1` is equivalent to `break` and `break 2` would break the outer loop?

`break 1` or `break 2`, albeit admirably minimal, seem like they could cause mistakes - e.g. moving nested code from one spot to another. Attributes historically haven't changed the runtime semantics of programs (apart from arguably native in niche float ways), so it feels bad to overload them for control flow. I like the solution dekkonot proposed even if it allows people to mangle their syntax a bit.

SPY:
> My preference would be to have the label follow the do I think, since it makes it more obvious that it's attached to that scope rather than just the loop.

I like it.

> my apologies to anyone who is in love with `::label::`

Never was attached to it. It looks ugly :)

Nicant02:
I like this, something like this would be nice to have.

I enjoy the extended do block syntax, an uncommon pattern I do sometimes when I need to check a bunch of different things that chain off of each other is something like:
```luau
-- execution starts
repeat
	local character = Player.Character
	if not character then
		break
	end
	
	local humanoid = character:FindFirstChildOfClass("Humanoid")
	if not humanoid then
		break
	end
	ApplyDamage(humanoid, 10)
until true
-- execution continues
```

InfraredGodYT:
> > My preference would be to have the label follow the do I think, since it makes it more obvious that it's attached to that scope rather than just the loop.
> 
> I like it.
> 
> > my apologies to anyone who is in love with `::label::`
> 
> Never was attached to it. It looks ugly :)

I was unsure about the syntax at first, but using `do: label` sealed the deal for me. If this isn’t too complex to add, it’ll definitely be a nice addition.

alexmccord:
Even `label: (do|for|while|repeat)` and `break label` are syntax that can be supported retroactively. No need for `::label::` gunk or postfix `kw: label`. And quite frankly, I don't want to read `for k, v in t do: label`, I want to read `label: for k, v in t do`.

nnullcolumn:
> Even `label: (do|for|while|repeat)` and `break label` are syntax that can be supported retroactively. No need for `::label::` gunk or postfix `kw: label`. And quite frankly, I don't want to read `for k, v in t do: label`, I want to read `label: for k, v in t do`.

imo, `label: for` is crazy visual noise and looks like a type annotation in all the wrong ways

SPY:
After internal discussion with the team we decided to split this PR in two moving away labeled `do` blocks.
The reason for that is potential inconsistency: do blocks going to support only labeled breaks, introduction of unlabeled breaks for do blocks breaks backward compatibility with code like:
```lua
for i in 1,10 do
  if i > 5 then do break end end
end
```
We still consider breaking labeled `do` blocks reasonable language feature, but we think it deserves own RFC.
If there is no significant objections for labeled loops/break/continue I'm going to merge that RFC next week

docs/syntax-labeled-break-continue.md
```diff
@@ -0,0 +1,201 @@
...
+    return nil
+end)()
+```
+
+This pattern has several drawbacks:
+
+- It reuses `return` with different semantics than the enclosing function, which can confuse readers.
+- Luau's type checker must reason through the IIFE boundary, which complicates type narrowing and return type inference.
+- It is stylistically alien — the construct exists solely to work around a language limitation rather than to express intent.
+- If compiled without optimization(or inlining doesn't happen), it introduces a closure allocation and call overhead. The IIFE takes zero arguments and captures locals as upvalues. Luau's inlining cost model deprioritizes such functions, so the call is unlikely to be inlined. Even when inlining does occur, the lambda's prototype remains in the bytecode — occupying space and contributing to parse time.
```
vegorov-rbx: Not sure this drawback is true any more with the adjusted cost model.
And bytecode remaining is something we want to address outside of this case.

Not saying it makes the approach less clunky, but I don't think RFC should include it.

docs/syntax-labeled-break-continue.md
```diff
@@ -0,0 +1,201 @@
...
+for i = 1, 10 do: outer
+    for j = 1, 10 do: outer  -- OK: shadows outer label in this scope
+        if j == 5 then break outer end  -- breaks the inner 'outer' label
+    end
+end
+```
+
+### Bytecode
+
+`break label` compiles to the same `JUMP` instruction as an unlabeled `break`, targeting the instruction immediately following the labeled statement. `continue label` compiles to the same `JUMP` as an unlabeled `continue`, targeting the loop condition of the labeled statement. No new bytecode instructions are required.
```
vegorov-rbx: This is very lightly specified considering that unlabeled breaks do more than just jumps, they also need to close upvalues.

docs/syntax-labeled-break-continue.md
```diff
@@ -0,0 +1,201 @@
...
+```
+breakstat    ::= break [Name]
+continuestat ::= continue [Name]
+```
+
+When no label is given, `break` and `continue` behave exactly as today: they refer to the innermost enclosing loop. When a label is given, `break label` exits the labeled statement and `continue label` skips to the next iteration of the labeled loop.
+
+### Scoping rules
+
+A label is in scope from the `do:` (or `repeat:`) token to the end of the enclosing block. It is visible in all nested blocks and functions — but only for the purpose of `break`/`continue`; a label cannot be referenced in any other context, and it is never captured as an upvalue.
```
vegorov-rbx: Does it have to be visible in nested functions?
What does breaking on it do, jump across functions?

At the end you mention it is never captured as an upvalue which seems like you want to prevent this, but labels are not values, so this is confusing.

docs/syntax-labeled-break-continue.md
```diff
@@ -0,0 +1,201 @@
...
+
+Label names must be unique within their immediately enclosing block. Redeclaring a label name at the same scope level is a compile-time error. Shadowing an outer label with an inner label of the same name is permitted, following the same rules as variable shadowing.
+
+It is a compile-time error to `break` or `continue` a label that is not in scope.
+
+### Grammar disambiguation
+
+`do` is a keyword with no existing interpretation that allows `:` to follow it, so `do:` is entirely unambiguous at the parser level. No lookahead is required — the parser commits to the labeled form as soon as it sees the `:` token immediately after `do` (or `repeat`).
+
+Label references in `break label` and `continue label` use a bare identifier. Since `break` and `continue` are keywords that currently take no arguments, any identifier following them on the same line is unambiguously a label reference.
```
vegorov-rbx: `continue` is only a contextual keyword.

docs/syntax-labeled-break-continue.md
alexmccord: Yeah, I don't want it to be visible across function scopes, that's basically reintroducing `goto` with extra syntax.

SPY: functions removed

docs/syntax-labeled-break-continue.md
SPY: Intention was to highlight the fact it is a compiler-only feature. Mentioned `CLOSEUPVALS` as well

docs/syntax-labeled-break-continue.md
SPY: RFC was submitted before the fix. Removed

docs/syntax-labeled-break-continue.md
```diff
@@ -0,0 +1,201 @@
...
+    return nil
+end)()
+```
+
+This pattern has several drawbacks:
+
+- It reuses `return` with different semantics than the enclosing function, which can confuse readers.
+- Luau's type checker must reason through the IIFE boundary, which complicates type narrowing and return type inference.
+- It is stylistically alien — the construct exists solely to work around a language limitation rather than to express intent.
+- If compiled without optimization(or inlining doesn't happen), it introduces a closure allocation and call overhead.
```
nnullcolumn, line 46: there should be a space before the parenthesized text here

docs/syntax-labeled-break-continue.md
```diff
@@ -0,0 +1,201 @@
+# Labeled break and continue
+
+## Summary
+
+Introduce a labeled form of loop, and extend `break` and `continue` to optionally reference a label. The label is written as a colon-suffix on the loop's `do` keyword — `for i = 1, 10 do: outer` — making it structurally bound to the loop it names. This allows structured non-local exits from nested loops without workarounds such as flag variables or immediately-invoked function expressions.
```
nnullcolumn, line 5: i think 'a labeled form of loop', 'extend `break and continue`', 'label is written as a colon-suffix on the loop's `do` keyword', etc... are good things to communicate, but could be said more concisely by replacing this entire section with something along the lines of 'introduce named control flow manipulation in loops with `do: label`, `repeat: label`, `break label`, and `continue label`'

docs/syntax-labeled-break-continue.md
```diff
@@ -0,0 +1,201 @@
...
+        if entity:matches(target) then
+            found = true
+            break
+        end
+    end
+    if found then break end
+end
+```
+
+**Immediately-invoked function expressions (IIFEs)** are commonly used in Luau codebases:
```
nnullcolumn, line 26: 'commonly used' seems very subjective, there are lots of luau codebases without this pattern. perhaps 'may be used' would be better?

docs/syntax-labeled-break-continue.md
```diff
@@ -0,0 +1,201 @@
...
+continuestat ::= continue [Name]
+```
+
+When no label is given, `break` and `continue` behave exactly as today: they refer to the innermost enclosing loop. When a label is given, `break label` exits the labeled statement and `continue label` skips to the next iteration of the labeled loop.
+
+### Scoping rules
+
+A label is in scope from the `do:` (or `repeat:`) token to the end of the enclosing block. It is visible in all nested blocks — but only for the purpose of `break`/`continue`; a label cannot be referenced in any other context, and it is never captured as an upvalue.
+
+Label names must be unique within their immediately enclosing block. Redeclaring a label name at the same scope level is a compile-time error. Shadowing an outer label with an inner label of the same name is permitted, following the same rules as variable shadowing.
```
nnullcolumn, line 81: > Label names must be unique within their immediately enclosing block.

this wording makes no sense to me. if the meaning is what follows it, why have this at all?

> Redeclaring a label name at the same scope level is a compile-time error. Shadowing an outer label with an inner label of the same name is permitted, following the same rules as variable shadowing.

why a compile error for making the same label twice in the same scope? does this imply:

```luau
repeat: retries
    local ok, err = attempt()
    if err == "fatal" then break retries end
until ok or retries_exceeded()
-- invalid
repeat: retries
    local ok, err = attempt_other()
    if err == "fatal" then break retries end
until ok or retries_exceeded()
```

if so, that seems *extremely* different from local shadowing semantics - those scopes are completely unrelated, since the label stops existing as soon as you leave the loop. secondly, there already exist lints for shadowing locals. i'd argue if anything we should lint or compile error the opposite way - inner shadowing disallowed because it results in rules which may be significantly more confusing, but otherwise there should just be no limitation imo

docs/syntax-labeled-break-continue.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Bytecode
+
+`break label` compiles to the same `JUMP` instruction as an unlabeled `break`, targeting the instruction immediately following the labeled statement(extra `CLOSEUPVALS` for all outer loops should be also emitted though). `continue label` compiles to the same `JUMP` as an unlabeled `continue`, targeting the loop condition of the labeled statement. No new bytecode instructions are required.
+
+## Drawbacks
+
+**`repeat` asymmetry.** All other labeled constructs attach the label to `do`, which is the shared block-opening keyword. `repeat` has no `do`, so its label sits on `repeat` itself (`repeat: name`). This is a minor inconsistency programmers must learn.
+
+**Contextual keyword extension.** `continue` is already a contextual keyword. `break label` and `continue label` require the parser to recognize that the identifier following `break`/`continue` on the same line is a label reference. This is straightforward but must be carefully specified to avoid edge cases at line boundaries.
```
nnullcolumn, line 184: > This is straightforward but must be carefully specified to avoid edge cases at line boundaries.

could this be explained further? i don't really see how this results in any edge cases, and line boundaries feel orthogonal

docs/syntax-labeled-break-continue.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Grammar disambiguation
+
+`do` is a keyword with no existing interpretation that allows `:` to follow it, so `do:` is entirely unambiguous at the parser level. No lookahead is required — the parser commits to the labeled form as soon as it sees the `:` token immediately after `do` (or `repeat`).
+
+Label references in `break label` and `continue label` use a bare identifier. Since `break` is a keyword and `continue` is contextual keywords that currently take no arguments, any identifier following them on the same line is unambiguously a label reference.
+
+### Interaction with `continue`
+
+`continue` is already a contextual keyword in Luau. The labeled form `continue Name` is parsed the same way: `continue` followed by an identifier on the same line. Since `continue` only appears inside loop bodies and cannot start a function call, there is no ambiguity with expression statements.
```
nnullcolumn, line 93: > Since `continue` only appears inside loop bodies and cannot start a function call, there is no ambiguity with expression statements.

this sentence doesn't make a lot of sense. why would a keyword start a function call, and how is that related? additionally, expressions and statements are fundamentally disjoint in luau - there are no expression statements. there are things which are valid as expressions and valid as statements, but to me this paragraph seems really unnecessary generally?

docs/syntax-labeled-break-continue.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Type checking
+
+Labels have no direct type implications. The type checker treats labeled `break` the same as an unconditional `break` for the purposes of control-flow analysis: it is a definite exit from the labeled statement. Labeled `continue` is treated as a definite jump to the loop condition of the labeled statement.
```
nnullcolumn, line 97: that doesn't make a lot of sense, you say it has no implications and then explain how it behaves in control flow analysis. isn't non-local jumping from an inner loop an inherently more complicated control flow which wasn't possible before? even the drawbacks section addresses this directly.

the behavior here is also worded somewhat confusingly explaining the implications of this, possibly because it seemingly ignores potential implications

docs/syntax-labeled-break-continue.md
```diff
@@ -0,0 +1,201 @@
...
+for i = 1, 10 do: outer
+    for j = 1, 10 do: outer  -- OK: shadows outer label in this scope
+        if j == 5 then break outer end  -- breaks the inner 'outer' label
+    end
+end
+```
+
+### Bytecode
+
+`break label` compiles to the same `JUMP` instruction as an unlabeled `break`, targeting the instruction immediately following the labeled statement(extra `CLOSEUPVALS` for all outer loops should be also emitted though). `continue label` compiles to the same `JUMP` as an unlabeled `continue`, targeting the loop condition of the labeled statement. No new bytecode instructions are required.
```
nnullcolumn, line 178: space missing before parenthesized section here

nnullcolumn, line 178: perhaps the `CLOSEUPVALS` bytecode should be mentioned in the drawbacks section? a vm developer would have to tell me if that's super significant because idrk the implementation of that

docs/syntax-labeled-break-continue.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Drawbacks
+
+**`repeat` asymmetry.** All other labeled constructs attach the label to `do`, which is the shared block-opening keyword. `repeat` has no `do`, so its label sits on `repeat` itself (`repeat: name`). This is a minor inconsistency programmers must learn.
```
nnullcolumn, line 182: although i'm not against a very exhaustive drawbacks section, i'm not sure this item is very significant. it could also be merged with the later one that talks about the learnability of this syntax.

nnullcolumn (commented):
this opinion is orthogonal to my feedback, but in addition to many quality concerns (a lot of the wording here is excessively fluffy imo - llms are well known for that), i think merging an ai-generated rfc as-is sets a potentially antisocial precedent for future collaborators.

PR #220, opened 2026-07-08
Create syntax-nullish-coalescing-assignment.md

consultart-artem:
Proposes adding ??= as a compound assignment operator that assigns only when the left-hand side is nil. Fills a gap in the existing compound assignment operator set (+=, -=, etc.). Addresses the common pattern of default value assignment where the or shorthand is unsafe for boolean values and the correct nil check form requires writing the variable path twice.

### DIFF
--- a/docs/syntax-nullish-coalescing-assignment.md
+++ b/docs/syntax-nullish-coalescing-assignment.md
@@ -1,0 +1,74 @@
+# syntax-nullish-coalescing-assignment
+
+## Summary
+
+Add a `??=` compound assignment operator that assigns to a variable only when its current value is `nil`.
+
+## Motivation
+
+Assigning a default value to a variable only when it is `nil` is one of the most common patterns in Luau: config table initialization, player data defaults, optional argument handling, and lazy caching all use it constantly. There is no concise and correct way to express this today.
+
+The shortest available shorthand is `or`:
+
+```lua
+config.Enabled = config.Enabled or true
+```
+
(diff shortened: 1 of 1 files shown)


docs/syntax-nullish-coalescing-assignment.md
```diff
@@ -0,0 +1,74 @@
...
+config.Enabled = config.Enabled or true
+```
+
+This works when the existing value is always truthy or nil, but it checks truthiness rather than nil-ness. A legitimate `false` value gets silently overwritten, making it unsafe for boolean flags and any other case where `false` is a meaningful value.
+
+The correct form requires an explicit nil check:
+
+```lua
+config.Enabled = config.Enabled == nil and true or config.Enabled
+```
```
aatxe, line 23: in Luau, this would actually be written
```luau
config.Enabled = if config.Enabled == nil then true else config.Enabled
```
which does still repeat `config.Enabled`, of course, but it feels very weird to use a programming pattern we do not recommend for conditional code in an RFC.

gaymeowing:
Is there a reason its `??=` and not `?=`? I would assume because of lookahead, but this isn't stated in the RFC.

Bottersnike:
Will this run into the same problems safe navigation did? That is to say that Roblox userdata error on missing fields, rather than returning nil, making `part.Foo ??= "bar"` error in a surprising way?

To potentially answer my own question, I think it might be more okay than safe navigation ended up, because `part.Foo = "bar"` would error anyway, so the scope for problematic code is drastically reduced.

PR #221, opened 2026-07-08
Add RFC for default parameters and named arguments

Aurora-Aeralis:
[Rendered](https://github.com/Aurora-Aeralis/rfcs/blob/syntax-default-parameters-named-arguments/docs/syntax-default-parameters-named-arguments.md)

## Summary

Proposes default parameter values and named call arguments for Luau functions.
* Includes [prototype implementation](https://github.com/Aurora-Aeralis/luau/tree/aurora/default-params-named-args) and local benchmark results.

### DIFF
--- a/docs/syntax-default-parameters-named-arguments.md
+++ b/docs/syntax-default-parameters-named-arguments.md
@@ -1,0 +1,224 @@
+# Default parameters and named call arguments
+
+## Summary
+
+This RFC proposes default parameter values and named call arguments for Luau functions. Default parameters allow function authors to declare fallback values directly in the parameter list, while named call arguments allow callers to provide specific parameters without manually passing placeholder values for earlier parameters.
+
+## Motivation
+
+Luau code commonly uses manual `nil` checks at the start of functions to provide fallback values:
+
+```luau
+local function createItem(name: string?, amount: number?, visible: boolean?)
+    name = if name == nil then "Item" else name
+    amount = if amount == nil then 1 else amount
+    visible = if visible == nil then true else visible
+
+    -- ...
+end
+```
+
+This pattern is repetitive, makes the real function body start later, and becomes harder to read as the number of optional parameters grows. It also makes calls awkward when a caller only wants to override a later argument:
+
+```luau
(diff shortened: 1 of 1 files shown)


InfraredGodYT:
There's already a pretty good RFC for [default arguments.](https://github.com/luau-lang/rfcs/pull/91)

Also, you should link your markdown with the title "Rendered" in the original comment.

Aurora-Aeralis:
> There's already a pretty good RFC for [default arguments.](https://github.com/luau-lang/rfcs/pull/91)
> 
> Also, you should link your markdown with the title "Rendered" in the original comment.

I implemented the "Rendered" suggestion. Additionally, I wanted to show that it **could be implemented without compromises** other than a **significantly bloated size increase to Luau**. The latest patch I implemented includes the **optimization to make it generally exceed the alternatives** I provided.

MagmaBurnsV:
Feeding type information into the compiler to enable desugaring like this is a huge no for me. Type annotations should never affect runtime behavior barring AOT compilation. People still get confused over how multivals get truncated when you ascribe a type on them, so how do you suppose they'll react when a type inference change breaks the order of their arguments?

Aurora-Aeralis:
This is just for the prototype what is supposed to show it can be done with maintaining old faster interpreter and optional toggle when a value with a default is present. Additionally, I did not want to recode everything, so I had to modify core functionality, also the speed optimizations would not be possible without doing so.

Bottersnike:
Is this mostly just #91 but also adding `foo(a=1)` at call time, and allowing use of argument values in the expressions for the default values of subsequent ones or am I missing any other differences?

vegorov-rbx:
We have already rejected an RFC for named arguments in https://github.com/luau-lang/rfcs/pull/135
This RFC takes the same compromise, no matter if names are resolved through type annotations or through function definition AST, it does not work across modules or for embedder APIs.

The remaining default argument part is already discussed in https://github.com/luau-lang/rfcs/pull/91 and this RFC only provides a small adjustment about previous argument visibility.

I would suggest taking default argument suggestions to #91 and not proceed with #135

PR #222, opened 2026-07-10
RFC: Buffer and string slice syntax

GameF200:
[Rendered](https://github.com/GameF200/rfcs/blob/master/docs/buffer-and-string-slices.md)

### DIFF
--- a/docs/buffer-and-string-slices.md
+++ b/docs/buffer-and-string-slices.md
@@ -20,7 +20,7 @@ local slice = string.sub(str, start, finish)
 
 ```lua
 local slice = buffer.create(size)
-buffer.copy(part, 0, source, offset, size)
+buffer.copy(slice, 0, source, offset, size)
 ```
 
 #### For high-frequency workloads such as:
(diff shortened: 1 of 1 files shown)


deviaze:
Just FYI, unfortunately buffer slices were already rejected: https://github.com/luau-lang/rfcs/pull/163.

I'd be interested in slicing with new range datatype constructed via concatting numbers though. 

```luau
local rang = 1..4
for i in rang do
end

local res = t[rang]
```

GameF200:
> Just FYI, unfortunately buffer slices were already rejected: #163.
> 
> I'd be interested in slicing with new range datatype constructed via concatting numbers though.
> 
> ```luau
> local rang = 1..4
> for i in rang do
> end
> 
> local res = t[rang]
> ```

oh i didnt know about #163. Thanks for the context!

whoamimaybeme:
how would a slice metamethod even work? how would you make it follow the semantics of a slice? within the code example you provide, you just leave the function blank.

GameF200:
> how would a slice metamethod even work? how would you make it follow the semantics of a slice? within the code example you provide, you just leave the function blank.

Oh, thats a tough question. I came up with __slice in the RFC just superficially. I think it's definitely worth reconsidering and standardizing somehow.

PR #223, opened 2026-07-11
Local Table Destructuring

xHeptc:
[rendered](https://github.com/xHeptc/rfcs/blob/master/docs/local-table-destructuring.md)

Note: I am fully aware of RFC 'Granural Imports', 'Table Destructing Syntax' proposals made by other people.
My RFC proposes simpler syntax, doesn't require new keywords, and is not limited to top-level scope with the usage of `require()`.

### DIFF
--- a/docs/improving-type-validation.md
+++ b/docs/improving-type-validation.md
@@ -1,0 +1,131 @@
+# Type Validation and Structural Comparison Utility
+
+## Summary
+
+Introduce a built-in method for `Type` objects in Type Function library (e.g., "Type:verifyFields" or "Type:matches") that validates whether a given object conforms to a target type definition, returning detailed error metadata.
+
+## Motivation
+
+Currently, implementing structural validation in Luau Type Functions requires manually iterating through properties using `Type:properties()`, checking existence via `readproeprty`, and comparing types using `Type:is()`.
+
+As shown in my example, validating a simple Interface requires over 20 lines of repetitive imperative code. This is:
+1. Error-prone: Develoeprs must manually handle edge cases like optional fields and indexers.
+2. Verbose: Common patterns (like ensuring a table matches an interface) are rewritten constantly.
+3. Inconsistent: Different developers will return different error formats, making library interoperability difficult.
+4. Needs a Luau table to mimic a type definiton and it's fields to check against.
+5. Becomes more complex the bigger your type becomes.
+6. Having the type to be an actual type, not a mimic can still be cruical to have exported, it can be bad when you end up having a mimic and a real type.
+
+A built-in utility would allow for "Type-Driven Validation" where the type system itself acts as the schema validator for runtime-adjacent logic.
+
+## Design
+
+I propose adding a method to the `Type` object accessible within type functions.
+
+API Signature
+
+```lua
+type TypeValidatorErrorKind = 'type-mismatch' | 'field-missing' | 'unexpected'
+type ValidationError = {
+    kind: TypeValidatorErrorKind,
+    field: string,
+    targetType: Type,
+}
+
+type ValidationOptions = {
+    strict: boolean
+}
+
+-- Within a type function context:
+function Type:matches(target: table, options: ValidationOptions): (boolean, ValidationError?)
+```
+
+Behavior
+
+**target**: The object to be validated against `self` type
+**options.strict**: If `true`, the presence of fields in `target` that are not defined in `slef` (the schema) will trigger an `unexpected` error. If `false` (default), extra fields are ignored.
+
+Error Mapping
+1. `field-missing`: A property exists in the type definition but is absent in the target.
+2. `type-mismatch` A property exists in both, but `:is()` returns false.
+3. `unexpected`: (Strict mode only) A property exists in the target but not in the definition.
+
+Examples
+
+Current Approach
+```lua
+type function Person(obj)
+    if not obj:is("table") then
+        error("Object must be a table")
+    end
+
+    local properties = obj:properties()
+    
+    local PersonType = {
+        Name = "string"
+    }
+
+    for Field in PersonType do
+        Field = obj:readproperty(types.singleton(Field))
+
+        if not Field then
+            error(`Person is missing {Field}`)
+        end
+    end
+
+    for k, v in properties do
+        local value = v.read
+        local key = k:value()
+
+        local targetType = PersonType[key]
+
+        if targetType and not value:is(targetType) then
+            error(`Person {key} must be {targetType}`)
+        end
+    end
+
+    return obj
+end
+
+local function CreatePerson<T>(person: T): Person<T>
+    return person
+end
+
+local Peter = CreatePerson({
+    Name = true --// Person Name must be string
+})
+```
+
+Proposed Approach
+
+```lua
+local Person = {
+    name = "john"
+}
+
+type PersonType = {
+    Name: string
+}
+
+local Success, Error = PersonType:matches(Person, { strict = false })
+
+if not Success then
+    if Error.kind == Enum.TypeValidatorErrorKind.Mismatch then
+       error(`Person field {Error.field} expected to be {Error.targetType}`)
+    elseif Error.kind = Enum.TypeValidatorErrorKind.FieldMissing then
+        error(`Required field {Error.field} is missing!`)
+    end
+end
+```
+
+## Drawbacks
+
+Must handle recursive types and unions carefullly to avoid infinite loops or performance degradation during analysis.
+
+## Alternatives
+
+Write a runtime object schema validator.
+
+## Unresolved Questions
+
+1. How should this behave with `indexers` (dictionary types), especially with `strict` mode?
--- a/docs/improving-type-validation.md
+++ b/docs/improving-type-validation.md
@@ -6,15 +6,15 @@ Introduce a built-in method for `Type` objects in Type Function library (e.g., "
 
 ## Motivation
 
-Currently, implementing structural validation in Luau Type Functions requires manually iterating through properties using `Type:properties()`, checking existence via `readproeprty`, and comparing types using `Type:is()`.
+Currently, implementing structural validation in Luau Type Functions requires manually iterating through properties using `Type:properties()`, checking existence via `readproperty`, and comparing types using `Type:is()`.
 
 As shown in my example, validating a simple Interface requires over 20 lines of repetitive imperative code. This is:
-1. Error-prone: Develoeprs must manually handle edge cases like optional fields and indexers.
+1. Error-prone: Developers must manually handle edge cases like optional fields and indexers.
 2. Verbose: Common patterns (like ensuring a table matches an interface) are rewritten constantly.
 3. Inconsistent: Different developers will return different error formats, making library interoperability difficult.
 4. Needs a Luau table to mimic a type definiton and it's fields to check against.
 5. Becomes more complex the bigger your type becomes.
-6. Having the type to be an actual type, not a mimic can still be cruical to have exported, it can be bad when you end up having a mimic and a real type.
+6. Having the type to be an actual type, not a mimic can still be crucial to have exported, it can be bad when you end up having a mimic and a real type.
 
 A built-in utility would allow for "Type-Driven Validation" where the type system itself acts as the schema validator for runtime-adjacent logic.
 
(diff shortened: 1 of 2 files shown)


jackhexed:
`const { foo }` is valid syntax today, and while it is possible to parse, its really annoying.

xHeptc:
> `const { foo }` is valid syntax today, and while it is possible to parse, its really annoying.

Is this syntax reserved for future?

deviaze:
Could you put some simpler examples for clarity, and include notes on how this works with array-likes, classes, and extern types on the RHS?

jackhexed:
> > `const { foo }` is valid syntax today, and while it is possible to parse, its really annoying.
> 
> 
> 
> Is this syntax reserved for future? 

No, its a function call.

deviaze:
> > `const { foo }` is valid syntax today, and while it is possible to parse, its really annoying.
> 
> Is this syntax reserved for future?

This is already valid syntax for a table call on a function named `const`

```luau
const const = function(t: { s: string })
    print(t.s)
    -- do smth with t
end

const { s = "meow" } -- this is a function call
```

edit: jack hit post first and don't trust the syntax highlighter (it's wrong)

xHeptc:
> > > `const { foo }` is valid syntax today, and while it is possible to parse, its really annoying.
> > 
> > 
> > Is this syntax reserved for future?
> 
> This is already valid syntax for a table call on a function named `const`
> 
> ```luau
> const const = function(t: { s: string })
>     print(t.s)
>     -- do smth with t
> end
> 
> const { s = "meow" } -- this is a function call
> ```
> 
> edit: jack hit post first and don't trust the syntax highlighter (it's wrong)

What about a special character indicating destructuring?
```luau
local @{ print, warn } = log
const @{ print, warn } = log
```

or, if this is not optimal, restricting this destructuring syntax to local declarations (exclude constants)

Bottersnike:
> Optimal way to distinguish destructuring syntax from function call now is to peek after } token and look for = token.

This isn't possible as outlined, at least not trivially. We could consider `const {foo=bar}` as the simplest case where looking for a `=` token has failed us. Parsing decisions would need to be made then later decide _what_ it was that was actually parsed.

xHeptc:
> > Optimal way to distinguish destructuring syntax from function call now is to peek after } token and look for = token.
> 
> This isn't possible as outlined, at least not trivially. We could consider `const {foo=bar}` as the simplest case where looking for a `=` token has failed us. Parsing decisions would need to be made then later decide _what_ it was that was actually parsed.

Got it. I've been brainstorming for some time and got two somewhat solutions.

1. Fully restrict destructuring to `local` declarations, excluding `const` 

2.  Currently, the order of parsing is `parsePrimaryExpr` and then `parseStat`, which is perfectly normal, what I propose is: pre-check in `parseStat`, before calling `parsePrimaryExpr`, when `lexer.current()` is `Name` with text `const` and `lexer.lookahead()` is `{`, save a lexer checkpoint, do the scan to the matching `}`, peek for `=`, then either restore and parse as a pattern or restore and fall through to the normal `parsePrimaryExpr` path.


First one is much simpler, but you lose the chance of having restrictions against rebinding your keys.

And the second one is much more complex to implement, even with type annotations,

0x57e11a:
> 2.  Currently, the order of parsing is `parsePrimaryExpr` and then `parseStat`, which is perfectly normal, what I propose is: pre-check in `parseStat`, before calling `parsePrimaryExpr`, when `lexer.current()` is `Name` with text `const` and `lexer.lookahead()` is `{`, save a lexer checkpoint, do the scan to the matching `}`, peek for `=`, then either restore and parse as a pattern or restore and fall through to the normal `parsePrimaryExpr` path.

you'll probably want to peek for `,` in addition to `=`, to handle the case of `const { foo = bar }, baz = a, b`, avoiding it being parsed as a primary expr (function call) followed by a syntax error

xHeptc:
> Could you put some simpler examples for clarity, and include notes on how this works with array-likes, classes, and extern types on the RHS?

Sorry for the very late reply.

Basic examples:
```luau
local { x, y } = Point

--// equivelent to
local _temp1 = Point
local x = _temp1.x
local y = _temp1.y
```
```lua
local player, { health, mana } = GetPlayer(), GetStats()
--// equivelent to
local player = GetPlayer()
local _temp1= GetStats()
local health = _temp1.health
local mana = _temp1.mana
```

Array-like
```lua
--// not supported
local { 1, 2 } = list

--// not supported
local { first = 1 } = list

--// if users want array unpacking, they should continue using:
local first, second = list[1], list[2]
-- or table.unpack
```

Classes
```luau
--// NOTE: I havent used classes yet, so I can only give few examples

--// basic class instance destructuring is same as the basic example

--// assuming you have a class
class Point
    public x: number
    public y: number

    function length(self)
 
    end

    function fromAxisLength(...)

    end
end

--// there are three possibilities

--// option 1: 
--// if "obj.field" is valid, then { field } is valid too.

--// option 2: only fields
--// only allow declared fields ("public x", "public y"), rejecting methods.

--// option 3: public members
--// allow both public fields and public methods, but reject private ones once private members exist.

--// this is probably what would naturally happen anyway if the language already prevents accessing private members via ----// obj.member
```

Extern types
```luau
type Config = {
    enabled: boolean,
    timeout: number,
}

local config: Config = GetConfig()

--// the destructured variables inherit the field types
local { enabled, timeout } = config
-- enabled -- boolean
-- timeout -- number
```

xHeptc:
although for array destructuring, I might edit the RFC and propose that too with syntax similar to JS `[] would be for arrays, {} for objects`, or at worse case scenario `[ ] for objects, and arrays not supported`, finally removing ambiguity between sugar function calls", but I need opinions here guys :)

TooManyLimits:
I agree with this idea over the other destructuring RFCs. The most common usecase for destructuring would be setting up shorthand for table accesses at the top of a block, so it’s not a big downside to prevent assignments. Personally I’d prefer the syntax `local x, y in foo()`, because of no noisy curly braces, no confusion with array destructuring (which is unnecessary anyway due to table.unpack), it’s trivial to parse with no backwards compatibility issues, and it fits well with the existing use of the `in` keyword which is to introduce bindings in the same way with for-loops, but I’ll take anything as long as I don’t need to bind a temp variable for the table and list all of the key names twice.

SeloTheGreat:
> I agree with this idea over the other destructuring RFCs. The most common usecase for destructuring would be setting up shorthand for table accesses at the top of a block, so it’s not a big downside to prevent assignments. Personally I’d prefer the syntax `local x, y in foo()`, because of no noisy curly braces, no confusion with array destructuring (which is unnecessary anyway due to table.unpack), it’s trivial to parse with no backwards compatibility issues, and it fits well with the existing use of the `in` keyword which is to introduce bindings in the same way with for-loops, but I’ll take anything as long as I don’t need to bind a temp variable for the table and list all of the key names twice.

This solution sort of feels like the import syntax RFC (https://github.com/luau-lang/rfcs/pull/103) but has a wider applicability to all tables. IMO I actually really like the syntax here, but table destructuring in general means that type imports from modules will just not be a thing. The team could end up both implementing table destructuring and `import` statements, but i'd assume `import` may be more favourable to match with the already existing `export`.

TooManyLimits:
Theres already special logic for when the right hand side of a local declaration is a require() call, to add the types to the current scope prefixed by the LHS. I think a good syntax could be maybe `type a, b, c in require(…)` to match the other one? Though it would be annoying to need to require() separately for the values and types, so maybe a dedicated import is better. Imports would be by *far* the most common usecase for destructuring

xHeptc:
> Theres already special logic for when the right hand side of a local declaration is a require() call, to add the types to the current scope prefixed by the LHS. I think a good syntax could be maybe `type a, b, c in require(…)` to match the other one? Though it would be annoying to need to require() separately for the values and types, so maybe a dedicated import is better. Imports would be by _far_ the most common usecase for destructuring

This RFC specifically states that object destructuring is not limited to / purely focused to `require`.

xHeptc:
> > I agree with this idea over the other destructuring RFCs. The most common usecase for destructuring would be setting up shorthand for table accesses at the top of a block, so it’s not a big downside to prevent assignments. Personally I’d prefer the syntax `local x, y in foo()`, because of no noisy curly braces, no confusion with array destructuring (which is unnecessary anyway due to table.unpack), it’s trivial to parse with no backwards compatibility issues, and it fits well with the existing use of the `in` keyword which is to introduce bindings in the same way with for-loops, but I’ll take anything as long as I don’t need to bind a temp variable for the table and list all of the key names twice.
> 
> This solution sort of feels like the import syntax RFC (#103) but has a wider applicability to all tables. IMO I actually really like the syntax here, but table destructuring in general means that type imports from modules will just not be a thing. The team could end up both implementing table destructuring and `import` statements, but i'd assume `import` may be more favourable to match with the already existing `export`.

```luau
local type { ObjectType }, { DefaultObj } = require() --// local type ","; "," marks it as custom variable
local type, { ObjectType } --// wrong, no types imported.

local { type ObjectType, DefaultObject } = require()
```

theres a few ways of suggesting how types are imported, personaly I think it should be one of these two

xHeptc:
in my opinion, per-binding `type` is the best design, give me a few days and I will commit these new ideas
```luau
local { RuntimeThing, type Config, type Options } = require("./Module")
```

TooManyLimits:
require() is the only circumstance where types are ever seemingly “stored in a table” as far as I know, so putting handling for them in a generic table destructuring construct doesn’t feel right to me. What happens if you try to destructure a type out of a non-require()’d table on the right hand side?

xHeptc:
> require() is the only circumstance where types are ever seemingly “stored in a table” as far as I know, so putting handling for them in a generic table destructuring construct doesn’t feel right to me. What happens if you try to destructure a type out of a non-require()’d table on the right hand side?

you're behaving like we are somehow excluding object destructuring from a require, all im saying is this isnt limited to require usage only, does not mean that type importing wouldnt be a thing, i already suggested two syntaxes above your reply

xHeptc:
>What happens if you try to destructure a type out of a non-require()’d table on the right hand side?

Also I forgot to hit comment lol, anyways, it would behave identical to 
```luau
local object = {}
type Person = object.Person
```
this adds no complications when destructuring types from a table that doesnt contain them
image

InfraredGodYT:
> I agree with this idea over the other destructuring RFCs. The most common usecase for destructuring would be setting up shorthand for table accesses at the top of a block, so it’s not a big downside to prevent assignments. Personally I’d prefer the syntax `local x, y in foo()`, because of no noisy curly braces, no confusion with array destructuring (which is unnecessary anyway due to table.unpack), it’s trivial to parse with no backwards compatibility issues, and it fits well with the existing use of the `in` keyword which is to introduce bindings in the same way with for-loops, but I’ll take anything as long as I don’t need to bind a temp variable for the table and list all of the key names twice.

I agree with this syntax as well. Something along the lines of:

```luau
local point = {x = 1, y = 2, z = 3}
const x, y, z from point -- `in` is fine in place of `from`, I just think `from` is less ambiguous
```

It looks much nicer than using curly brackets and reads well without breaking compatibility.

xHeptc:
> > I agree with this idea over the other destructuring RFCs. The most common usecase for destructuring would be setting up shorthand for table accesses at the top of a block, so it’s not a big downside to prevent assignments. Personally I’d prefer the syntax `local x, y in foo()`, because of no noisy curly braces, no confusion with array destructuring (which is unnecessary anyway due to table.unpack), it’s trivial to parse with no backwards compatibility issues, and it fits well with the existing use of the `in` keyword which is to introduce bindings in the same way with for-loops, but I’ll take anything as long as I don’t need to bind a temp variable for the table and list all of the key names twice.
> 
> I agree with this syntax as well. Something along the lines of:
> 
> ```luau
> local point = {x = 1, y = 2, z = 3}
> const x, y, z from point -- `in` is fine in place of `from`, I just think `from` is less ambiguous
> ```
> 
> It looks much nicer than using curly brackets and reads well without breaking compatibility.

while this syntax looks nice, again, it doesnt really fit luau, it looks like a complete different variable declaration syntax, destructuring is supposed to be very identical to regular variable declaration
```luau
local x, y, z = point.x, point.y, point.z
local { x, y, z } = point
```
this keeps left hand side structure, assignment operator and right hand side, your version replaces `=` with `from` / `in` keyword, which is not ideal for assignment 

not to mention your version doesnt allow extending:
```luau
local foo, bar = 5, 7
-- i can easily extend variables to
local foo, bar, { magic_foo, magic_bar } = 5, 7, magic
```

PR #225, opened 2026-07-12
Document new math.sum function design

HarmosCreations:
[rendered](https://github.com/HarmosCreations/rfcs_harmos/blob/patch-1/function-math-sum.md)

I would like to request for a new function inside the math library
math.sum a fundemental function who adds numbers together perfect
for unknown vardiac arguments and simplifies other user-defined functions like [math.avg](https://github.com/luau-lang/rfcs/pull/226)
More info inside the md file.

### DIFF
--- a/function-math-sum.md
+++ b/function-math-sum.md
@@ -1,0 +1,45 @@
+# math.sum
+
+## Summary
+
+``` math.sum(...: number): number ```
+
+Add a ```sum``` function to the ```math``` library, which adds the numbers together.
+
+## Motivation
+
+Numerical addition is one of the most fundamental mathematical operations. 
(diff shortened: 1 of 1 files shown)


InfraredGodYT:
No real reason to use this over the `+` operator.

HarmosCreations:
> No real reason to use this over the `+` operator.

I had many reasons, but to keep it short: It is a function not an operator and operators act on variables not variadic.

plus it will simplify other mathimatical functions like [math.avg](https://github.com/luau-lang/rfcs/pull/226)

aatxe:
`math.avg` would not be made meaningfully different to implement for having `math.sum` as a builtin.

PR #226, opened 2026-07-12
new function math.avg

HarmosCreations:
[rendered](https://github.com/HarmosCreations/rfcs_harmos/blob/master/docs/feature_math.avg.md)

I would like to request for a new function inside the math library math.avg takes in vardiac of numbers and return the average of them.

This request can be compensated for if [math.sum](https://github.com/luau-lang/rfcs/pull/225) is implemented.

More info inside the md file.

### DIFF
--- a/function-math-avg.md
+++ b/function-math-avg.md
@@ -1,9 +1,16 @@
-# math.avg
+# Feature name
 
 ## Summary
 
 Fundemental function that takes in vardiact numbers and returns
-the avarage of them all.
+the mean.
+
+Averaging a sequence of numbers also called arithmetic average
+where the sum is devided by count.
+
+```luau 
+type avg = (...:number) -> number
+```
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


function-math-avg.md
```diff
@@ -0,0 +1,66 @@
...
+
+print(avg) --> 50
+
+```
+
+## Design
+
+The speciality about this function is the arguments,
+it only takes one vardiac argument and automatically
+calculates the average of it.
```
SPY, line 82: what if it is called without arguments? exception? 0?

razvanpacku: Should throw an exception imo, so that it has similar behaviour with the other variadic math functions, min and max.

razvanpacku:
Any news on this?

HarmosCreations:
> Any news on this?

not yet with class sytax and new luau syntax they are quite busy

function-math-avg.md
nnullcolumn: i would think it should return 0 personally (likely to be used with table.unpack, exception big footgun) but additionally it should be typed `(number, ...number)` if it is gonna throw an exception

nnullcolumn:
in the future please include a rendered link, you can see examples of them in other rfcs

function-math-avg.md
```diff
@@ -0,0 +1,66 @@
...
+  total += num
+ end
+
+ return total/#{...}
+end
+
+print(math.avg(1,2,4,6)) --> 3.25
+print(math.avg(table.unpack({5, 5, 3, 2})) --> 3.75
+print(math.avg(string.byte("123", 1, 3)) --> 50
+
```
nnullcolumn, line 100: we should probably address potential accuracy gains - a naïve implementation like this will encounter floating point precision error, and a built-in library function has to
A: address that in motivation
B: decide how it should be implemented in the design

function-math-avg.md
```diff
@@ -0,0 +1,66 @@
...
+## Motivation
+
+Averaging numbers is a common operation in game development
+and to avoid user-defined repetition code like this:
+
+```lua
+
+local total= 0
+
+for _, num in {string.byte("123", 1, 3)} do
```
nnullcolumn, line 24: we could come up with a better example use case than averaging the bytes present in a string

function-math-avg.md
```diff
@@ -0,0 +1,66 @@
...
+
+```
+
+## Drawbacks
+
+This adds another function to the standard library.
+
+## Alternatives
+
+Not implementing this will result in user-defined alternatives and This would keep the same problems as before.
```
Kampfkarren: What problem? You barely if at all describe a problem in the RFC body. The problem appears to be that it's code you write.

HarmosCreations: Yes thanks, thats the only problem i can think of.
I changed the line.

function-math-avg.md
HarmosCreations: Good point but i can promise i had to use that before anyways i commited the new changes.

is it enough for text based examples or should i replace the string.bytes example.

function-math-avg.md
HarmosCreations: Hard request so i had to google some methods and wrote them down in the commited file and in my opinion a balanced one should be sufficient.

But those optimizations are mostely in c++ and i don't know much about c++ so i just added something for the motivation tab.

function-math-avg.md
HarmosCreations: I have wrote down this concern and analyzed the possible outcomes so it should be more clear how the function should act under unexpected moments.

function-math-avg.md
razvanpacku: A Kahan/Neumaier summation could be used for accuracy gains.

nnullcolumn:
i think it should be called `math.average` and also we should get `math.median` :3

PR #229, opened 2026-07-16
Classes

andyfriesen:
[Rendered](https://github.com/luau-lang/rfcs/blob/classes-v2/docs/syntax-classes.md)

We jumped the gun on merging this RFC.  Some light edits are necessary.

Most notably, creating new class instances is now always done via a static `.new()` method.

Class constructors are described separately at https://github.com/luau-lang/rfcs/pull/210
Inheritance is described at https://github.com/luau-lang/rfcs/pull/230

PR #230, opened 2026-07-16
Class inheritance

andyfriesen:
[Rendered](https://github.com/luau-lang/rfcs/blob/class-inheritance/docs/class-inheritance.md)

See https://github.com/luau-lang/rfcs/pull/229 for the main discussion thread on classes as a whole.

### DIFF
--- a/docs/class-inheritance.md
+++ b/docs/class-inheritance.md
@@ -1,0 +1,233 @@
+# Implementation Inheritance for Classes
+
+## Summary
+
+```luau
+class Animal
+	public species: string
+	
+	function __tostring(self)
+		return "I am an animal."
+	end
+	
+	function live(self)
+		return "I am alive"
+	end
+end
+
+class Cat extends Animal
+	public meowMult: number
+	
+	function __tostring(self): string
+		return `{Animal.__tostring(self)} I am a cat.`
+	end
+	
+	function meow(self): string
+		return string.rep("meow\n", self.meowMult)
+	end
+end
+
+local dog = Animal { species = "Canis familiaris" }
+local cat = Cat { species = "Felis catus", meowMult = 3 }
+
+```
+
+## Motivation
+
+Other languages with support for object-oriented programming support inheritance in some way. We (the Luau team) have even written libraries which would benefit from having classes with nominal inheritance. Inheritance also makes it easier to write more modular code and provides another guidepost for programmers coming from other languages. Additionally, most Lua and Luau libraries which provide object oriented programming affordances make implementation inheritance available in some way.
+
+## Design
+
+### Syntax
+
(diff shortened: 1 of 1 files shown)


Bottersnike:
~~Why is field redefinition with subtypes disallowed but methods are allowed to be redefined with a subtype?~~ nvm I can see why it would cause issues depending on where it's accessed from

deviaze:
I like that inheritable classes must be prefixed by `open`. It's a great indicator of when a class is *supposed* to be inherited upon, makes inheritance opt in instead of a default behavior, and prevents library authors from inadvertently adding a prop to a class that wasn't meant to be inherited and breaking downstream users who extend it.

TenebrisNoctua:
This looks pretty good, except that I think Java's step of using `final` is better than `open`. An opt-out is better IMO, where I can extend any class, except the ones the developer does not want, versus I cannot extend any class except the ones the developer specifically goes out of their way to allow for inheritance, which is a thought that doesn't occur most of the time, even when the developer isn't against inheritance of their classes.

That aside, the reasoning for `super` is quite strong especially within the presence of inheriting constructors. It should be re-considered.

docs/class-inheritance.md
```diff
@@ -0,0 +1,279 @@
...
+c.foo() // OK
+```
+
+### Runtime
+
+#### Declaration Order
+
+The original class RFC states that class names are hoisted and so can be used before the class declaration has been evaluated.  This is unchanged by this RFC, but carries with it a consequence that is important to call out:
+
+Class declarations have effects at the top level.  Therefore, a class cannot inherit from another class that occurs lexically after it within the module.
```
MagmaBurnsV, line 129: Again, going to mention this. Implicit hoisting is a hack and doesn't even work for 90% of the cases it would be useful for. Get rid of it.

docs/class-inheritance.md
```diff
@@ -0,0 +1,279 @@
+# Implementation Inheritance for Classes
+
+## Summary
+
+```luau
+open class Animal
```
alexmccord, line 6: Orthogonal to this RFC, but could we also have syntax for abstract methods? Using `error("override me")` as the implementation in the base class won't scale since that can error at runtime, but an abstract method can error during `class` prototyping.

Bottersnike: It's worth potentially also considering interfaces, and deciding on abstract vs interfaces vs both (both can be quite useful to have, but also potentially _too_ much stuff going on depending on the desired design direction).

TenebrisNoctua: an `abstract` keyword seems to be the best option personally.

karl-police:
``public class`` ... but instead ``open class`` o7

skberkeley:
Closing in favor of #242. Future discussions on inheritance will take place there!

PR #232, opened 2026-07-18
Enums

Bottersnike:
[Rendered](https://github.com/Bottersnike/luau-rfcs/blob/enums/docs/enums.md)

---

This is very much still in-draft, but I figured I'd put it out there to get feedback sooner rather than later :)

Some comments regarding Roblox-specific details, omitted from the RFC body due to being Roblox-specific, but worth noting:

---

The "efficient serialisation" discussed in this RFC is focused primarily on the ability for enums to function properly when using them on Roblox in places such as a `RemoteFunction`.

The restrictions presented here allow for static analysis of a workspace to identify and tag all present enums (such as during publishing). Each enum can then be globally represented across both client and server code by means of its unique tag. This can then be used for an incredibly efficient network packing.

We can consider an over-the-wire format such as:

`0 [20 bit: tag] [11 bit: index in enum]` (32-bit compact format)

`1 [20 bit: tag] [43 bit: index in enum]` (64-bit extended format)

With a packing such as this, we can now efficiently support up to 1,048,576 unique enums across a codebase with up to 8,796,093,022,208 members per enum (a ridiculous amount). The extended format could also be restricted to only 48 bits in total, which still permits up to 134,217,728 members per enum. The compact format still permits 2048 members per enum.

One could imagine other over-the-wire packings using a similar approach that greatly benefit from the restrictions imposed in this RFC.

---

Roblox already exposes the `Enum` and `EnumItem` datatypes. Whether these existing engine datatypes are unified with the `enum` and `enumitem` primitives proposed here is intentionally outside the scope of this RFC and would be an embedder decision.

However, the runtime representation should ideally avoid unnecessarily preventing such unification. In particular, the implementation should consider allowing embedders to associate metatables or otherwise provide embedder-specific behaviour for enum objects and enum items. The `vector` primitive vs Roblox's `Vector3` comes to mind as an example of why this consideration is required during initial implementation.

This RFC does not require custom metatables to be exposed to Luau programs, nor does it require Roblox to unify the two families. It only recommends avoiding implementation constraints that would make future integration impossible.

### DIFF
--- a/docs/enums.md
+++ b/docs/enums.md
@@ -1,0 +1,219 @@
+## Summary
+
+This RFC introduces first-class enumeration datatypes into Luau, both at runtime and in the type system.
+
+## Motivation
+
+Luau currently supports basic enum-like fields in the type system by means of string singleton unions (`"foo" | "bar" | "buzz"`). While useful in many cases, this functionality cannot be used to express enumerations that have meaningful values associated with the enumeration. It also faces challenges regarding singleton to `string` demotion.
+
+At current, it can be very challenging to express an enumeration that has both names and values in Luau in a way that is both type-safe and ergonomic to use.
+
+First-class enumerations also present an opportunity for more efficient serialisation of some data structures in environments where this is desired.
+
+## Design
+
+A new contextual keyword `enum` is introduced, which is used as shown:
+
+```luau
+enum DemonstrationEnum
+	Foo = 1
+	Bar = 2
+end
+```
+
+Type annotations can optionally be specified on an enum item, annotating the value of that item:
+
+```luau
+enum DemonstrationEnum
+	Foo: number = 1
+	Bar: number = 2
+end
+```
+
+Not all members of an enum must share the same type. That is to say, the following code is legal:
+
+```luau
+enum DemonstrationEnum
+	Foo = 1
+	Bar = "two"
+end
+```
+
+Further, multiple names may share the same value:
+
+```luau
+enum DemonstrationEnum
+	Foo1 = 1
+	Foo2 = 1
+end
+```
+
+All names must be unique within a single enum. The following code is illegal:
+
+```luau
+enum DemonstrationEnum
+	Foo = 1
+	Foo = 2  -- syntax error: Duplicate enum name "Foo" (see previous definition on line 2)
+end
+```
+
+An enum requires a minimum of one item. An enum with no items would have an  type of `enum<never>` and produce a type alias of `never`, which is meaningless.
+
+```luau
+enum DemonstrationEnum  -- syntax error: No enum items defined
+```
+
+An enum can be exported using `export enum Name`. Two scripts defining an enum with identical names define two distinct enums.
+
+
+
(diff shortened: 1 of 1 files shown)


InfraredGodYT:
I would love if user-defined Enums were supported, but [Roblox also has Enums.](https://create.roblox.com/docs/en-us/reference/engine/datatypes/Enums) Would this pose any problems for this feature being used for Roblox development and can they be solved or avoided?

deviaze:
This would be useful as an alternative/sugar for unions of classes, but I highly dislike the restrictions on data types. These should be like Rust enums instead, where variants have a name and carry data, paving the way for nominal pattern matching:

```luau
enum HttpResult
    variant Success: {
        code: number,
        headers: { [string]: string },
        body: string,
    }
    variant Error: {
        code: number,
        headers: { [string]: string },
        body: string,
    }
    variant IoError: string
    variant Timeout: {
        which: "Global" | "Per-Call" | "RecvBody" | "RecvHeaders",
        configured: Duration,
    }
    variant Other
    function is_ok(self)
        if enum.isvariant(self, HttpResult.Success)
            return true
        end
        return false
    end
end
```

Alternatively, each variant could be a class instead of holding a structural data primitive whose shape can differ at runtime.

Bottersnike:
What I was going for here was "enum" in the sense of "enum of constant values" akin to what you'll find in TypeScript, Python, Java, C#, etc.

What you're proposing there feels like it could be represented just as well using a union of class types. If the nominal nature isn't required, it's already entirely possible using a union of table types.

The data-type restrictions are part of the "constant value" semantics. The two exceptions to that are `nil` and `nan`. `nil` is excluded because people are used to `{foo=nil} == {}`, though could potentially be allowed. `nan` is excluded because equality is always false so usages like `fromValue` would cease to function.

Bottersnike:
> I would love if user-defined Enums were supported, but [Roblox also has Enums.](https://create.roblox.com/docs/en-us/reference/engine/datatypes/Enums) Would this pose any problems for this feature being used for Roblox development and can they be solved or avoided?

I've just added a note in the top comment of the PR regarding this (as Roblox-specific semantics don't feel like they fit in the RFC body). Provided the internal implementation of `enum` and `enumitem` allows for metatables and embedder-modification, I can't see a particular reason why the Roblox types could not be defined in terms of the language primitives. I couldn't say if Roblox would _want_ to do that, but I think if implemented right it could become a "well why not?" sort of thing; `vector`/`Vector3` is of course a cautionary tale of why the implementation really matters for that though.

SeloTheGreat:
One of the nice things for enums would be automatic enumeration, i.e.

```luau
enum Thingies
   Thingy --> 1
   Thingamajig --> 2
   Thingamabob --> 3
end
```

It would be cumbersome for many if all you wanted from an enum was _just_ a unique identifier, but you had to specify the value of each identity even if you didn't want to use their values; i believe automatic enumeration would give the ergonomics to fix this issue.

We could tighten down the semantics to disallow intermixing both manual and automatic enumerations. (syntax error tbd)

```luau
enum Thingies
   Thingy = 5
   Thingamajig -- causes a syntax error, manual and automatic assignment being intermixed is impermissible
   Thingamabob
end
```

This would most likely be necessary because it's not really possible to continue string enumerations in a logical way unlike number enumerations.

Something like Go's `iota` identifier could also be considered, but i'm not sure how i'd go about doing it in luau.

PR #234, opened 2026-07-24
buffer.compare

Elentium:
[Rendered](https://github.com/Elentium/rfcs/blob/buffer-comparison/docs/buffer-comparison.md)

This PR proposes adding `buffer.compare(b1, offset1, b2, offset2?, count?)` to the standard buffer library. It performs an in-place, three-way lexicographic comparison over the underlying memory of two buffers, mimicking C `memcmp`. It returns `0` if the buffers are equal, a negative integer if `b1` is less than `b2`, and a positive integer if `b1` is greater.

### DIFF
--- a/docs/buffer-comparison.md
+++ b/docs/buffer-comparison.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-This RFC proposes extending the Luau buffer library to support comparing the byte sequences of two buffers. The primary API is `buffer.compare(b1, b2)`, which performs a three-way lexicographic comparison (similar to C `memcmp`) and returns an integer: `0` if the sequences are equal, a negative value if `b1` is less than `b2`, and a positive value if `b1` is greater.
+This RFC proposes extending the Luau buffer library to compare byte sequences in two buffers, including optional sub-regions. The primary API is `buffer.compare(b1, b1_offset, b2, b2_offset?, count?)`, which performs a three-way lexicographic comparison (like C `memcmp` on a chosen window) and returns an integer: `0` if the compared regions are equal, a negative value if the `b1` region is less than the `b2` region, and a positive value if it is greater.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


alicesaidhi:
I think it'd be nice if we can also specify the offsets and length of the parts we want to compare

Bottersnike:
I think the "Ranged comparison support" alternative is a much more powerful option, and with the suggested parameter list would be able to gracefully fall back anyway. I suspect `(b1, offset1, b2, offset2, length)` might be more consistent with the rest of the buffer library though.

Elentium:
> I think the "Ranged comparison support" alternative is a much more powerful option, and with the suggested parameter list would be able to gracefully fall back anyway. I suspect `(b1, offset1, b2, offset2, length)` might be more consistent with the rest of the buffer library though.

Yeah, though imo `(b1, b2, offset1?, offset2?, count?)` would be more convenient in cases where you need full 0 to end comparison and you can just pass the 2 buffers

InfraredGodYT:
> > I think the "Ranged comparison support" alternative is a much more powerful option, and with the suggested parameter list would be able to gracefully fall back anyway. I suspect `(b1, offset1, b2, offset2, length)` might be more consistent with the rest of the buffer library though.
> 
> Yeah, though imo `(b1, b2, offset1?, offset2?, count?)` would be more convenient in cases where you need full 0 to end comparison and you can just pass the 2 buffers

Considering the existence of `buffer.copy(target, targetOffset, source, sourceOffset, count)`, I also think `buffer.compare(b1, offset1, b2, offset2, length)` makes more sense to add.

Elentium:
Alright that makes sense, I will update the rfc tomorrow

itsmath:
Sort of unrelated, but can we get bitwise ops for buffers too (maybe implemented with SIMD)? A lot of the use cases I have for comparing buffers also benefit from bitwise ops. Just as a quick example, bitflags with more than 32 bits, currently you have to do a Luau loop and do it in chunks of u32

InfraredGodYT:
> Sort of unrelated, but can we get bitwise ops for buffers too (maybe implemented with SIMD)? A lot of the use cases I have for comparing buffers also benefit from bitwise ops. Just as a quick example, bitflags with more than 32 bits, currently you have to do a Luau loop and do it in chunks of u32

That would need to be a separate RFC entirely.

Elentium:
> Sort of unrelated, but can we get bitwise ops for buffers too (maybe implemented with SIMD)? A lot of the use cases I have for comparing buffers also benefit from bitwise ops. Just as a quick example, bitflags with more than 32 bits, currently you have to do a Luau loop and do it in chunks of u32

That would be a game changing feature, Im interested to see your rfc on it if you decide to make one

Elentium:
Alright updated it a bit

nnullcolumn:
> > > I think the "Ranged comparison support" alternative is a much more powerful option, and with the suggested parameter list would be able to gracefully fall back anyway. I suspect `(b1, offset1, b2, offset2, length)` might be more consistent with the rest of the buffer library though.
> > 
> > 
> > Yeah, though imo `(b1, b2, offset1?, offset2?, count?)` would be more convenient in cases where you need full 0 to end comparison and you can just pass the 2 buffers
> 
> Considering the existence of `buffer.copy(target, targetOffset, source, sourceOffset, count)`, I also think `buffer.compare(b1, offset1, b2, offset2, length)` makes more sense to add.

i don't think the api needs to be designed around `buffer.copy`, since that function requires the first three arguments anyways. i think it should be `(b1, b2, offset1?, offset2?, count?)`.

PR #235, opened 2026-07-26
Amend Proposed Classes

Jediweirdo:
[Rendered](https://github.com/Jediweirdo/rfcs/blob/Jediweirdo-amended-classes/docs/amended-classes.md). Note that if this ever leaves a draft state (which it may not), this will likely be split into multiple PRs. For now, I think it would be easier to dump all of my thoughts on classes in 1 place. Let me know if you would prefer breaking up some of these ideas in the future

### DIFF
--- a/docs/amended-classes.md
+++ b/docs/amended-classes.md
@@ -1,0 +1,464 @@
+# Amended Classes (Draft)
+
+## Summary
+
+```luau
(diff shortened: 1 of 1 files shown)


Jediweirdo:
Closed, as the team is already moving forward with other/better alternatives

PR #236, opened 2026-07-29
table.swapremove

zermelon:
[Rendered](https://github.com/zermelon/rfcs/blob/table-swapremove-rfc/docs/table-swapremove.md)
This RFC proposes a new table function, `table.swapremove`, that implements swap-and-pop for an array, with a similar signature to the existing `table.remove` function.

A [previous RFC](https://github.com/luau-lang/rfcs/pull/119) was made for this request (though it was dubbed as "uremove" rather than "swapremove", and did not return the removed element), but it was not well motivated. This RFC covers various more substantive motivations for this function.

### DIFF
--- a/docs/table-swapremove.md
+++ b/docs/table-swapremove.md
@@ -20,7 +20,7 @@ t[i], t[#t] = t[#t], nil
 ```
 While this idiom is concise, it obscures the intent of performing an unordered removal and leaves the details of the operation to individual implementations. For example, a manual implementation does not return the removed value like `table.remove`, does not consistently validate (namely, throwing an error for non-numeric indices) or floor (for non-integers) the provided index, and may not update the VM's cached array length in the same way as the `table` library. A native implementation could improve performance.
 
-A dedicated `table.swapremove` function would provide a single, well-defined abstraction for this common operation, offering behavior that is consistent with `table.remove`. It would also complete the table library by providing an unordered counterpart to `table.remove`. Users may implement swap-and-pop themselves, just as they might implement insertion with `t[i] = value` or append with `t[#t + 1]` = value. The value of `table.insert`, however, is not that it enables something previously impossible or impractical, but that it provides a clear, consistent, and optimised implementation of a common operation. This RFC proposes doing the same for unordered removal.
+A dedicated `table.swapremove` function would provide a single, well-defined abstraction for this common operation, offering behavior that is consistent with `table.remove`. It would also complete the table library by providing an unordered counterpart to `table.remove`. Users may implement swap-and-pop themselves, just as they might implement insertion with `t[i] = value` or append with `t[#t + 1] = value`. The value of `table.insert`, however, is not that it enables something previously impossible or impractical, but that it provides a clear, consistent, and optimised implementation of a common operation. This RFC proposes doing the same for unordered removal.
 
 Rust (`Vec`'s `swap_remove`), Zig (`ArrayList`'s `swapRemove`), and Unreal Engine (`TArray::RemoveAtSwap`) implement this.
  
@@ -37,7 +37,7 @@ table.swapremove(t: {any}, pos: number): Variant
 - Uses `pos = #t` if `pos` is omitted
 - Returns the removed element
 
-We propose using Rust and Zig's `swapremove` nomenclature (in luaucase), as opposed to Unreal Engine's `RemoveAtSwap`, or something else like `swapandpop` or `moveandpop`.
+We propose using Rust and Zig's `swapremove` nomenclature (in luaucase), as opposed to something like `swapandpop` or `moveandpop`, or Unreal Engine's `RemoveAtSwap`.
 
 ## Drawbacks
 
--- a/docs/table-swapremove.md
+++ b/docs/table-swapremove.md
@@ -18,7 +18,7 @@ Currently, to perform swap and pop, one might usually do the following:
 ```luau
 t[i], t[#t] = t[#t], nil
 ```
-This obscures the intent of performing an unordered removal and leaves the details of the operation to individual implementations. For example, a manual implementation does not return the removed value like `table.remove`, does not consistently validate (namely, throwing an error for non-numeric indices) or floor (for non-integers) the provided index, and it is certainly less clear than a dedicated `table.swapremove`. It also has unexpected behaviour for negative indices or indices that are greater than maxn, namely converting all the keys into string keys and turning the entire array sparse respectively.
+This obscures the intent of performing an unordered removal and leaves the details of the operation to individual implementations. For example, a manual implementation does not return the removed value like `table.remove`, does not consistently validate (namely, throwing an error for non-numeric indices) or floor (for non-integers) the provided index, and it is certainly less clear than a dedicated `table.swapremove`. It also has unexpected behaviour for negative indices or indices that are greater than maxn, namely turning it into a mixed table without any validation, and turning the entire array sparse respectively.
 
 A dedicated `table.swapremove` function would provide a single, well-defined abstraction for this common operation, offering behavior that is consistent with `table.remove`. It would also complete the table library by providing an unordered counterpart to `table.remove`. 
 
@@ -46,12 +46,12 @@ Swapping the last element with element 99 is already a major issue. Firstly, if 
 
 The second part, removing the last element, is also now at best ambiguous. It is unclear whether we remove the original last element, at index 4, or the new "last" element, at index 99. Using the former is what `table.remove` would do on this mutated array, but again that still leaves holes in the array which we did not have before. Using the latter would trivially have made the swapremove operation a no-op. There is no sensible way to swap-and-pop in this way.
 
-Caling `table.swapremove` with a non-positive index should be an error. Consider the following swap part of the process:
-```luau
+Calling table.swapremove with a non-positive index should be an error. Consider the following:
+
 local myFavouriteFruits = {"apple", "banana", "kiwi", "orange"}
 myFavouriteFruits[-4] = myFavouriteFruits[#myFavouriteFruits]
-```
-This would convert all the numeric keys in the array into string keys which is entirely unexpected. Trying to circumvent this by keeping all other keys numeric and turning -4 into a string key (which is unnecessary complexity), and then somehow continuing from there on a mixed table, is inconsistent with how we (and table.remove) handled other unacceptable indices like strings. It is safer and clearer to keep this as an error.
+
+This introduces a negative key into what was previously a contiguous array, turning it into a mixed table. This is an unexpected result for an operation whose purpose is simply to remove an element from an array. It is therefore safer and clearer to reject non-positive indices.
 
 ## Drawbacks
 
--- a/docs/table-swapremove.md
+++ b/docs/table-swapremove.md
@@ -22,7 +22,7 @@ While this idiom is concise, it obscures the intent of performing an unordered r
 
 A dedicated `table.swapremove` function would provide a single, well-defined abstraction for this common operation, offering behavior that is consistent with `table.remove`. It would also complete the table library by providing an unordered counterpart to `table.remove`. Users may implement swap-and-pop themselves, just as they might implement insertion with `t[i] = value` or append with `t[#t + 1]` = value. The value of `table.insert`, however, is not that it enables something previously impossible or impractical, but that it provides a clear, consistent, and optimised implementation of a common operation. This RFC proposes doing the same for unordered removal.
 
-Rust (`Vec`'s `swap_remove`), Zig (`ArrayList`'s `swapRemove`), and Unreal Engine (`TArray::RemoveAtSwap`), implement this.
+Rust (`Vec`'s `swap_remove`), Zig (`ArrayList`'s `swapRemove`), and Unreal Engine (`TArray::RemoveAtSwap`) implement this.
  
 ## Design
 
--- a/docs/table-swapremove.md
+++ b/docs/table-swapremove.md
@@ -37,14 +37,14 @@ table.swapremove<T>(t: {T}, pos: number): T?
 
 For arrays with holes, calling `table.swapremove` on an index with a nil value will not silently fail, and will as prescribed move the last element to this index and return nil correctly, as it is the removed element.
 
-Calling `table.swapremove` beyond the maximum numerical index of the array, or on an index below 1, should do nothing. This may seem odd given we do not do this for holes in the array, but an alternative would be to continue as normal, which has undesirable effects. Consider the following:
+Calling `table.swapremove` beyond the maximum numerical index of the array, should do nothing. This may seem odd given we do not do this for holes in the array, but an alternative would be to continue as normal, which has undesirable effects. Consider the following:
 ```luau
 local myFavouriteFruits = {"apple", "banana", "kiwi", "orange"}
 table.swapremove(myFavouriteFruits, 99)
 ```
-Swapping the last element with element 99 is already a major issue. Firstly, if we naively did `myFavouriteFruits[99] = myFavouriteFruits[4]`, this results in the table becoming `{[1]: "apple", [2]: "banana", [3]: "kiwi", [4]: "orange", [99]: "orange"}`. This introduces a large gap in what was previously a contiguous sequence, and turns it into a sparse array. Creating holes in this manner is undesirable and gives perhaps classically unexpected results, including #t (4) and table.remove (removes `[4]: "orange"`).
+Swapping the last element with element 99 is already a major issue. Firstly, if we naively did `myFavouriteFruits[99] = myFavouriteFruits[4]`, this results in the table becoming `{[1]: "apple", [2]: "banana", [3]: "kiwi", [4]: "orange", [99]: "orange"}`. This introduces a gap in what was previously a contiguous sequence, and turns it into a sparse array. Creating holes in this manner is undesirable and gives perhaps classically unexpected results, such as #t giving 4 and table.remove with no position argument  removing `[4]: "orange"`.
 
-The second part, removing the last element, is also now at best ambiguous. It is unclear whether we remove the original last element, at index 4, or the new "last" element, at index 99. Using the former is what `table.remove` would do on this mutated array, but again that still leaves holes in the array which we did not have before. Using the latter would trivially have made the swapremove operation a no-op. There is no sensible way to swap-and-pop in this way.
+The second part, removing the last element, is also now at best ambiguous. It is unclear whether we remove the original last element, at index 4, or the new "last" element, at index 99. Using the former is what `table.remove` would do on this mutated array, but again that still leaves holes in the array which we did not have before. Using the latter would trivially have made the swapremove operation a no-op. There is no sensible way to swap-and-pop here.
 
 Calling table.swapremove with a non-positive index should be an error. Consider the following:
 
--- a/docs/table-swapremove.md
+++ b/docs/table-swapremove.md
@@ -1,0 +1,49 @@
+# `table.swapremove`
+
+## Summary
+
+This RFC introduces a new `table` library function, `table.swapremove(t, i)`, that implements the swap-and-pop algorithm for an array.
+
+## Motivation
+
+Swap-and-pop is a very common algorithm used to remove an element from an array in O(1) by replacing said element with the final element, and then removing that last element. This is significantly faster, especially for larger arrays, than a regular `table.remove`-like operation, which must shift all subsequent elements to preserve a contiguous layout. It is used frequently in places like ECS implementations, particle systems, object pooling, and plenty more. A brief demonstration is shown below:
+
+```lua
+local myFavouriteFruits = {"apple", "banana", "kiwi", "orange"}
+table.swapremove(myFavouriteFruits, 2)
+print(myFavouriteFruits) -- {"apple", "orange", "kiwi"}
+```
+
+Currently, to perform swap and pop, one might usually do the following:
+```lua
+t[i], t[#t] = t[#t], nil
+```
+While this idiom is concise, it obscures the intent of performing an unordered removal and leaves the details of the operation to individual implementations. For example, a manual implementation does not return the removed value like `table.remove`, does not consistently validate (namely, throwing an error for non-numeric indices) or floor (for non-integers) the provided index, and may not update the VM's cached array length in the same way as the `table` library. A native implementation could improve performance.
+
+A dedicated `table.swapremove` function would provide a single, well-defined abstraction for this common operation, offering behavior that is consistent with `table.remove`. It would also complete the table library by providing an unordered counterpart to `table.remove`. Users may implement swap-and-pop themselves, just as they might implement insertion with `t[i] = value` or append with `t[#t + 1]` = value. The value of `table.insert`, however, is not that it enables something previously impossible or impractical, but that it provides a clear, consistent, and optimised implementation of a common operation. This RFC proposes doing the same for unordered removal.
+
+Rust (`Vec`'s `swap_remove`), Zig (`ArrayList`'s `swapRemove`), and Unreal Engine (`TArray::RemoveAtSwap`), implement this.
+ 
+## Design
+
+The proposed signature would be very similar to that of the existing `table.remove` (using the style of the [current documentation](https://create.roblox.com/docs/reference/engine/libraries/table#remove)):
+```lua
+table.swapremove(t: {any}, pos: number): Variant
+```
+
+`table.swapremove` behaves similarly to `table.remove`. Namely:
+- Updates the VM's cached value for #t
+- Does nothing (silently) on integer indices which have a `nil` value, floors non-integer indices, and throws errors for non-numeric indices
+- Uses `pos = #t` if `pos` is omitted
+- Returns the removed element
+
+We propose using Rust and Zig's `swapremove` nomenclature (in luaucase), as opposed to Unreal Engine's `RemoveAtSwap`, or something else like `swapandpop` or `moveandpop`.
+
+## Drawbacks
+
+It is yet another function to the table library that has a relatively simple implementation, like `table.insert` or `table.remove`, that users could reasonably implement themselves as discussed earlier.
+It is not immediately apparent to an unfamiliar user whether or not `swapremove` would return the swapped element or the removed element.
+
+## Alternatives
+
+We could simply not implement this function. Users could continue to use the swap assignment.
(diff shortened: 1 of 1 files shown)


docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+
+A dedicated `table.swapremove` function would provide a single, well-defined abstraction for this common operation, offering behavior that is consistent with `table.remove`. It would also complete the table library by providing an unordered counterpart to `table.remove`. Users may implement swap-and-pop themselves, just as they might implement insertion with `t[i] = value` or append with `t[#t + 1] = value`. The value of `table.insert`, however, is not that it enables something previously impossible or impractical, but that it provides a clear, consistent, and optimised implementation of a common operation. This RFC proposes doing the same for unordered removal.
+
+Rust (`Vec`'s `swap_remove`), Zig (`ArrayList`'s `swapRemove`), and Unreal Engine (`TArray::RemoveAtSwap`) implement this.
+ 
+## Design
+
+The proposed signature would be very similar to that of the existing `table.remove` (using the style of the [current documentation](https://create.roblox.com/docs/reference/engine/libraries/table#remove)):
```
aatxe: > using the style of the [current documentation](https://create.roblox.com/docs/reference/engine/libraries/table#remove))

This is not Luau documentation, and is not a Luau type signature. It's not good documentation for Roblox to be producing since it is fundamentally incorrect, and it definitely does not belong in our RFCs.

docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+Swap-and-pop is a very common algorithm used to remove an element from an array in O(1) by replacing said element with the final element, and then removing that last element. This is significantly faster, especially for larger arrays, than a regular `table.remove`-like operation, which must shift all subsequent elements to preserve a contiguous layout. It is used frequently in places like ECS implementations, particle systems, object pooling, and plenty more. A brief demonstration is shown below:
+
+```lua
+local myFavouriteFruits = {"apple", "banana", "kiwi", "orange"}
+table.swapremove(myFavouriteFruits, 2)
+print(myFavouriteFruits) -- {"apple", "orange", "kiwi"}
+```
+
+Currently, to perform swap and pop, one might usually do the following:
+```lua
```
aatxe:
```suggestion
```luau
```

The programming language is Luau.

docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Summary
+
+This RFC introduces a new `table` library function, `table.swapremove(t, i)`, that implements the swap-and-pop algorithm for an array.
+
+## Motivation
+
+Swap-and-pop is a very common algorithm used to remove an element from an array in O(1) by replacing said element with the final element, and then removing that last element. This is significantly faster, especially for larger arrays, than a regular `table.remove`-like operation, which must shift all subsequent elements to preserve a contiguous layout. It is used frequently in places like ECS implementations, particle systems, object pooling, and plenty more. A brief demonstration is shown below:
+
+```lua
```
aatxe:
```suggestion
```luau
```

The programming language is Luau.

docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+
+A dedicated `table.swapremove` function would provide a single, well-defined abstraction for this common operation, offering behavior that is consistent with `table.remove`. It would also complete the table library by providing an unordered counterpart to `table.remove`. Users may implement swap-and-pop themselves, just as they might implement insertion with `t[i] = value` or append with `t[#t + 1] = value`. The value of `table.insert`, however, is not that it enables something previously impossible or impractical, but that it provides a clear, consistent, and optimised implementation of a common operation. This RFC proposes doing the same for unordered removal.
+
+Rust (`Vec`'s `swap_remove`), Zig (`ArrayList`'s `swapRemove`), and Unreal Engine (`TArray::RemoveAtSwap`) implement this.
+ 
+## Design
+
+The proposed signature would be very similar to that of the existing `table.remove` (using the style of the [current documentation](https://create.roblox.com/docs/reference/engine/libraries/table#remove)):
+```lua
```
aatxe:
```suggestion
```luau
```

The programming language is Luau.

docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+table.swapremove(t: {any}, pos: number): Variant
```
aatxe: This should be an actual Luau type signature.

```luau
table.swapremove<T>(table: {T}, index: number): T
```

docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+`table.swapremove` behaves similarly to `table.remove`. Namely:
```
aatxe: Comparison to `table.remove` doesn't really help explain the semantics here. You should just explain what the swap-and-pop algorithm does directly. That's the semantics of the function. That is something like:

`table.swapremove` will remove the element at the given position by moving the last element of the array into its position, and then return the element that was at the given position before the swap.

Perhaps even being more explicit that the last index is then set to `nil`.

docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+- Updates the VM's cached value for #t
+- Does nothing (silently) on integer indices which have a `nil` value, floors non-integer indices, and throws errors for non-numeric indices
+- Uses `pos = #t` if `pos` is omitted
+- Returns the removed element
+
+We propose using Rust and Zig's `swapremove` nomenclature (in luaucase), as opposed to something like `swapandpop` or `moveandpop`, or Unreal Engine's `RemoveAtSwap`.
```
aatxe: I don't think this is really a value-add of a sentence. It's already obvious that you're proposing the name you're proposing, and we don't really consider "it shares a name with another language" to be a strongly compelling reason to use a name. `swapremove` is a fine name because it's a description of the operation, and that stands on its own.

docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+## Design
+
+The proposed signature would be very similar to that of the existing `table.remove` (using the style of the [current documentation](https://create.roblox.com/docs/reference/engine/libraries/table#remove)):
+```lua
+table.swapremove(t: {any}, pos: number): Variant
+```
+
+`table.swapremove` behaves similarly to `table.remove`. Namely:
+- Updates the VM's cached value for #t
+- Does nothing (silently) on integer indices which have a `nil` value, floors non-integer indices, and throws errors for non-numeric indices
```
aatxe: I think this is a drawback of this design that is not described in the drawbacks section. While it may be the reasonable choice to no-op on arrays with holes (you don't make any argument for why it is, so it's hard to say!), this is also very readily a footgun. The alternative of swapping the last element to there and returning `nil` feels viscerally like it'd meet my expectations more, for instance. You don't have to propose that on my account, but the RFC should actually explore and discuss this design decision.

docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+ 
+## Design
+
+The proposed signature would be very similar to that of the existing `table.remove` (using the style of the [current documentation](https://create.roblox.com/docs/reference/engine/libraries/table#remove)):
+```lua
+table.swapremove(t: {any}, pos: number): Variant
+```
+
+`table.swapremove` behaves similarly to `table.remove`. Namely:
+- Updates the VM's cached value for #t
```
aatxe: You do not need to describe tiny implementation details like this. I don't think anyone expects removing an element from the array to leave the size of the array unchanged. 

Certainly though, you don't need to describe them in this particular way. I'd encourage you to think about and author proposals in terms of what you see as a user of the language, not the implementation details. It doesn't matter to a user that there is a VM for Luau or that the value of the length of an array is cached. The observable behavior at a language level is that the array's length is one smaller after removing the element.

docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+local myFavouriteFruits = {"apple", "banana", "kiwi", "orange"}
+table.swapremove(myFavouriteFruits, 2)
+print(myFavouriteFruits) -- {"apple", "orange", "kiwi"}
+```
+
+Currently, to perform swap and pop, one might usually do the following:
+```lua
+t[i], t[#t] = t[#t], nil
+```
+While this idiom is concise, it obscures the intent of performing an unordered removal and leaves the details of the operation to individual implementations. For example, a manual implementation does not return the removed value like `table.remove`, does not consistently validate (namely, throwing an error for non-numeric indices) or floor (for non-integers) the provided index, and may not update the VM's cached array length in the same way as the `table` library. A native implementation could improve performance.
```
aatxe: > A native implementation could improve performance.

How? Just stating this without an explanation for the mechanism by which this would happen is not compelling.

nnullcolumn: > while this idiom is concise

this disclaimer could probably be removed, `table.swapremove` is as concise as a bare swap remove
> it obscures the intent of performing an unordered removal

does it? many libraries already define swapremove functions to make the intent clear.

> may not update the VM's cached array length

i am pretty sure the length will shift down if `tbl[last]` is set to nil
> floor (for non-integers) the provided index

is that the best behavior here? i'd argue for putting "validating arguments" into the drawbacks section, user code not having to do that is probably a good trait
> a native implementation could improve performance.

by what meaningful metric
> leaves the details of the operation to individual implementations

what details?

docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+
+A dedicated `table.swapremove` function would provide a single, well-defined abstraction for this common operation, offering behavior that is consistent with `table.remove`. It would also complete the table library by providing an unordered counterpart to `table.remove`. Users may implement swap-and-pop themselves, just as they might implement insertion with `t[i] = value` or append with `t[#t + 1] = value`. The value of `table.insert`, however, is not that it enables something previously impossible or impractical, but that it provides a clear, consistent, and optimised implementation of a common operation. This RFC proposes doing the same for unordered removal.
+
+Rust (`Vec`'s `swap_remove`), Zig (`ArrayList`'s `swapRemove`), and Unreal Engine (`TArray::RemoveAtSwap`) implement this.
```
nnullcolumn: these are all statically typed languages with explicit array features. not that they're bad examples, but even fastcalls aren't free in the vm and validation would be kinda scary

docs/table-swapremove.md
nnullcolumn: > updates the vm's cached value for #t

might it be better to use maxn(t)

> Does nothing (silently) on integer indices which have a `nil` value, floors non-integer indices, and throws errors for non-numeric indices

the first two here (especially the latter) are not immediately obvious to me. why should we floor non-integers instead of erroring or silently cancelling?

docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+The proposed signature would be very similar to that of the existing `table.remove` (using the style of the [current documentation](https://create.roblox.com/docs/reference/engine/libraries/table#remove)):
+```lua
+table.swapremove(t: {any}, pos: number): Variant
+```
+
+`table.swapremove` behaves similarly to `table.remove`. Namely:
+- Updates the VM's cached value for #t
+- Does nothing (silently) on integer indices which have a `nil` value, floors non-integer indices, and throws errors for non-numeric indices
+- Uses `pos = #t` if `pos` is omitted
+- Returns the removed element
```
nnullcolumn: what if you want access to the element that got placed into the removed slot? what if you want access to the length of the array? as someone who has worked on ECS implementations, object pooling setups, and reactive stuff... those are all useful pieces of information that i wouldn't get from this function, it'd make the function almost exclusive with any of those usecases

docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+
+We propose using Rust and Zig's `swapremove` nomenclature (in luaucase), as opposed to something like `swapandpop` or `moveandpop`, or Unreal Engine's `RemoveAtSwap`.
+
+## Drawbacks
+
+It is yet another function to the table library that has a relatively simple implementation, like `table.insert` or `table.remove`, that users could reasonably implement themselves as discussed earlier.
+It is not immediately apparent to an unfamiliar user whether or not `swapremove` would return the swapped element or the removed element.
```
nnullcolumn, line 62: > It is not immediately apparent to an unfamiliar user whether or not `swapremove` would return the swapped element or the removed element.

why not both?

docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Alternatives
+
+We could simply not implement this function. Users could continue to use the swap assignment.
```
nnullcolumn, line 77: an explicit array datatype would be a potential alternative

docs/table-swapremove.md
nnullcolumn: this type signature should be `table.swapremove<V>(t: { V }, pos: number?): V?` rather than following the (very outdated) engine docs format

docs/table-swapremove.md
nnullcolumn: this is luau, not lua.

docs/table-swapremove.md
nnullcolumn: the biggest thing unsaid in this motivation section is that it handles "the item provided is the last item already" pretty nicely, whereas many libraries may specialize their code for that or even have bugs for it. thats a nice detail i didnt notice earlier

docs/table-swapremove.md
```diff
@@ -0,0 +1,49 @@
...
+- Uses `pos = #t` if `pos` is omitted
+- Returns the removed element
+
+We propose using Rust and Zig's `swapremove` nomenclature (in luaucase), as opposed to something like `swapandpop` or `moveandpop`, or Unreal Engine's `RemoveAtSwap`.
+
+## Drawbacks
+
+It is yet another function to the table library that has a relatively simple implementation, like `table.insert` or `table.remove`, that users could reasonably implement themselves as discussed earlier.
+It is not immediately apparent to an unfamiliar user whether or not `swapremove` would return the swapped element or the removed element.
+
```
nnullcolumn, line 63: i think we should mention in the drawbacks section that this encourages a pattern of several subsequent removals to parallel arrays, which may be slower or even lead to unsafe behavior

```luau
-- this is fine
table.swapremove(entites, row)
-- this is instantly slower in theory than a user implementation, and it also gets very yucky if the parallel array turns out to have holes. if we assume that the arrays are parallel in user code, neither issue would surface
table.swapremove(column, row)
```

docs/table-swapremove.md
nnullcolumn: i think the return signature would probably need to be optional

docs/table-swapremove.md
nnullcolumn: > You do not need to describe tiny implementation details like this. I don't think anyone expects removing an element from the array to leave the size of the array unchanged.

i think it was included because the author has some kind of misconception about what is actually going on with length, given the earlier mention of:

> may not update the VM's cached array length in the same way as the `table` library.

docs/table-swapremove.md
zermelon: I apologise. In future, I will use the proper signature style.

docs/table-swapremove.md
zermelon: I apologise - I didn't know if Github's MD files had support for luau codeblock rendering. Probably should have checked instead of naively assuming.

docs/table-swapremove.md
zermelon: > > it obscures the intent of performing an unordered removal

> does it?

Yes. I would argue that it is not immediately clear what this does, why it is beneficial, or why it is fast, when one sees this idiom, unless you are sufficiently familiar with arrays or know of swap and pop. if `t[i] = value` was similarly obscure enough to justify `table.insert`, I'm quite sure that `t[i], t[#t] = t[#t], nil` definitely is.

> > may not update the VM's cached array length

> i am pretty sure the length will shift down if tbl[last] is set to nil

Ah; I did not know this. I assumed naively that it would just be `table.insert`/`table.remove` (which are the two, evidently non-exhaustive, examples given in https://luau.org/performance/#optimized-table-length)

> > floor (for non-integers) the provided index

> is that the best behavior here? i'd argue for putting "validating arguments" into the drawbacks section, user code not having to do that is probably a good trait

I am trying to preserve parity with `table.remove` which does precisely this. I believe that general familiarity in implementation and signature would be beneficial. I'm not sure if I would consider this a drawback; it seems like a metric to optimise given we ideally want to maximise parity.
```luau
local myFavouriteFruits = {"apple", "banana", "kiwi", "orange"}
table.remove(myFavouriteFruits, 2.4)
print(myFavouriteFruits) -- {"apple", "kiwi", "orange"}
```
> > a native implementation could improve performance.

> by what meaningful metric

Good point. I haven't implemented it myself, though I presume that adding it natively would help as you don't compute #t twice (which, granted, would typically be O(1), but that ignores that this may still be a large constant). In any case, it was wrong of me to move to performance; I thought it would be at least worth mentioning but clearly not since I have no empirical metrics to back my claim.

> > leaves the details of the operation to individual implementations

> what details?

Whether or not to return the removed element, validation on inputs, and perhaps misleading function signatures. Standardising all of this is beneficial for these reasons. I imagine that whatever reason was used to motivate `table.insert` over `t[#t+1] = value`/ `t[i] = value`, other than immediate clarity, could equally work here.

Thank you for the comments.

docs/table-swapremove.md
zermelon: This is a good point. Drawing inspiration from statically typed languages isn't necessarily bad though, as you say. I wouldn't move to make it a fastcall, especially given `table.remove` isn't.

docs/table-swapremove.md
zermelon: > what if you want access to the element that got placed into the removed slot? 

You used the index in `table.swapremove(myarray, **index**)`, I think you can quite easily do `myarray[index]` afterwards. Even if it is the last element which you swapped-and-popped, myarray[#myarray] would (correctly) return nil.

> what if you want access to the length of the array?

`#myarray` could be of use.

> those are all useful pieces of information that i wouldn't get from this function

Why would you need this information? If we applied this logic to `table.remove`, you are effectively asking for something akin to "which element got shifted down first / last" and "how do I get the length of the array after removal". These are all trivial things to implement for a user and are beyond the scope of either of these functions. I do not believe these should be added as returns to this function.

docs/table-swapremove.md
zermelon: That would unnecessarily complicate the return signature, and turn it into `(T?, T?)` - the order isn't clear unless you look at documentation. Furthermore, most users who are indeed familiar with swapremove expect it to return only the removed element. The swapped element can easily be determined as I suggested previously. Also, in the case where we remove the last element (setting the positional argument to #t, or omitting it), the swapped element could equally be nil (as you don't really need to swap it with anything), or the last element itself, depending on implementation. This ambiguity should be avoided.

docs/table-swapremove.md
zermelon: Good point.

docs/table-swapremove.md
zermelon: Great point. I didn't actually think about that; I agree that it would make a lot of sense to perform the naive swap-and-pop and do exactly as you suggest for a sparse array.

docs/table-swapremove.md
zermelon: > You do not need to describe tiny implementation details like this. I don't think anyone expects removing an element from the array to leave the size of the array unchanged.

Will do, haha :)

> Certainly though, you don't need to describe them in this particular way. I'd encourage you to think about and author proposals in terms of what you see as a user of the language, not the implementation details. It doesn't matter to a user that there is a VM for Luau or that the value of the length of an array is cached. The observable behavior at a language level is that the array's length is one smaller after removing the element.

I'll amend the RFC significantly given to address this and all other points. I think I'll also remove the tag at the end of the motivation section about performance since I personally don't have anything to support my claim, and it is, as you say correctly, rather vague.

Kampfkarren:
I am not convinced that Luau needs to add something that can amount to three lines in a module. Performance as an argument is unremarkable as an argument because this pattern is both very easy to detect if we happened to have a special way to make this faster, and function inlining/cross function inlining are an obvious general solution.

zermelon:
> I am not convinced that Luau needs to add something that can amount to three lines in a module.

I've modified the RFC to address a number of excellent points made. I also don't think this logic stands particularly well in the face of `table.insert` or `table.remove`; as I explain in the RFC, the value of these is not in the fact that they are impossible to implement manually, but in standardisation, prevention of unexpected mistakes like providing an index outside of [1, maxn], and the returning of the removed value (which is beneficial for cleanup). You're right-on about performance though.

Kampfkarren:
> I also don't think this logic stands particularly well in the face of table.insert or table.remove

Inserting and removing something from an list are foundational functionalities of what a list even is. You can't seriously compare a precise method of removal to the base concept of inserting and removing.

docs/table-swapremove.md
nnullcolumn: > Why would you need this information? If we applied this logic to `table.remove`, you are effectively asking for something akin to "which element got shifted down first / last" and "how do I get the length of the array after removal". These are all trivial things to implement for a user and are beyond the scope of either of these functions. I do not believe these should be added as returns to this function.

i don't think they should be added as returns; i mean that these are all things readily available to me without any extra overhead in a user implementation, whereas doing them before or after a `table.swapremove` would carry extra overhead and be prone to implicitly invalidating that info if you got it before the swap remove. table.remove has a similar situation but i think the volume of info which swapremove interacts with is a lot higher

docs/table-swapremove.md
nnullcolumn: > Thank you for the comments.

Thank you for the thoughtful response.

> > > it obscures the intent of performing an unordered removal
> 
> > does it?
> 
> Yes. I would argue that it is not immediately clear what this does, why it is beneficial, or why it is fast, when one sees this idiom, unless you are sufficiently familiar with arrays or know of swap and pop. if `t[i] = value` was similarly obscure enough to justify `table.insert`, I'm quite sure that `t[i], t[#t] = t[#t], nil` definitely is.

That's a fair point, but people still need to learn what `table.swapremove` does

> > > floor (for non-integers) the provided index
> 
> > is that the best behavior here? i'd argue for putting "validating arguments" into the drawbacks section, user code not having to do that is probably a good trait
> 
> I am trying to preserve parity with `table.remove` which does precisely this. I believe that general familiarity in implementation and signature would be beneficial. I'm not sure if I would consider this a drawback; it seems like a metric to optimise given we ideally want to maximise parity.

My point was that in the VM, the validation expense and builtin call are *necessary* expenses we need to pay. In user code, the type system can validate it and users choose whether they'd like to add runtime branching.

> ```luau
> local myFavouriteFruits = {"apple", "banana", "kiwi", "orange"}
> table.remove(myFavouriteFruits, 2.4)
> print(myFavouriteFruits) -- {"apple", "kiwi", "orange"}
> ```
> 
> > > a native implementation could improve performance.
> 
> > by what meaningful metric
> 
> Good point. I haven't implemented it myself, though I presume that adding it natively would help as you don't compute #t twice (which, granted, would typically be O(1), but that ignores that this may still be a large constant). In any case, it was wrong of me to move to performance; I thought it would be at least worth mentioning but clearly not since I have no empirical metrics to back my claim.

It's true that `#t` being computed twice is bad, but all of my swap-remove code already does something like:

```luau
local count = #arr
arr[i], arr[count] = arr[count], nil
```

> > > leaves the details of the operation to individual implementations
> 
> > what details?
> 
> Whether or not to return the removed element, validation on inputs, and perhaps misleading function signatures. Standardising all of this is beneficial for these reasons. I imagine that whatever reason was used to motivate `table.insert` over `t[#t+1] = value`/ `t[i] = value`, other than immediate clarity, could equally work here.

I don't think "I imagine that whatever reason was used to motivate <other thing>" is a compelling argument that stands on its own. Validation on inputs can be handled by the type system. That said, I think I've come around to the idea now and a library fn would be pretty nice.

zermelon: > That's a fair point, but people still need to learn what `table.swapremove` does

Yes, but one could similarly argue that "people still also need to learn what `table.remove` does". It is unclear how exactly this removal works just from the name or signature alone. Without looking at documentation (which tbf, is something that you would have ready access to), you wouldn't know as a first time user that it also moves elements in the array down to preserve contiguity, or that it errors for non-numerical keys (which is even more counter-intuitive given that table.insert, which is its "twin" so far in the table library, _does_ accept any key), or that it floors numerical arguments, or does nothing for invalid numerical arguments like 0 or -1. I am not sure what your point overall is though, and I consider it moot. How does this differ from people still needing to learn what, say, `math.frexp` does as well? If you are familiar with what frexp and swapremove do, then you will of course be familiar with what this function would do.

> My point was that in the VM, the validation expense and builtin call are necessary expenses we need to pay. In user code, the type system can validate it and users choose whether they'd like to add runtime branching.

OK, this is fair enough. But this raises the natural question of why `table.remove` performs such checks anyway. I think both _should_ perform these checks, fwiw. The Luau team set a precedent that we should validate and/or correct invalid arguments, and so we should conform to this. In an ideal world, I wouldn't run these checks, but we are not there. With regards to the second point - I assume you mean something like numeric subtypes. If indeed we do get these, that would make this quite nice, though we would still need runtime checks anyway as static analysis presently has almost no effect on runtime behaviour, save for NCG and MULK/MUL as I understand it. Nevertheless, I write this RFC with the constraints of the present in mind, not the hopes for the future.

> It's true that `#t` being computed twice is bad, but all of my swap-remove code already does something like:
```luau
local count = #arr
arr[i], arr[count] = arr[count], nil
```

Good.

> I don't think "I imagine that whatever reason was used to motivate " is a compelling argument that stands on its own. Validation on inputs can be handled by the type system. That said, I think I've come around to the idea now and a library fn would be pretty nice.

I think it is a compelling argument. The Luau team willingly set the precedent that a function like `table.insert` is a worthwhile addition, even if it can be trivially written in a single statement, and I intend to make use of this precedent to justify this addition. I'm glad you've come around though! :)

zermelon:
> > I also don't think this logic stands particularly well in the face of table.insert or table.remove
> 
> Inserting and removing something from an list are foundational functionalities of what a list even is. You can't seriously compare a precise method of removal to the base concept of inserting and removing.

I don't disagree that insertion and ordered removal are more fundamental operations than `swapremove`; my comparison was meant to address your claim that something being able to be written in a few lines is, by itself, a reason not to include it. I don't think that criterion is sufficient, especially given that the standard library already provides these operations whose implementations are similarly trivial but whose value comes from having standardised semantics and a well-known API.

I would argue that it is a common pattern in performance-sensitive code where preserving order is unnecessary, and that having a standard operation makes this intent explicit while giving it defined behavior and an opportunity for a native implementation. Though I agree, that this is a stronger burden to meet than for `table.insert` or `table.remove`.

InfraredGodYT:
I don't think 2 separate tasks (removing and swapping) should be jammed into one stdlib function.

PR #237, opened 2026-08-03
Update cyclic requires RFC

annieetang:
[Rendered](https://github.com/luau-lang/rfcs/blob/2235a9fd1881a5f313b53ecb7e87d5138b4af5bc/docs/support-for-cyclic-requires.md)

Updating after some implementation iterations and discussions with the team! We found this approach to work best, and added some more technical details

### DIFF
--- a/docs/support-for-cyclic-requires.md
+++ b/docs/support-for-cyclic-requires.md
@@ -4,7 +4,7 @@
 
 Now that Luau is adding classes to the language, it's much more important that we afford some way modules to cyclically import one another.
 
-This RFC proposes that `require()` be augmented to pass an export table into the module.  This allows the runtime to close the loop and allow many cyclic import scenarios to work as desired.
+This RFC proposes that `require()` be augmented to automatically support cyclic imports for modules that use the `export` keyword.  This allows the runtime to close the loop and allow many cyclic import scenarios to work as desired.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


docs/support-for-cyclic-requires.md
```diff
@@ -47,73 +47,93 @@ Option 1 is laborious and sacrifices the fidelity of the type system.  Option 2
 
 ## Runtime Design
 
-For modules that return tables, we can solve this issue by having `require` tie the knot: When it encounters a cyclic import, `require` will instead return an empty table that will later be populated with the export surface of the module.  As long as the requesting module doesn't access it at the topmost global scope, that table will eventually be populated and everything will work out.  The system will temporarily attach a metatable to surface these issues and produce a clear error message.
+For modules that use the `export` keyword, we can solve this issue by having `require` tie the knot: When it encounters a cyclic import, `require` will instead return an empty table that will later be populated with the export surface of the module.  As long as the requesting module doesn't access it at the topmost global scope, that table will eventually be populated and everything will work out.  The system will temporarily attach a metatable to surface these issues and produce a clear error message.
```
hgoldstein: > ... everything will work out ...

I know this was here before, can we tighten that language a bit?

docs/support-for-cyclic-requires.md
```diff
@@ -4,7 +4,7 @@
 
 Now that Luau is adding classes to the language, it's much more important that we afford some way modules to cyclically import one another.
 
-This RFC proposes that `require()` be augmented to pass an export table into the module.  This allows the runtime to close the loop and allow many cyclic import scenarios to work as desired.
+This RFC proposes that `require()` be augmented to automatically support cyclic imports for modules that use the `export` keyword.  This allows the runtime to close the loop and allow many cyclic import scenarios to work as desired.
```
hgoldstein: Would it be worth linking to the export RFC?

I am not sure where you'd put it, but it also might be worth spelling out that the requirement is that all modules which want to participate _in_ a cycle must use `export`. It's not like you can have:
```
A -> B -> C -> D -> E
```
... then change `E` to `E'` which uses export syntax and add a cycle ...
```
A -> B -> C -> D -> E' -> A
```

docs/support-for-cyclic-requires.md
```diff
@@ -47,73 +47,93 @@ Option 1 is laborious and sacrifices the fidelity of the type system.  Option 2
...
+For modules that use the `export` keyword, we can solve this issue by having `require` tie the knot: When it encounters a cyclic import, `require` will instead return an empty table that will later be populated with the export surface of the module.  As long as the requesting module doesn't access it at the topmost global scope, that table will eventually be populated and everything will work out.  The system will temporarily attach a metatable to surface these issues and produce a clear error message.
 
-There are subtle edge cases to consider here:
+The restriction for cyclic module support is deliberate: 
+1. `export` modules have a well-defined export surface that the runtime can reason about
+2. the `export` keyword signals intent to participate in the module system's advanced features.
```
hgoldstein: I don't think we need this line, but if you keep it:
```suggestion
2. The `export` keyword signals intent to participate in the module system's advanced features.
```

docs/support-for-cyclic-requires.md
```diff
@@ -47,73 +47,93 @@ Option 1 is laborious and sacrifices the fidelity of the type system.  Option 2
...
+export function getA() return A.name end
+
+--- main.luau
+require("./A")
+```
+
+Order of operations:
 
-#### Reentrant Accesses
+1. `main.luau` requires `A.luau`. A placeholder is created for A (it uses `export`).
```
hgoldstein: Can you do me a favor and give these modules all non-single-letter names? "A placeholder is created for A" made my head spin.

It is also probably worth spelling out that, at this point, requiring `A` now begins executing `A.luau`. "A placeholder is created for A" also makes it sound like we continue requiring `B` rather than preserving the current depth-first semantics.

docs/support-for-cyclic-requires.md
```diff
@@ -183,25 +203,59 @@ If we naively execute our planned resolution order, things proceed as follows:
...
+3. `B.luau` requires `A.luau`. Cycle detected, but A has no placeholder — error: `"Requested module was required recursively"`.
+
 ## Type System Design
 
 The user-facing behaviour of the type inference engine should be unchanged as a result of this RFC, but the internal structure of the type checker is going to need significant changes.
 
 Today, typechecking is driven by a class called `Frontend`.  It accepts a set of modules that need checking, builds a DAG from that, and checks modules one after another.
 
-We will augment this class to instead work on one strongly-connected component\* at a time.  All modules within an SCC use the same arena and are typechecked together in a single pass through the solver.
+We will augment this class to instead work on one strongly-connected component\* at a time.  We will use Tarjan's algorithm to identify strongly-connected components (SCCs) in the module dependency graph.  Then, all modules within an SCC use the same arena and are typechecked together in a single pass through the solver.
```
hgoldstein: I don't think you need to namedrop the algorithm, and if you abbreviate strongly connected component, you should do so after the first usage, not the second.

docs/support-for-cyclic-requires.md
```diff
@@ -183,25 +203,59 @@ If we naively execute our planned resolution order, things proceed as follows:
...
 
 \* A "strongly connected component" is a set of 2 or more modules that all mutually `require()` one another. (eg if you had a require chain of `A -> B -> C -> A`, the SCC would consist of `A`, `B`, and `C`)
```
hgoldstein: We could also link to the wikipedia page here.

docs/support-for-cyclic-requires.md
```diff
@@ -221,10 +275,13 @@ class ClassBOne ... end
...
+An earlier design passed the export table to the module as a vararg (`...`), requiring modules to opt in with `local exports = ...` or `local exports = ... or {}`. This was implemented then amended to use the `export` keyword instead because:
 
-The current proposal is not to do this because `...` is a preexisting mechanism that works really well to solve this class of problem and because the edge cases don't seem very difficult to deal with.
+- This would be a breaking change for existing code that used the vargs mechanism
+- It required boilerplate in every module that wanted cycle support.
+- The `export` keyword provides a cleaner, more explicit signal of intent, and clear boundary where only modules with a known, table-shaped export surface can participate in cycles.
+- This simplifies some edge cases because there would be no way, for instance, to attach a metatable to the current module's exports.
```
hgoldstein: Should this focus more on the downsides here, rather than justifying the chosen design?

docs/support-for-cyclic-requires.md
```diff
@@ -17,214 +17,274 @@ This problem becomes much more difficult to deal with when classes are added to
...
+### Large Cycle Warning
 
 A problem that a developer might run into is that, if their application consists of a very large SCC (their whole application, perhaps\!), their incremental typechecking performance will be very bad: Luau will have to recheck all files whenever any file in the SCC has changed.
 
-To mitigate this and put some soft pressure on the developer, we'll report a warning when we encounter an SCC that consists of too many modules.  This warning will explain that large clusters of cyclic modules can cause typechecking performance to degrade badly.  We'll allow this limit to be configured via `FrontendOptions`.
+To mitigate this and put some soft pressure on the developer, we'll report a warning when we encounter an SCC that consists of too many modules.  This warning will explain that large clusters of cyclic modules can cause typechecking performance to degrade badly.  We'll allow this limit to be configured via `FrontendOptions`, defaulting to 4 modules.  The warning will list the names of the modules in the SCC, up to a configurable limit (default 10).
```
nnullcolumn, line 255: what is the approach users should take to mitigating this warning? to me it seems non-obvious where such a warning would span and how people would go about fixing it if they've deliberately ignored it or cranked the limits in the past.

docs/support-for-cyclic-requires.md
```diff
@@ -17,214 +17,274 @@ This problem becomes much more difficult to deal with when classes are added to
...
+
+### Type System Algorithm
 
-We will augment this class to instead work on one strongly-connected component\* at a time.  All modules within an SCC use the same arena and are typechecked together in a single pass through the solver.
+1. Detect strongly-connected components in the module dependency graph.
+2. Created a shared type arena for all modules in the SCC, allowing types from any member to reference types from other members.
+3. Pre-allocate a `BlockedType` placeholder for each module in the SCC.  This allows the typechecker to resolve references to types from other modules in the SCC, even if they haven't been fully defined yet.
+4. Run constraint generation for each module in the SCC. Each module in the SCC gets its own scope and data flow graph, but all constraints flow into a single shared `ConstraintGraph`. After constraint generation for each module, the `BlockedType` placeholder is bound to the actual inferred return type.
+5. Run constraint solver once over the combined constraint graph for the entire SCC. This enables cross-cycle type inference, so constraints from `foo` that depend on types from `bar`, for example, are solved together.
```
nnullcolumn, line 248: > Pre-allocate a `BlockedType` placeholder for each module in the SCC.

> After constraint generation for each module, the `BlockedType` placeholder is bound to the actual inferred return type.

> This enables cross-cycle type inference, so constraints from `foo` that depend on types from `bar`, for example, are solved together.

would this have a lot of side effects? it seems like changing the *entire module* to be populated with a blocked type would have untold implications on constraint generation

docs/support-for-cyclic-requires.md
```diff
@@ -17,214 +17,274 @@ This problem becomes much more difficult to deal with when classes are added to
...
-2. `A.luau` starts importing `B.luau`
-3. `B.luau` attempts to import `A.luau`.  We sense the cycle and short circuit; the incomplete module `A` is returned immediately.
-4. `B.luau` attempts to access `A.Tree`.  The value `A` is still incomplete and therefore has the `CyclicDependencyError` metatable attached to it.  We tell the developer that a cyclic dependency error has been encountered and raise an exception.  The developer can use the stack trace to understand the cycle.
+1. `main.luau` requires `folder.luau`. A placeholder is created for `folder` and it begins executing.
+2. `folder.luau` requires `file.luau`. A placeholder is created for `file` and it begins executing.
+3. `file.luau` requires `folder.luau`. Cycle detected — `folder`'s locked placeholder is returned immediately.
+4. `file.luau` attempts to access `folder.Folder`.  The placeholder is still incomplete and therefore has the `CyclicDependencyError` metatable attached to it.  We tell the developer that they `"Cannot access the exported field 'Folder' because it has a cyclic dependency on its requiring module"` and raise an exception.  The developer can use the stack trace to understand the cycle.
```
nnullcolumn, line 171: > The developer can use the stack trace to understand the cycle.

can they? a really long stack (or several of them!) could be really annoying to sift through.

it's not as simple as ModuleA -> ModuleB -> ModuleC -> ModuleD -> ModuleA. A stack might contain several traces from the same module, might be very long, might contain leading traces unrelated to the main cycle, it might even contain mutually recursive traces, etc...

docs/support-for-cyclic-requires.md
annieetang: The `BlockedType` isn't replacing the module's internal types, it's just the return type seen by other modules in the SCC cycle during just constraint generation. Each module's local scope is still generated normally with concrete types. The `BlockedType` is just the placeholder to `require(otherModuleInSCC)` before we've solved it. So once constraint generation finishes for a module, we bind the `BlockedType` to the actual inferred return type, and then the solver resolves everything together.

docs/support-for-cyclic-requires.md
annieetang: The mitigation would be to break the cycle via an interface module or restructuring the dependency graph, something users today would be doing anyways. For ex, if you have `A->B->C->D->A`, you could extract the shared types/interfaces into a separate non-cyclic modules that the others all require, which can reduce the SCC or even remove a cycle entirely.

Regular type errors are still reported in each module, but SCC size warning currently attaches to the first module in the cycle, though we're still working on improving the error/warning process in SCCs

docs/support-for-cyclic-requires.md
annieetang: Ah that's a fair point, it could get very noisy, and we can note this as a future improvement needed, like somehow including the module cycle path in error messages...

docs/support-for-cyclic-requires.md
nnullcolumn: i'm asking if things will behave in visibly different ways to users - it seems necessary to me that they would, but i can't picture how that would be side effect free. If I took, say, a 500,000 line luau codebase, got my depgraph to form the *entire thing* into an scc, and we ignore all of the solver's time limits and whatnot - would typechecking behave the same as before? on a smaller scale, can modules influence the types of modules they require if they're in the same scc? i don't totally see why not (i mean, wouldn't you want that for some things--like two datastructures or functions depending on eachother?) or, what about cases like this:

image

[(playground link)](https://play.luau.org/#N4IgbiBcBMA0IDMqgE4FMAOKD2A6ANgK4CGhUI+2AxsfgAQC2a2A7nQLx3oCOhAlugAUAHRC4A9E1aiAlMIB2ChVWzyAzgBc6CQvKoa+quhrSbBfeQjQp0AEzny6jZi1wMbGc5et2HT9BqEKI4WVjZotgpo8pGK8iDwUq5EpOQq6lo6egZGSYLEkHTyhAwARtYO0bEKAUGOwApO7igYHM7S8gC+IJ3wxOToWHgpZPBqyCDYGAYMfABexDnyADJoYGj4UHAgtmilhADmq+ubkAAM8CoMGHz41gBKaAzEKADW45AaKIRonZ1AA)

would the type solver still be able to infer `number` here if `meow`'s value were a blocked type? my hope would be yes because `meow.mrrp` would infer to `index<meow, "mrrp">`, but i don't really know that (and currently if i reproduce the same type function with a generic, i get an error about `Cannot call a value of type index<value, "mrrp">`):

image

[(playground link)](https://play.luau.org/#N4IgbiBcBMA0IDMqgE4FMAOKD2A6ANgK4CGhUI+2AxsfgAQKEB2VALgJbZN0C22AJoXxoAPGFqE0APgAUxSHXZN+aAB5iJaWHQA6IHihQY9UgJQ7udYjICM0AMzmmaZSAC+8AM7IQ2DBx52AC9iDi4AGTQwNHwoOBAVACNCAHNI6NjIAAZ4KmweDHZhFAAlNB5iFABrb0hWFEk3NyA)

to say i have limited experience would be an overstatement of my skill-set, so apologies if i have the wrong idea entirely or am missing something important. "each module will get a blocked type placeholder" just sounds scary to me with only a vague idea of what a blocked type actually is beyond "something we can't know yet"

docs/support-for-cyclic-requires.md
nnullcolumn: > The mitigation would be to break the cycle via an interface module or restructuring the dependency graph, something users today would be doing anyways. For ex, if you have `A->B->C->D->A`, you could extract the shared types/interfaces into a separate non-cyclic modules that the others all require, which can reduce the SCC or even remove a cycle entirely.

this doesn't really address my question. not every cycle is a sparse graph, it might be
A->B
A->C
C->A
C->B
B->A
B->C
A->D
D->A
D->B

...etc. My question was mostly about what we're actually telling users to do if their graphs start looking like that and they want their graphs to stop looking like that. Four also seems a tad low to me

> Regular type errors are still reported in each module, but SCC size warning currently attaches to the first module in the cycle, though we're still working on improving the error/warning process in SCCs

right, but i think attaching a warning to potentially dozens of files with no clear starting point (literally any one of them might be removable entirely and you could still have the warning in 11 other files) will just lead users to crank the limits with no clear path to refactor it without starting from scratch. the solver isn't aware of how your code evolved, it can only see that there's a problem. and there might be a problem for... hundreds of files, which goes unnoticed or ignored, and now it's impossible to really find the source of it

docs/support-for-cyclic-requires.md
nnullcolumn: that seems okay, but i think it should be addressed in the drawbacks section

PR #238, opened 2026-08-06
`if local` statements and expressions

annieetang:
status: implementing

[Rendered](https://github.com/luau-lang/rfcs/blob/if_local/docs/syntax-if-local.md)

This RFC introduces the syntax `if local`, and is a tighter scoped version of https://github.com/luau-lang/rfcs/pull/110! We'd love to start the work towards what was discussed there and think that a single binding per `if` is the minimal useful case and would still be valuable on its own.

### DIFF
--- a/docs/syntax-if-local.md
+++ b/docs/syntax-if-local.md
@@ -1,0 +1,197 @@
+# `if local`
+
+## Summary
+
+Allow a `local` or `const` binding in the condition of an `if` or `elseif`. The variable is scoped to the `then`\-block, and the branch is taken when the initializer is truthy. `const` is included alongside `local` because it is already a contextual keyword valid everywhere `local` is valid.
+
+```lua
+if local x = expr() then
+    -- x is truthy here
+end
+
+if const y = expr() then
+    -- y is truthy here, and cannot be reassigned
+end
+```
+
+## Motivation
+
+A common pattern is to call something that might return nil, bind the result, and then check it:
+
+```lua
+-- users: {[number]: User}, where User = {email: string?}
+local user = users[id]
+if user then
+    local email = user.email
+    if email then
+        sendEmail(email)
+    end
+end
+```
+
+Every nil-checked access takes two lines, and the variable leaks into the surrounding scope where it may be nil and shouldn't be used. This shows up anywhere an API returns `T?`, like in table lookups, string matching, and Roblox code like `Player.Character`, or `FindFirstChild`
+
+With `if local`:
+
+```lua
+if local user = users[id] then
+    if local email = user.email then
+        sendEmail(email)
+    end
+end
+```
+
+The variable's scope is exactly where it's guaranteed to be non-nil. Note that this still requires nesting for the two-step lookup above; a single binding per `if` is the minimal useful case, and chaining multiple bindings into one condition is left for future work (see below).
+
+## Design
+
+### Syntax
+
+When the parser sees `local` or `const` immediately after `if` or `elseif`, it parses:
+
+```lua
+('local' | 'const') Name [':' Type] '=' exp
+```
+
+The initializer is the entire condition: nothing else appears between it and `then`. Since neither `local` nor `const` can start an expression in Luau, there is no ambiguity. `const` here is the same contextual keyword already used for `const` statements – no new keyword is being introduced.
+
+### Semantics
+
+The initializer is evaluated. If the result is truthy (not `nil`, not `false`), the branch is taken and the binding is available inside the `then`\-block. Otherwise execution falls through to `elseif` or `else`. The variable does not exist outside its block. A `const` binding follows the same reassignment restriction it has as a statement: the binding itself can't be reassigned inside the `then`\-block, though a mutable value it refers to (e.g. a table) can still be mutated.
+
+### Scoping
+
+Each branch introduces its own independent binding:
+
+```lua
+-- admins and guests: {[number]: {name: string}}
+if local admin = admins[id] then
+    print(admin.name) -- admin is in scope
+elseif local guest = guests[id] then
+    print(guest.name) -- guest is in scope, admin is not
+else
+    -- neither in scope
+end
+-- neither in scope
+```
+
+A binding from one branch is never visible while evaluating a later branch's own condition, since that condition only runs if the earlier one was falsy.
+
+If a variable of the same name already exists in an enclosing scope, `if local x = ...` (or `if const x = ...`) shadows it for the duration of the `then`\-block, exactly as an ordinary `local x = ...` or `const x = ...` statement would. The outer `x` is unaffected and reappears after the block ends.
+
+### Type narrowing
+
+The type checker narrows the binding using the same truthiness refinement it already applies to `if x then` — removing `nil` and `false` from the type:
+
+```lua
+-- users[id] returns User?
+if local user = users[id] then
+    -- user: User (not User?)
+    user.lastSeen = os.time()
+end
+```
+
+### Desugaring
+
+`if local x = E then B end` is equivalent to:
+
+```lua
+do
+    local x = E
+    if x then
+        B
+    end
+end
+```
+
+and likewise `if const x = E then B end` desugars to a `do` block with a `const x = E` statement in place of `local`. No new bytecode instructions or evaluation rules are needed for either form. It is syntactic sugar with tighter scoping.
+
+## Drawbacks
+
+**No compound conditions yet.** There is no way to write `if local x = expr() and x > 0 then`, where the branch is taken when x is greater than zero. The binding *is* the condition. See Alternatives for why this is out of scope for now rather than an open design question.
+
+## Alternatives
+
+### Do nothing
+
+The two-line pattern works, but it is verbose enough that developers often skip the nil-check entirely, leading to "attempt to index nil" errors on instance access.
+
+### Compound conditions
+
+A previous [RFC (\#110)](https://github.com/luau-lang/rfcs/pull/110) proposed multi-binding stacked declarations and an `in` keyword for additional conditions. Discussion on that RFC did not have a consensus on the syntax for chaining a boolean test after the binding: every candidate delimiter is either an existing token being given new meaning (`;`, which the language has otherwise avoided assigning semantics to) or a new reserved keyword (`where`). We think chaining can be added later, given that truthy checks will be the vast majority of usage.
+
+## Possible Future Work
+
+These can build on top without breaking changes, and are intentionally not resolved here:
+
+### Compound conditions
+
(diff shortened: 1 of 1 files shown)


deviaze:
I would've scoped my RFC smaller if I were given actionable feedback and clearer communication to that end.

`while local/const` bindings are a major usecase, and previous discussion has held them as more important than `if local/const`; I don't feel they should be left for future work.

Bottersnike:
> When the parser sees local or const immediately after if or elseif, it parses:
> ```lua
> ('local' | 'const') Name [':' Type] '=' exp
> ```
> [snip] Since neither `local` nor `const` can start an expression in Luau, there is no ambiguity

This is not true. `const` is a valid variable name and as such can absolutely be the start of an expression.

```lua
local const = true
if const then
  ...
end
```
Is legal code. This can probably be resolved with an additional lookahead as I don't think `'if' Name('const') Name` is currently legal.

annieetang:
> I would've scoped my RFC smaller if I were given actionable feedback and clearer communication to that end.
> 
> `while local/const` bindings are a major usecase, and previous discussion has held them as more important than `if local/const`; I don't feel they should be left for future work.

just recapping here from the discussion in OSS but we will (and always were going to!) list you as a co-author on this as well and really appreciate the discussions and ideas on this, and plan to keep working with you to extend this to more of the use cases that we've been discussing in other threads/rfcs 😊

Vighnesh-V:
@deviaze I would agree with `while local` being an important part of the feature as a whole - it is something the team actively wants to deliver. I would disagree with it being more important that `if local` (anecdotally).

deviaze:
> @deviaze I would agree with `while local` being an important part of the feature as a whole - it is something the team actively wants to deliver. I would disagree with it being more important that `if local` (anecdotally).

Oh, I personally agree with that take too. I need `if local` much more than `while local`, but I've been told multiple times that 

> "`if local` doesn't clear the -100 points bar -- you can already do `if local` with `do end` or another line of code, whereas `while local` is the feature I genuinely need".

deviaze:
Intro should clarify that this feature only applies to `if local` statements; `if local` expressions are not in scope, and future work should include why `if local` *expressions* are significantly more complex (compiler work).

Vighnesh-V:
While I personally agree with you and think that the consistent minimal feature set here should also include `while`, I don't think `if local` on it's own doesn't clear the -100 points bar. It is useful on its own, and its addition to the language would be a net positive for folks. I can support a tighter scoped version of this rfc because I think its a) extensible and b) delivers value pretty quickly for people.

annieetang:
> Intro should clarify that this feature only applies to `if local` statements; `if local` expressions are not in scope, and future work should include why `if local` _expressions_ are significantly more complex (compiler work).

Ah that's a good callout, thanks! I'm happy to extend the scope to supporting `if local` expressions, so something like 

```lua
local y = if local x = foo() then x else "a"
```

should be supported. But I'd still want to keep `while local` for future work to avoid scope creeping, and also because `if` and `while` are different syntactic constructs.

deviaze:
> > Intro should clarify that this feature only applies to `if local` statements; `if local` expressions are not in scope, and future work should include why `if local` _expressions_ are significantly more complex (compiler work).
> 
> Ah that's a good callout, thanks! I'm happy to extend the scope to supporting `if local` expressions, so something like
> 
> ```lua
> local y = if local x = foo() then x else "a"
> ```
> 
> should be supported. But I'd still want to keep `while local` for future work to avoid scope creeping, and also because `if` and `while` are different syntactic constructs.

Yeah I'd love `if local` expressions. I was just told they were much more significant work! I've wanted both. If possible, would the compiler work to unblock `if local` expressions also technically unblock binding declarations in expressions (for future RFCs on block expression syntax ala Rust)?

vegorov-rbx:
> I was just told they were much more significant work!

After we evaluated compiler closer, it is more work, but does not look significantly harder to avoid if we like the feature (same for block expressions).

annieetang:
just updated to add `if local` expressions :)

annieetang:
hey everyone! let us know if you have further design comments about this. since it's been about 2 weeks, we're going to move on and start implementing this, but will leave the RFC open to keep the discussion open if there are new things that come up

SeloTheGreat:
small thing, the `Rendered` link points to a specific commit in the branch instead of the most recent version of the file; people that come here to check out the RFC document might initially be confused from this.

This should be the proper link: https://github.com/luau-lang/rfcs/blob/if_local/docs/syntax-if-local.md

PhoenixWhitefire:
> This is not true. `const` is a valid variable name and as such can absolutely be the start of an expression.

Bottersnike's comment appears to never have received a response, and the issue she was referring to has manifested in the experimental implementation with previously valid code failing to parse:

https://github.com/luau-lang/luau/issues/2968

vegorov-rbx:
> > This is not true. `const` is a valid variable name and as such can absolutely be the start of an expression.
> 
> Bottersnike's comment appears to never have received a response, and the issue she was referring to has manifested in the experimental implementation with previously valid code failing to parse:
> 
> [luau-lang/luau#2968](https://github.com/luau-lang/luau/issues/2968)

RFC received a direct update to address that in https://github.com/luau-lang/rfcs/pull/238/commits/0d6ceaa1318ab95cbd9d272e01381f74e2b1704b explaining the situation with `const`.

We already detected the implementation issue and fixed it yesterday, it will be in the sync next week.

docs/syntax-if-local.md
```diff
@@ -0,0 +1,291 @@
+# `if local` statements and expressions
+
+## Summary
+
+Allow a `local` or `const` binding in the condition of `if` or `elseif` statements or expressions. The variable is scoped to the `then`\-block, and the branch is taken when the initializer is truthy. `const` is included alongside `local` because it is already a contextual keyword valid everywhere `local` is valid.
+
+```lua
```
nnullcolumn: the language is Luau

docs/syntax-if-local.md
```diff
@@ -0,0 +1,291 @@
...
+    -- x is truthy here
+end
+
+-- expression
+local y = if local x = expr() then x else "default"
+```
+
+`const` works the same way, with the additional guarantee that the binding cannot be reassigned within the block:
+
+```lua
```
nnullcolumn: re: luau

docs/syntax-if-local.md
```diff
@@ -0,0 +1,291 @@
...
+if const x = expr() then
+    -- x is truthy here and cannot be reassigned
+end
+```
+
+## Motivation
+
+A common pattern is to call something that might return nil, bind the result, and then check it:
+
+```lua
```
nnullcolumn: re: luau

docs/syntax-if-local.md
```diff
@@ -0,0 +1,291 @@
...
+        sendEmail(email)
+    end
+end
+```
+
+Every nil-checked access takes two lines, and the variable leaks into the surrounding scope where it may be nil and shouldn't be used. This shows up anywhere an API returns `T?`, like in table lookups, string matching, and Roblox code like `Player.Character`, or `FindFirstChild`
+
+With `if local`:
+
+```lua
```
nnullcolumn: re: luau

docs/syntax-if-local.md
```diff
@@ -0,0 +1,291 @@
...
+
+The variable's scope is exactly where it's guaranteed to be non-nil. Note that this still requires nesting for the two-step lookup above; a single binding per `if` is the minimal useful case, and chaining multiple bindings into one condition is left for future work (see below).
+
+## Design
+
+### Syntax
+
+When the parser sees `local` or `const` immediately after `if` or `elseif`, it parses:
+
+```lua
```
nnullcolumn: this fence shouldn't have `lua`

docs/syntax-if-local.md
```diff
@@ -0,0 +1,291 @@
...
+
+### Semantics
+
+The initializer is evaluated. If the result is truthy (not `nil`, not `false`), the branch is taken and the binding is available inside the `then`\-block. Otherwise execution falls through to `elseif` or `else`. The variable does not exist outside its block. A `const` binding follows the same reassignment restriction it has as a statement: the binding itself can't be reassigned inside the `then`\-block, though a mutable value it refers to (e.g. a table) can still be mutated.
+
+### Scoping
+
+Each branch introduces its own independent binding:
+
+```lua
```
nnullcolumn: re: luau

docs/syntax-if-local.md
```diff
@@ -0,0 +1,291 @@
...
+
+A binding from one branch is never visible while evaluating a later branch's own condition, since that condition only runs if the earlier one was falsy.
+
+If a variable of the same name already exists in an enclosing scope, `if local x = ...` (or `if const x = ...`) shadows it for the duration of the `then`\-block, exactly as an ordinary `local x = ...` or `const x = ...` statement would. The outer `x` is unaffected and reappears after the block ends.
+
+### Type narrowing
+
+The type checker narrows the binding using the same truthiness refinement it already applies to `if x then` — removing `nil` and `false` from the type:
+
+```lua
```
nnullcolumn: re: luau

docs/syntax-if-local.md
```diff
@@ -0,0 +1,291 @@
...
+    -- user: User (not User?)
+    user.lastSeen = os.time()
+end
+```
+
+#### Explicit type annotations
+
+Since `if local` is syntactic sugar for a `local` declaration followed by a truthiness test (see [Desugaring](#desugaring)), an annotation is checked exactly as it would be for that `local`, so
+
+```lua
```
nnullcolumn: re: luau

docs/syntax-if-local.md
```diff
@@ -0,0 +1,291 @@
...
+if local x: T = y then
+    -- ...
+end
+```
+
+behaves like
+
+```lua
```
nnullcolumn: re: luau

docs/syntax-if-local.md
```diff
@@ -0,0 +1,291 @@
...
+local x: T = y
+if x then
+    -- ...
+end
+```
+
+Additionally, like a `local` declaration, the annotation is optional. When the initializer returns an optional or `any` type, we annotate with the optional type (or omit it) and let the branch refine it:
+
+```lua
```
nnullcolumn: re: luau

docs/syntax-if-local.md
```diff
@@ -0,0 +1,291 @@
...
+-- expr(): number?
+if local x: number? = expr() then
+    -- declared type is number?, but x is refined to number here
+end
+```
+
+The value being bound still has to match the annotation's type, so an incompatible annotation is a type error:
+
+```lua
```
nnullcolumn: re: luau

PR #239, opened 2026-08-09
`@nolint` syntax attribute

nnullcolumn:
[rendered](https://github.com/nnullcolumn/luau-lang-rfcs/blob/warning-suppression-attribute/docs/syntax-attributes-nolint.md)

### DIFF
--- a/docs/syntax-attributes-nolint.md
+++ b/docs/syntax-attributes-nolint.md
@@ -23,7 +23,7 @@ more specific, the `@nolint` attribute should only silence warnings if they are 
 
 `@nolint` will also have optional attribute parameters allowing users to exclude specific lint names with either string
 literals (`@[nolint "LocalUnused"]`), or lists of string literals (`@[nolint("LocalUnused", "MisleadingCondition")]`).
-This is very coherent with the existing top-level comment `--!nolint LocalShadow`.
+This is very coherent with the existing top-level comment capabilities `--!nolint LocalShadow`.
 
 For now, pattern matching, message matching, et cetera are out of scope for this RFC, because they
 would be significantly more complex and leave wild room for tools to vary in implementation without
--- a/docs/syntax-attributes-nolint.md
+++ b/docs/syntax-attributes-nolint.md
@@ -52,9 +52,9 @@ and of itself.
 
 ## Alternatives
 
-- `@lint` / `@[lint { name = true } ]`. This is more coherent with `.config.luau` files, but most cases for in-line lint
-silencing only need to specify which lints they are looking for, values of `true` seem like they'd become repetitive
-when compared to a declaration with a larger scope, are cumbersome to type, and don't allow people to easily mask all
+- `@lint` / `@[lint { name = true } ]`. This is more coherent with `.config.luau` files. most cases for inline lint
+silencing at most need to specify exclusions, so values of `true` seem like they'd become repetitive when compared to a
+declaration with a larger scope. this would also be cumbersome to type, and doesn't allow people to easily mask all
 lints as they could with `@nolint`.
 
 - Name it `nowarn` instead. This is disconnected from the file-wide flag `--!nolint`, which could be interpreted as a
--- a/docs/syntax-attributes-nolint.md
+++ b/docs/syntax-attributes-nolint.md
@@ -61,8 +61,15 @@ lints as they could with `@nolint`.
 boon or a flaw. However, users might also conflate `nowarn` with runtime warnings, and we want to avoid confusing
 people.
 
-- Use a comment instead. The only direct benefit to this is compatibility when running in a lua runtime, which seems
-moot at this point as many luau features today (and from the start) are exclusive with lua's syntax expectations.
+- Use a comment instead.  This would make the implementation a fair
+bit simpler, as it would mean that all of the logic could be confined
+to the linter itself, but the ergonomics are poor:  Because comments
+are blind to the structure of the source code, there is nothing to
+prevent a developer from inadvertently disabling a warning for more
+than the intended span.  It's quite nice to be able to write a small
+annotation at the top of a declaration or statement to disable a
+warning for that entire construct with no need for an extra directive
+at the end.
 
 - Include `start` / `finish` style syntax instead of scoping it syntactically. This makes scoping much more ambiguous to
 users, and is cumbersome for simply allowing e.g. `@nolint func()`. However, this idea isn't really exclusive with this
--- a/docs/syntax-attributes-nolint.md
+++ b/docs/syntax-attributes-nolint.md
@@ -71,7 +71,7 @@ RFC as it serves a different purpose.
 - Do nothing, which means linter projects will need to invent their own syntax or hot comments for this (e.g.,
 the formatter project StyLua relies on `--stylua: ignore` / `--stylua: ignore start`). Each project may have different
 syntax or rules, which is a double edged sword; those projects become more expressive, but users relying on multiple of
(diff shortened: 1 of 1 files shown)


Kampfkarren:
Are attributes supported outside statements yet? You would need to support everything, including expressions, in order for this to be better than a comment. 

It also feels weird to both repeatedly mention the use case for external tooling and linters and also say:

> However, it seems like linting will eventually move to being the sole responsibility of lute

In which case, custom tooling *has to* not support this, since they don't want their names being linted.

I think that problem clears up with namespacing (e.g. `@nolint("lute.ABC")`, making it the responsibility of the namespace to check for valid lints, while Luau itself lints any other name in the global namespace that it doesn't recognize.

nnullcolumn:
> Are attributes supported outside statements yet? You would need to support everything, including expressions, in order for this to be better than a comment.

this RFC proposes allowing that for an `@nolint` attribute. I don't think we should rely on comments.

> It also feels weird to both repeatedly mention the use case for external tooling and linters and also say:
> 
> > However, it seems like linting will eventually move to being the sole responsibility of lute
> 
> In which case, custom tooling _has to_ not support this, since they don't want their names being linted.

sorry, i don't think i communicated too well. might need to update the body a bit. i was thinking `.config.luau` or similar would help configure to

A) ignore it entirely

B) specify all of the lints we care about

that was what i meant when i wrote "emit a configurable warning in and of itself"

> I think that problem clears up with namespacing (e.g. `@nolint("lute.ABC")`, making it the responsibility of the namespace to check for valid lints, while Luau itself lints any other name in the global namespace that it doesn't recognize.

this seems nicer than what i originally had in mind / mentioned above. it'd probably require changing the `.config.luau` structure up a bit, but seems like a big improvement

p.s., `@nolint("foo")` is currently disallowed syntax. you're forced to do `@[nolint("foo")]` for parameters.

andyfriesen:
If we built this system using comments as an extension to `--!nolint`, we could implement it entirely within the linter itself.  No changes to the language would be necessary.

How about this:

We change the meaning of `--!nolint` so that it can occur at any position in the file.  We change its meaning to indicate that, from its position on, the named lints should be disabled.

We also add a new hot comment `--!poplint` (name TBD) which has the effect of undoing the prior `--!nolint`:

```luau
blah blah

--!nolint A
lint A is disabled here

--!nolint B, C
A, B, and C are disabled here
--!poplint

Only A is disabled here
--!poplint

-- back to default settings
```

nnullcolumn:
> If we built this system using comments as an extension to `--!nolint`, we could implement it entirely within the linter itself. No changes to the language would be necessary.
> 
> How about this:
> 
> We change the meaning of `--!nolint` so that it can occur at any position in the file. We change its meaning to indicate that, from its position on, the named lints should be disabled.
> 
> We also add a new hot comment `--!poplint` (name TBD) which has the effect of undoing the prior `--!nolint`:
> 
> ```luau
> blah blah
> 
> --!nolint A
> lint A is disabled here
> 
> --!nolint B, C
> A, B, and C are disabled here
> --!poplint
> 
> Only A is disabled here
> --!poplint
> 
> -- back to default settings
> ```

`--!` comments historically have only been used at the top of a module AFAIK. That aside, this feels very unsatisfying:

- refactoring code can involve cutting and pasting things here to there, which might miss comments

- you can't easily do things like `@nolint fun()`, which is kinda unfun. i don't want to add potentially multiple lines of comments that effectively serve to be visual noise

- forgetting to pop the lint would *really* hurt because it'd be more or less silent
  - there are some mitigations for this; we maybe able to warn people if they forget to pop a nolint but that in and of itself is kinda an example of something that could get awkward when multiple implementors are involved

- the range isn't entirely obvious and if you're skimming you might end up with confusing linter behavior; we shouldn't let these cross multiple lexical scopes IMO.

- we run into trouble if we ever want to extend the possible params for nolint

among many other things. I don't think comments do this job well on their own, or I would've opened with one of the comment-based alternatives I already listed. I'm very much in favor of not relying on mid-file hot comments for developer experience, especially as someone who has been unsatisfied with comment-based silencing in external linters and other languages.

Is extending the language a dealbreaker?

andyfriesen:
> Is extending the language a dealbreaker?

mmm.  I don't think so.

`#[allow(X)]` really is quite nice in Rust.

docs/syntax-attributes-nolint.md
```diff
@@ -0,0 +1,78 @@
...
+silencing at most need to specify exclusions, so values of `true` seem like they'd become repetitive when compared to a
+declaration with a larger scope. this would also be cumbersome to type, and doesn't allow people to easily mask all
+lints as they could with `@nolint`.
+
+- Name it `nowarn` instead. This is disconnected from the file-wide flag `--!nolint`, which could be interpreted as a
+boon or a flaw. However, users might also conflate `nowarn` with runtime warnings, and we want to avoid confusing
+people.
+
+- Use a comment instead. The only direct benefit to this is compatibility when running in a lua runtime, which seems
+moot at this point as many luau features today (and from the start) are exclusive with lua's syntax expectations.
```
andyfriesen:
```suggestion
- Use a comment instead.  This would make the implementation a fair
bit simpler, as it would mean that all of the logic could be confined
to the linter itself, but the ergonomics are poor:  Because comments
are blind to the structure of the source code, there is nothing to
prevent a developer from inadvertently disabling a warning for more
than the intended span.  It's quite nice to be able to write a small
annotation at the top of a declaration or statement to disable a
warning for that entire construct with no need for an extra directive
at the end.
```

docs/syntax-attributes-nolint.md
```diff
@@ -0,0 +1,78 @@
...
+between tools this could lead to confusion. However, it seems like linting will eventually move to being the
+sole responsibility of `lute`, which would be able to identify registered lint names and emit a configurable warning in
+and of itself.
+
+- People writing code to be compatible with lua & luau would not be able to silence lints with an attribute, because
+`@attribute` syntax is a luau-specific feature.
+
+## Alternatives
+
+- `@lint` / `@[lint { name = true } ]`. This is more coherent with `.config.luau` files. most cases for inline lint
```
andyfriesen: Rust uses `allow` which seems a bit strange if you're reading it in a spec, but when you're reading actual code, `#[allow(unused)]` reads very clearly.

docs/syntax-attributes-nolint.md
nnullcolumn: lgtm :+1:

docs/syntax-attributes-nolint.md
nnullcolumn: i think i'll probably remove the alternative which is highlighted because it makes almost no sense. i'll also add an alternative for `allow`, but i think `@nolint` makes more sense in the context of luau because you wouldn't be forced to specify specific lint rules. `@[allow("LocalUnused")]` or similar would be nice, but `@allow call()` on its own reads in a way that irks me a bit.

SeloTheGreat:
I wonder if it'd be worth it to tightly scope this feature as _just-a-new-attribute_ without augmenting how attributes currently work, so this feature would have an easier time getting approved.

I definitely think `nolint` would be awesome, but this RFC feels like it's introducing two seperate things at once? I think a seperate RFC like gaymeowings "attributes everywhere" RFC (https://github.com/luau-lang/rfcs/pull/147) would be a better place to discuss about letting attributes on other code bits & pieces rather than just functions.

On that note, yeah, i do think that attributes on statements and expressions would be great too, i just thought a seperate RFC for that kind of thing would be better.

...though, if there is a reason that you decided to write the RFC this way, do let me know, thanks!

PR #240, opened 2026-08-11
RFC: Class Constructor Modifier

soyuta:
[Rendered](https://github.com/soyuta/luau-rfcs/blob/class-constructor-modifier/docs/class-constructor-modifier.md)

This RFC introduces a constructor modifier for declaring a class's constructor while allowing its name to follow existing Luau conventions.

```luau
class Car
	public IsDestroyed: boolean
	public Speed: number

	constructor function new(speed: number)
		self.IsDestroyed = false
		self.Speed = speed
	end
end

const car = Car.new()
```

<sub>Classes should not become difficult to understand or use, and I am concerned that the current direction may be making them unexpectedly more complex.
Even if this proposal is not accepted, I still hope classes will move in a direction that keeps them simple and intuitive.</sub>

### DIFF
--- a/docs/class-constructor-modifier.md
+++ b/docs/class-constructor-modifier.md
@@ -1,0 +1,151 @@
+# Constructor Modifier
+
+## Summary
+
+To provide constructors that are compatible with existing Luau conventions, this RFC proposes the following syntax:
+
+```luau
+class Car
+	public IsDestroyed: boolean
+	public Speed: number
+
+	constructor function new(speed: number)
+		self.IsDestroyed = false
+		self.Speed = speed
+	end
+end
+```
+
+This proposal introduces a `constructor` modifier, similar to the `public` modifier.
+
+## Motivation
+
+Using the `__init` metamethod as a constructor may be difficult for beginners, since the connection between `__init` and `new` is not immediately intuitive.
+
+On the other hand, abandoning the long-standing Luau convention of constructors such as `Class.new()` and only allowing classes to be instantiated through function-call syntax like `Class { prop = a }` would also be inconsistent with existing conventions.
+
+Furthermore, `Class.new()` is not the only constructor naming convention in Luau. 
+Native objects also use names such as `create`, as seen in `buffer.create()` and `vector.create()`.
+
+For this reason, forcing all constructors to be named `new` would not be appropriate either.
+
+Since Luau classes are being designed with inheritance in mind, constructors of superclasses must also be callable.
+
+Therefore, each class should have exactly **one** constructor, while the **name** of that constructor should be customizable.
+In most cases, `new` would be the preferred constructor name, but it should be possible to use another name, such as `create`, when appropriate.
+
+## Design
+
+Inside a constructor function, `self` is implicitly available, and the constructed instance is returned even without an explicit `return self`.
+
+```luau
+class Car
+	public IsDestroyed: boolean
+	public Speed: number
+
+	constructor function new(speed: number)
+		self.IsDestroyed = false
+		self.Speed = speed
+	end
+end
+
+const myCar = Car.new(50)
+```
+
+If inheritance is added in the future, `super` could also be implicitly available as the superclass constructor and could be called directly.
+`super` refers to the superclass's constructor regardless of that constructor's declared name.
+
+```luau
+class Car
+	public IsDestroyed: boolean
+	public Speed: number
+
+	constructor function new(speed: number)
+		self.IsDestroyed = false
+		self.Speed = speed
+	end
+end
+
+class ElectricCar extends Car
+	public Capacity: number
+
+	constructor function new(capacity: number)
+		super(math.max(50 - capacity * 2, 0))
+		self.Capacity = capacity
+	end
+end
+```
+
+If a different constructor name is desired, it can be specified using `constructor function <name>`:
+
+```luau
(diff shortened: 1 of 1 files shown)


InfraredGodYT:
I've got a couple questions for this RFC:

1. Why not just reserve `constructor` as a function like you do with `super`?
2. Have you considered using `static` for factory methods instead of making `constructor` a modifier?

soyuta:
> I've got a couple questions for this RFC:
> 
> 1. Why not just reserve `constructor` as a function like you do with `super`?
> 2. Have you considered using `static` for factory methods instead of making `constructor` a modifier?

1. Why not just reserve `constructor` as a function like you do with `super`:

If `constructor` is a reserved function name, then we lose the ability to use conventional constructor names like `new` or `create`.

There was already a lot of pushback against the `__init` -> `.new` approach in #210 because it wasn't considered intuitive. 
If we avoid that, we'd basically have to construct classes only through function-call syntax like `Class()`, but there is already so much Luau code following conventions like `.new` and `.create` that I don't think we should simply throw those conventions away.

If we're going to use function-call syntax anyway, I think something more like the `__init` metamethod proposed in #210 would make more sense.

But if we want to preserve conventions like `.new()` and `.create()`, I think it's better to let the constructor name be written explicitly, as in `constructor function new`.

2. Have you considered using `static` for factory methods instead of making `constructor` a modifier:

If constructors are implemented as static functions like they are now, inheritance becomes difficult to design around.

```luau
class Car
	public IsDestroyed: boolean
	public Speed: number

	function new(speed: number)
		return Car { IsDestroyed = false, Speed = speed }
	end
end

class ElectricCar extends Car
	public Capacity: number

	function new(capacity: number)
		const self = Car.new(math.max(50 - capacity * 2, 0))
		-- How do we know whether this Car is intended to become an ElectricCar, 
		-- or whether it should remain just a Car?
		-- We'd have no choice but to report an error here.
		self.Capacity = capacity 
		return self
	end
end
```

A normal static function has no way to express that `Car.new(...)` is constructing the base portion of an `ElectricCar`, rather than simply returning a `Car`.

That's why I think constructors need semantics beyond those of an ordinary static factory method.
Of course, there can still be static functions like `fromFile`, `fromString`, or `Angles` that build on the primary constructor.

InfraredGodYT:
> If `constructor` is a reserved function name, then we lose the ability to use conventional constructor names like `new` or `create`.
> 
> There was already a lot of pushback against the `__init` -> `.new` approach in #210 because it wasn't considered intuitive. If we avoid that, we'd basically have to construct classes only through function-call syntax like `Class()`, but there is already so much Luau code following conventions like `.new` and `.create` that I don't think we should simply throw those conventions away.
> 
> If we're going to use function-call syntax anyway, I think something more like the `__init` metamethod proposed in #210 would make more sense.
> 
> But if we want to preserve conventions like `.new()` and `.create()`, I think it's better to let the constructor name be written explicitly, as in `constructor function new`.

Standard conventions should be optional instead of enforced, and I believe that makes the function-call syntax superior to using/reserving a `new` or `create` method when instantiating a class.

If `constructor` is a reserved function name and you allow `static` factory methods like that utilize the primary constructor, then using conventional constructor names like `new` or `create` becomes a choice instead of being mandatory. While I like this RFC much more than #210, they both share the same problem of enforcing a specific pattern without a strong rationale. Common patterns shouldn't be treated like they're absolute.

SeloTheGreat:
I would advise to rewrite this RFC to fit into the [classes-mega-rfc](https://github.com/luau-lang/rfcs/blob/classes-mega-rfc/docs/classes-mega-rfc.md) document and open a pull request to the [classes-mega-rfc branch](https://github.com/luau-lang/rfcs/tree/classes-mega-rfc) specifically. Since it would be quite difficult to get this accepted into the master branch _before_ the classes mega-RFC is merged.

soyuta:
> I would advise to rewrite this RFC to fit into the [classes-mega-rfc](https://github.com/luau-lang/rfcs/blob/classes-mega-rfc/docs/classes-mega-rfc.md) document and open a pull request to the [classes-mega-rfc branch](https://github.com/luau-lang/rfcs/tree/classes-mega-rfc) specifically. Since it would be quite difficult to get this accepted into the master branch _before_ the classes mega-RFC is merged.

Actually I'm fine with the Mega RFC as it stands.
When I wrote this RFC, I wasn't really expecting it to get merged in the first place. 
I mostly just wanted to put my thoughts out there.

Things felt like they were getting too complicated and too constrained by convention. 
I just wanted to make the point that Luau shouldn't become some impossibly difficult language full of alien-looking syntax. 
I wanted to push for keeping things as simple as possible

In the end, it seems like Roblox has decided to bake the existing naming conventions in Luau and Roblox as a whole directly into the language itself.

Considering that one of the community's complaints was that naming conventions shouldn't become part of the language contract, I guess that's a little disappointing. 
But if that's the direction they've decided to go in, I'm in favor of where things are heading now.

I still wonder whether making every class use `.new()` just to reduce the ambiguity between `__init` and `.new` is really the best solution.

But honestly, I can't think of a better approach there either, and it doesn't seem overly complicated.

soyuta:
Now that the Mega RFC has been added, I don't think there's a strong enough reason to prefer a `constructor` modifier over the current approach, so I'll close this PR.

PR #242, opened 2026-08-13
Classes mega-RFC

skberkeley:
[Rendered](https://github.com/luau-lang/rfcs/blob/classes-mega-rfc/docs/classes-mega-rfc.md)

Per the discussion on OSS Discord, this collates the existing PRs into a single "mega-RFC" that once merged, will serve as the source of truth for how the final feature should behave.

Future class-related features should be proposed as PRs against this branch, and once those discussions have settled, this PR will be merged.

I've taken the liberty of updating the existing RFCs slightly as I collated so that the full document is internally consistent.

The mega-RFC reflects the team's stance on what classes with just the features discussed should look like. This stance may change as we consider additional features, and PRs against this branch will update existing sections if necessary.

Future discussions on the general design of classes, constructors, and implementation inheritance should take place on this thread.


TenebrisNoctua:
Since we're moving discussions to this thread, I'd like to link some of my previous comments that I believe still holds important value for constructors:

https://github.com/luau-lang/rfcs/pull/210#issuecomment-5227827477
https://github.com/luau-lang/rfcs/pull/210#issuecomment-5052728389

skberkeley:
> Since we're moving discussions to this thread, I'd like to link some of my previous comments that I believe still holds important value for constructors:
> 
> [#210 (comment)](https://github.com/luau-lang/rfcs/pull/210#issuecomment-5227827477) [#210 (comment)](https://github.com/luau-lang/rfcs/pull/210#issuecomment-5052728389)

To summarize, I believe the two main points of concern are:
- Using `.new()` as the blessed path for constructing new class instances.
- Allowing constructors to be inheritable.

I can't speak more on the first point beyond saying that this is a hard requirement from internal stakeholders, but maybe @andyfriesen can.

On the second, we share your concern, and have included your suggestion that child classes must define constructors if their parents do so! This is specified in [Constructors - Typechecking](https://github.com/luau-lang/rfcs/blob/classes-mega-rfc/docs/classes-mega-rfc.md#typechecking).

TenebrisNoctua:
> > Since we're moving discussions to this thread, I'd like to link some of my previous comments that I believe still holds important value for constructors:
> > [#210 (comment)](https://github.com/luau-lang/rfcs/pull/210#issuecomment-5227827477) [#210 (comment)](https://github.com/luau-lang/rfcs/pull/210#issuecomment-5052728389)
> 
> To summarize, I believe the two main points of concern are:
> 
>     * Using `.new()` as the blessed path for constructing new class instances.
> 
>     * Allowing constructors to be inheritable.
> 
> 
> I can't speak more on the first point beyond saying that this is a hard requirement from internal stakeholders, but maybe @andyfriesen can.
> 
> On the second, we share your concern, and have included your suggestion that child classes must define constructors if their parents do so! This is specified in [Constructors - Typechecking](https://github.com/luau-lang/rfcs/blob/classes-mega-rfc/docs/classes-mega-rfc.md#typechecking).

I would really want the team to present all of these potential pain points in implementation and syntax to internal stakeholders. We have seen this play-out before, and it is understandable that the team would consider their requirements first, because, well, Roblox is the primary customer, but this requirement benefits no one. 

There can be workarounds found for those who wish to easily have a method such as `.new()` available for them. With the current syntax, it is significantly easier to define a static new method inside the class and construct an instance of a class.

I've also proposed the `new` keyword as an alternative solution, it brings in similar semantics to how objects are created in C++/C#, and removes the reserving requirement.

The team could also consider implementing C++ like template classes, with generic support. C++'s template classes are quite useful, and if brought to Luau in a way that doesn't increase the syntax complexity too much, it would be an amazing alternative. The user would create a class from a template that contains `.new()`, and move on.

I would really suggest that the team listens the community on this one, because I believe community will use classes more than anyone else. If this requirement stays, then I believe the community backlash and confusion will be imminent.

For the second point, Andy did update the previous RFC to include an additional rule about that, so that was great.

docs/classes-mega-rfc.md
```diff
@@ -0,0 +1,624 @@
...
+
+This is a really big feature that has lots of moving parts!
+
+We need to introduce multiple new contextual keywords: `class` and `public` to start and `private` later.  We also introduce two new top types `object` and `class`.
+
+Allowing code to grab unbound method references (ie `local m = o.someMethod`) seems risky because it opens the doorway to a lot of difficult-to-optimize dynamism, but it also makes a bunch of nice things like `pcall` work exactly the way developers expect.  We're making the bet here that this does not materially affect our ability to optimize more mundane attribute access or method calls.
+
+The word `class` is doing triple duty under this RFC: It is a contextual keyword, the name of a top-level library, and the name of the top type for class objects.
+
+Object oriented codebases tend to have far more cyclic dependencies between modules because every piece of data is also coupled to a whole bunch of functions that operate on that data.  We are probably going to have to work out a way to relax the restrictions on cyclic module imports.
```
andyfriesen:
```suggestion
Object oriented codebases tend to have far more cyclic dependencies between modules because every piece of data is also coupled to a whole bunch of functions that operate on that data.  We'll also build out better support for cyclically-dependent modules.  See https://github.com/luau-lang/rfcs/blob/master/docs/support-for-cyclic-requires.md
```

docs/classes-mega-rfc.md
```diff
@@ -0,0 +1,624 @@
...
+    public x: number
+    public y: number
+end
+
+class ThreeDPoint extends Point
+    public z: number
+end
+
+local threedpoint = ThreeDPoint.new { x = 0, y = 0, z = 0 }
+local erroneous = ThreeDPoint.new { z = 1 } -- Type error.  x and y are uninitialized.  What their values are is undefined behavior.
```
andyfriesen:
```suggestion
local erroneous = ThreeDPoint.new { z = 1 } -- Type error.  x and y are uninitialized.  They will be nil at runtime.
```

docs/classes-mega-rfc.md
```diff
@@ -0,0 +1,624 @@
...
+
+Classes are still constructed using the static method `.new`.
+
+Let `T` be a class object that defines a constructor.  When `T.new(...args)` is invoked with any arguments, the following happens:
+
+1. A fresh, uninitialized instance of `T` is allocated.  We'll call it `t` here.  All of its fields are initially `nil` irrespective of any type annotations.  
+2. `T.__init(t, ...args)` is invoked.  
+3. `t` is produced as the result of the expression.
+
+To make explicit the other case, let `T` be a class object that does not contain a user-defined constructor.  It is illegal to invoke `T.new` without any arguments. When `T.new(arg, ...args)` is invoked with one or more arguments, the following happens:
```
andyfriesen: Something this framing doesn't mention is that the default constructor is an ordinary function that can be explicitly called.

docs/classes-mega-rfc.md
```diff
@@ -0,0 +1,624 @@
...
+
+Constructors add more complexity to the language.  Some classes can be initialized via `T.new { x = x, y = y}` syntax and others must be initialized with different kinds of arguments.
+
+We intentionally only check that fields are initialized in type checking.  This means that our runtime still has to be able to cope with fields that have been left uninitialized.
+
+Some developers will feel inconvenienced by the restrictions on uninitialized class fields.  The current rules do not, for instance, permit a developer to write a helper function that partially (or completely) initializes a new class instance.
+
+### Alternatives
+
+We could follow in Python's footsteps and do away with the default `T{}` constructor, but this means that developers have to write a lot of dull code in the typical "plain old data" case:
```
andyfriesen:
```suggestion
We could follow in Python's footsteps and do away with the default `T.new {}` constructor, but this means that developers have to write a lot of dull code in the typical "plain old data" case:
```

docs/classes-mega-rfc.md
```diff
@@ -0,0 +1,624 @@
...
+
+Class objects behave like class instances in most ways, but are always `const` and frozen.
+
+Taking references to class methods via `ClassName.method` syntax is allowed so that classes can easily compose with existing popular APIs:
+
+```luau
+local n = pcall(SomeClass.getName, someClassInstance)
+```
+
+To construct an instance of a class, call the class object's `.new` static mehtod.  It accepts a single argument: a table-like value that contains initial values for all the fields.  While it will typically be most useful to pass a table literal to this function, that isn't the only use.  For example, any class object can be shallowly cloned by passing it to its class constructor: `local clone = MyClass.new(original)`
```
andyfriesen:
```suggestion
To construct an instance of a class, call the class object's `.new` static method.  It accepts a single argument: a table-like value that contains initial values for all the fields.  While it will typically be most useful to pass a table literal to this function, that isn't the only use.  For example, any class object can be shallowly cloned by passing it to its class constructor: `local clone = MyClass.new(original)`
```

InfraredGodYT:
> Since we're moving discussions to this thread, I'd like to link some of my previous comments that I believe still holds important value for constructors:
> 
> [#210 (comment)](https://github.com/luau-lang/rfcs/pull/210#issuecomment-5227827477) [#210 (comment)](https://github.com/luau-lang/rfcs/pull/210#issuecomment-5052728389)

I agree. The RFC states the motivation behind this by saying:

> The choice to construct instances via a .new static function is motivated by Lua libraries that effect objects. We sacrifice the ability for classes to define their own .new method but in exchange, code using classes looks more consistent with what came before.

This is really just enforcing a status quo rather than guaranteeing consistency due to the drawbacks of making a `.new` method mandatory. Luau’s own native `buffer` and `vector` libraries using `.create` (possibly because of coroutines) also somewhat undermine the motivation for this constructor syntax since this is a Luau feature that isn’t limited to Roblox.

docs/classes-mega-rfc.md
```diff
@@ -0,0 +1,655 @@
...
+
+### Design
+
+Class definitions are a block construct.  They can only be written at the topmost scope.  `export class X` is allowed.
+
+Defining two classes with the same name in the same module is forbidden.
+
+Within a class block, two declarations are allowed: Fields and methods.
+
+For now, fields are introduced with the new `public` keyword.  We also plan to eventually offer private fields, but it is sufficiently complex that it merits its own RFC.
```
andyfriesen:
```suggestion
For now, fields are introduced with the new `public` keyword.  Private fields will be described in their own section of this document when they are ready.

Class fields are not strictly required to have an annotation, but it is very strongly encouraged.  Unannotated fields will have type `any` and strict mode will flag the field with a warning.
```

funwolf7:
The main concern I have with `new` being the forced constructor name is that in certain cases where there is a need to enforce invariants, it almost feels like there *wouldn't* be a way to make a public `new` constructor.

Since there is only 1 constructor, when designing a class you have to make sure that constructor really counts. If there are any static factory methods, this 1 constructor must support constructing a class in any way the factory functions support constructing the class. If there are no invariants and all properties are public, this is fine, since the default constructor will work for this, but when you start needing invariants (or if private properties become a thing), this becomes problematic.

If you just make a simple constructor that allows setting the properties in any way (or use the default constructor), invariants become difficult to enforce. Perhaps you could enforce the invariants through assertions in the constructor, but the invariants may be difficult or slow to check without additional context from the factory methods, and you might not want some properties to even be settable from outside of public code. It would just quickly become a mess as you add more factory methods, each with their own rules for what is allowed.

So, the simple solution one might do is to instead just make the constructor private but powerful and with no enforcement of invariants (obviously private doesn't exist right now, but we are looking to the future). Then, invariants can be enforced in the factory methods where they have full context, and the factory methods can construct arbitrary class objects however they need. This is very similar to what Rust requires you to do.

However, because we have made the constructor private, we are now blocked from using a constructor named `new` in the public API. So, ironically, we are now forced *away* from the standard of using `new` as a constructor. That's my biggest issue.

Languages like C++ and Java bypass this problem by allowing you to overload a private, more powerful constructor. That isn't really an option in Luau, however.

Kenji-Shore:
This might be too little too late since the discussion around inheritance has pretty much settled down, but is there any chance the concept of `abstract` classes could make its way in? The usefulness of `abstract` comes from being able to constrain the type of any derived classes.

```lua
abstract class Base
   abstract function method(self, x: number|string): {}?
end

class ChildOne extends Base
    function method(self, x: number|string?): {x: string}
        return {x="hello"}
    end
end

class ChildTwo extends Base
    function method(self, x: number|string?): {x: string}
        return {x="other"}
    end
end

const function operateOnClassObjectsWithMethod(object: Base)
   object:method("stuff")
end
```
The example function accepts a parameter `object: Base` which could be ChildOne or ChildTwo (but not Base itself); we know that `method(self, x: number|string): {}?` is guaranteed to exist in `object`.

`abstract` could be a standalone keyword that also designates the class as `open`, or alternatively the two keywords could be used in conjunction: `abstract open`

---

If `abstract` is off the table, would it instead be possible to create a type alias that is denoted to be a class type which defines field and method types, and then allow classes to extend off of said type alias (aka how classes `implements` interfaces in typescript). 

The other aspect of `abstract` that is not covered by the aforementioned interface-like functionality would be that `abstract` classes can also contain non-abstract methods, which derived classes inherit normally. The lack of `abstract` communicates that all classes are supposed to be valid classes that can be initialized, which may not always be the case (such as the common use case where a superclass defines methods that are used in its derived classes, but the superclass is not functional on its own and would not make sense to be instantiated).

I know the whole inheritance thing is pretty much settled at this point, but I feel like existing discussions overlooked a lot of the key functionality that `abstract` provides which the current proposed classes have no answer for.

skberkeley:
Hello! We've had both interface inheritance and marking individual methods as `open` or `inheritable` or `abstract` floated in the past. To avoid taking on too much complexity, we decided that they are not necessary for this initial version of classes, but we're open to discussing them in the future!

TenebrisNoctua:
> Hello! We've had both interface inheritance and marking individual methods as `open` or `inheritable` or `abstract` floated in the past. To avoid taking on too much complexity, we decided that they are not necessary for this initial version of classes, but we're open to discussing them in the future!

I'd also love to see abstract classes eventually. I personally think they should be a part of the classes syntax rather than a new "interface" syntax. Would make things easier and more consistent.

Though comes the question, if we're going to use `extends` to implement abstract classes, then what about multi-interface inheritance? Should that be an exception?

Kenji-Shore:
> Hello! We've had both interface inheritance and marking individual methods as `open` or `inheritable` or `abstract` floated in the past. To avoid taking on too much complexity, we decided that they are not necessary for this initial version of classes, but we're open to discussing them in the future!

Would it be possible to include `abstract` in the writeup under a section about possible future expansions? I think `abstract` carries a lot of key functionality that there is no equivalent to currently, and it feels like a natural, intuitive extension of what classes can do. I would hate to see there be preemptive design/implementation decisions made that lock us out of adding `abstract` in the future.

---
Also, another request I have is:
I know the discussion around constructors methods (__init(self) and Class.new()) are still somewhat up in the air, but I noticed that there hasn't been much discussion about **destructors**. I know technically destructors could be implemented by the user, but I think it would be valuable to have a consistent format for how destructors work across all classes, rather than it being different case-by-case depending on what destructor/cleanup method names people choose in their own code.

For example, following in the footsteps of Python (since we are already using __init), we could have a destructor method called __del, which would be triggered by calling Class:Destroy() (similar to how Class.new() routes to __init()). Inheritance would work similarly to __init, except whereas in __init the superclass must be called first, in __del the superclass must be called last.

This would provide a consistent standard for how to interact with classes, with .new() for creation and :Destroy() for deletion. One other thing to note, it would be nice if :Destroy() could lock the class object at the very end, causing any attempts to interact with the object to error. Without this, it is possible for code to have an unintended stray reference to a "deleted" object and continue interacting with said object without erroring, leading to unintended consequences.

lucaskiforishin:
> Hello! We've had both interface inheritance and marking individual methods as `open` or `inheritable` or `abstract` floated in the past. To avoid taking on too much complexity, we decided that they are not necessary for this initial version of classes, but we're open to discussing them in the future!

I think method-level `open` should be reconsidered before the class inheritance semantics are finalized. Currently `open class` establishes an opt-in model for inheritance, but methods are always able to be overridden. I think it would be much cleaner and more consistent for overriding methods to also be opt-in.

```luau
open class Widget
    function Destroy(self)
        -- inherited, but cannot be overridden
    end

    open function Update(self)
        -- intentional extension point
    end
end

class MyWidget extends Widget
    function Destroy(self)
        -- error: Destroy is not open
    end
    
    function Update(self)
        -- valid: Update is open
    end
end
```

`open class` and `open function` answer two different questions: whether the class can be extended, and whether a particular implementation can be overridden. An open class should not need to contain any open functions; it may simply be intended as a base class whose implementations are all closed. Conversely, an open function must belong to an open class, since there would be no point in making a function able to be overridden in a class that can't be extended.

If methods are initially able to be overridden by default, introducing opt-in overriding later would either break existing class hierarchies (not an option) or require another modifier like `final` to opt-out of overriding. That would leave us with `open class` for opt-in inheritance but `final function` for opt-out overriding rather than having the same `open` model at both levels.

This would provide a much cleaner foundation if abstract classes and methods are added later. `open` and `abstract` could naturally form two levels of the same model: an `open function` *can* be overridden, while an `abstract function` *must* be implemented by a concrete subclass. This would make `abstract` a natural extension of the existing `open` semantics.

Rob07erto:
Im new to luau how would I install and link Luau into vscode. I wanna try out the first experiemental additions of classes. (Im on windows)

PR #243, opened 2026-08-13
[meta] Add a `Prior Art` section to the RFC template

hgoldstein:
Like most engineering, language design heavily benefits from looking at prior art, and the _vast_ majority of language features have prior art in a widely used programming language:
- Having similar features look and behave similarly helps widen the pit of success: it's good that a `while` loop in Luau and one in JavaScript work roughly the same.
- You can dodge known pitfalls with commonly implemented features: as of this writing we are using other implementations of classes to prove out the design for classes in Luau.
- Other languages with extensive RFC processes can provide hints and reasoning for direction on _our_ RFCs, [for example the FAQ about why Javascript uses the `#` sigil to mean `private`](https://github.com/tc39/proposal-class-fields/blob/f372de4f4f5a245203fbb6cf5b461ed62d34e27c/PRIVATE_SYNTAX_FAQ.md#whats-this-about-).

This can also apply to prior art for library functions. If we're adding a new `math` function, do other languages standard libraries have the same functions? The same constants?

This is worth adding a section to push _all_ RFC writers in the right direction.

### DIFF
--- a/TEMPLATE.md
+++ b/TEMPLATE.md
@@ -19,3 +19,8 @@ Why should we *not* do this?
 ## Alternatives
 
 What other designs have been considered? What is the impact of not doing this?
+
+# Prior Art
+
+* Do other programming languages have similar features? What do those look like? How do they compare and contrast to this design? 
+* What supporting features do those languages have that might _not_ be included in this design?
--- a/TEMPLATE.md
+++ b/TEMPLATE.md
@@ -20,7 +20,7 @@ Why should we *not* do this?
 
 What other designs have been considered? What is the impact of not doing this?
 
-# Prior Art
+## Prior Art
 
 * Do other programming languages have similar features? What do those look like? How do they compare and contrast to this design? 
 * What supporting features do those languages have that might _not_ be included in this design?
--- a/TEMPLATE.md
+++ b/TEMPLATE.md
@@ -24,3 +24,4 @@ What other designs have been considered? What is the impact of not doing this?
 
 * Do other programming languages have similar features? What do those look like? How do they compare and contrast to this design? 
 * What supporting features do those languages have that might _not_ be included in this design?
+* Are there other similar features or libraries in Luau already? How does this feature align with _those_ features in terms of naming, syntax, and semantics?
--- a/TEMPLATE.md
+++ b/TEMPLATE.md
@@ -22,6 +22,6 @@ What other designs have been considered? What is the impact of not doing this?
 
 ## Prior Art
 
-* Do other programming languages have similar features? What do those look like? How do they compare and contrast to this design? 
-* What supporting features do those languages have that might _not_ be included in this design?
-* Are there other similar features or libraries in Luau already? How does this feature align with _those_ features in terms of naming, syntax, and semantics?
+* Do other programming languages have similar features? What do those look like? How do they compare and contrast to this design? Are there unique constraints for Luau that interact with said design?
+* What supporting features do those languages have that might _not_ be included in this design? If we're adding something feature A, are there features B, C, and D that are often used with A?
+* Are there other similar features or libraries in Luau already? How does this feature align with _those_ features in terms of naming, syntax, and/or semantics?


InfraredGodYT:
I never thought about an approach like this, but I could see it being widely used across future RFCs.

bradsharp:
Supportive of us adding this. We should also include a point about the existing language design being a key piece of prior art. For example if we introduced a new global called `array` then you would create one through `array.create` for consistency with `table.create` or `coroutine.create` rather than calling it something like `array.init`.

TEMPLATE.md
```diff
@@ -19,3 +19,9 @@ Why should we *not* do this?
 ## Alternatives
 
 What other designs have been considered? What is the impact of not doing this?
+
+## Prior Art
+
+* Do other programming languages have similar features? What do those look like? How do they compare and contrast to this design? 
+* What supporting features do those languages have that might _not_ be included in this design?
```
nnullcolumn: i think "supporting features" should be defined better or replaced with "related features" so that people are less likely to take mental shortcuts (maybe?)

TEMPLATE.md
```diff
@@ -19,3 +19,9 @@ Why should we *not* do this?
...
+* Are there other similar features or libraries in Luau already? How does this feature align with _those_ features in terms of naming, syntax, and semantics?
```
nnullcolumn: > naming, syntax, and semantics

i would replace with "naming, syntax, or semantics" or "naming, syntax, and/or semantics"

TEMPLATE.md
```diff
@@ -19,3 +19,9 @@ Why should we *not* do this?
 ## Alternatives
 
 What other designs have been considered? What is the impact of not doing this?
+
+## Prior Art
+
+* Do other programming languages have similar features? What do those look like? How do they compare and contrast to this design?
```
nnullcolumn: "How do they compare and contrast to this design" might invite very shallow comparisons especially so late into the body. Maybe it'd be better to include this section below Motivation, above Design, and rework the wording to something along the lines of "What design considerations are unique to luau or that feature which are different here?" That way we flow from Motivation -> Similar Designs + Declaring Unique Constraints -> Design

nnullcolumn: > Do other programming languages

I'd be tempted to change this to "other software projects", prior art can come from almost anywhere

TEMPLATE.md
nnullcolumn: re: languages -> projects(?)

nnullcolumn (commented):
are there any examples of other projects which have this as part of their RFC process? /j

i am in support of this idea, but i think we could probably deliberate a bit over the organization and reading flow

TEMPLATE.md
hgoldstein: I see your angle, but I mean programming languages. This section is meant to be a guard against completely untethered design: if you are doing something no other programming language has done, or you're doing it _in_ a way no other language has done, you want a damn good reason. I don't think you're going to get that kind of validation from other software projects.

TEMPLATE.md
hgoldstein: I think people will take mental shortcuts both ways, but I'll extend the language.

TEMPLATE.md
hgoldstein: I think the idea of identifying unique design considerations is important, but my worry about that ordering is that it encourages a terse prior art section. I also think it makes sense to live along side Alternatives (e.g.: "an alternative is copy-pasting JavaScript's design"). I'll extend the language but keep its spot in the template.

nnullcolumn: > my worry about that ordering is that it encourages a terse prior art section.

yeah, understandable. Just as a reader, I think I'd prefer to read

- "similar designs"

- "here's what's different about our situation"

- "here's what we're actually proposing"

in that order. reading a design proposal followed by design considerations feels like it invites unnecessary repetition people might not otherwise write in to begin with

nnullcolumn: understandable.

aatxe (approved):

PR #244, opened 2026-08-14
Tweak __eq rule in the face of inheritance

skberkeley:
We realized that the `__eq` semantics for classes could be overly restrictive in the face of inheritance. This PR loosens it to align more closely with Lua's behavior.

This PR amends the existing classes mega-RFC. If merged, it does not mean that the classes mega-RFC has been merged too, but rather that the mega-RFC is now under consideration with this change.

### DIFF
--- a/docs/classes-mega-rfc.md
+++ b/docs/classes-mega-rfc.md
@@ -110,7 +110,7 @@ type(inst) == "object"
 typeof(inst) == "object"
 ```
 
-Comparisons between object instances is the same as with tables: If `__eq` is not defined, object comparisons use physical (pointer) equality.  `__eq` is only invoked if both operands are the same type.
+Comparisons between object instances is the same as with tables: If `__eq` is not defined, object comparisons use physical (pointer) equality.  As with Lua, `__eq` is only invoked if both operands share the same definition.
 
 ### The `class` library
 
@@ -330,6 +330,28 @@ Class instances are the same:
 ```luau
 local c = Child.new({})
 c.foo() // OK
+```
+
+Above, we mentioned that `__eq` between two class instances is only invoked if both operands share it. This rule is still followed in the presence of inheritance:
+
+```luau
+open class Point
+    x: number
+    y: number
+
+    function __eq(self, other)
+        return self.x == other.x and self.y == other.y
+    end
+end
+
+class NamedPoint extends Point
+    name: string
+end
+
+local p = Point.new(0, 0)
+local np = NamedPoint.new(0, 0, "Bob")
+
+p == np -- invokes Point.__eq and evaluates to true, not a proof
 ```
 
 ### Declaration Order
--- a/docs/classes-mega-rfc.md
+++ b/docs/classes-mega-rfc.md
@@ -312,7 +312,7 @@ end
 
 ### Name resolution
 
-Methods defined on base classes are accessible via the derived class.
+Methods (including metamethods) defined on base classes are accessible via the derived class, with the exception of `__init`, which has its own set of rules described in the Constructors section.
 
 ```luau
 open class Base
@@ -351,7 +351,7 @@ end
 local p = Point.new(0, 0)
 local np = NamedPoint.new(0, 0, "Bob")
 
-p == np -- invokes Point.__eq and evaluates to true, not a proof
+print(p == np) -- invokes Point.__eq and evaluates to true, not a proof
 ```
 
 ### Declaration Order


andyfriesen:
A valid point in this design space would be to require that the two operands have exactly the same concrete type, else the result is unconditionally `false`.  This would be completely consistent with how `__eq` works with the rest of Luau, including scenarios involving metatable-based inheritance.

In the event that this behaviour isn't desired, the developer can simply write their own `:equal` method.

docs/classes-mega-rfc.md
```diff
@@ -332,6 +332,28 @@ local c = Child.new({})
...
+end
+
+class NamedPoint extends Point
+    name: string
+end
+
+local p = Point.new(0, 0)
+local np = NamedPoint.new(0, 0, "Bob")
+
+p == np -- invokes Point.__eq and evaluates to true, not a proof
```
nnullcolumn: this is a syntax error, we should wrap it in `print()` or `return` or something

also holy shit!! p = np proven by a luau lang rfc body!! this is a huge achievement

TenebrisNoctua:
> A valid point in this design space would be to require that the two operands have exactly the same concrete type, else the result is unconditionally `false`. This would be completely consistent with how `__eq` works with the rest of Luau, including scenarios involving metatable-based inheritance.
> 
> In the event that this behaviour isn't desired, the developer can simply write their own `:equal` method.

I quite agree with this. The requirement would be that the second argument would need to be an `object` from a class that inherits the class the `__eq` method is defined in. Though this does bring in some questions such as: "Will metamethods ever be inherited?"

docs/classes-mega-rfc.md
skberkeley: Will do!

skberkeley:
> A valid point in this design space would be to require that the two operands have exactly the same concrete type, else the result is unconditionally `false`. This would be completely consistent with how `__eq` works with the rest of Luau, including scenarios involving metatable-based inheritance.

This isn't totally accurate, the following prints `true` today:

```luau
local function comp(self, other)
  return self.x == other.x and self.y == other.y
end

local p = setmetatable({ x = 0, y = 0 }, { __eq = comp })

local np = setmetatable({ x = 0, y = 0, name = "Bob" }, { __eq = comp })

print(p == np)
```

andyfriesen:
I'm talking about cases of actual implementation inheritance using Lua metatables.

Pardon the wall of code:

```luau
local Point = {}

function Point.abs(self)
    return math.sqrt(self.x * self.x + self.y * self.y)
end

function Point.__eq(self, other)
    return self.x == other.x and self.y == other.y
end

function Point.__tostring(self)
    return `Point \{ x={self.x}, y={self.y} \}`
end

function Point.new()
    return Point {}
end

Point.__index = Point

function Point.__newindex(self, index, value)
    error("Cannot set property " .. index .. " of Point")
end

local PointMT = {}
function PointMT.__call(self, args)
    -- default initializers are evaluated here
    local x = if args.x == nil then 0 else args.x
    local y = if args.y == nil then 0 else args.y

    return setmetatable({x=x, y=y}, Point)
end

setmetatable(Point, PointMT)

-------

local NamedPoint = {}

function NamedPoint.__tostring(self)
    return `NamedPoint \{ x={self.x}, y={self.y}, name={self.name} \}`
end

NamedPoint.__index = NamedPoint
function NamedPoint.__newindex(self, index, value)
    error("Cannot set property " .. index .. " of NamedPoint")
end

local NamedPointMT = {}
function NamedPointMT.__call(self, args)
    local x = if args.x == nil then 0 else args.x
    local y = if args.y == nil then 0 else args.y
    local name = if args.name == nil then error("Missing argument name") else args.name

    return setmetatable({x=x, y=y, name=name}, NamedPoint)
end

NamedPointMT.__index = Point

setmetatable(NamedPoint, NamedPointMT)

print('-------')

local a = Point { x=2, y=3 }
local b = NamedPoint { x=2, y=3, name="The Named Point" }

print(a:abs(), b:abs())

print(if a == b then "a == b" else "a ~= b")
```

The important part is the comparison at the very end.  `a` and `b` do not have exactly the same metatable and so the metamethod is not invoked.

andyfriesen:
oh, I see.  Some details I'd overlooked:

The rule is just that the `__eq` metamethod itself has to be the same for both operands, and
The `__index` metaproperty is not used to find `__eq`.

PR #245, opened 2026-08-15
Typed Positional Table Properties

IFMNB:
This RFC series extends Luau's table type system with precise representations for statically known table keys and heterogeneous positional structures.


jLn0n:
the type number literals proposal looks similar to [this](https://github.com/luau-lang/rfcs/pull/203) proposal but your proposal integrates another dependency

aatxe:
Hi there, thanks for your RFCs. In order for them to be reviewed properly, can you please split them up into separate pull requests? We don't currently have a way to deal with three distinct proposals being made in one thread, and it seems like these proposals can all very reasonably be considered separately.

IFMNB:
This version of the RFC will be fragmented and reformulated into a more appropriate form, because it:

 1. does not solve the problems sufficiently, 
 2. does not accurately describe the affected things, 
 3. revision of the essence of the documents

I'll leave links to new versions later (maybe)

PR #247, opened 2026-08-19
Private class fields

skberkeley:
This PR introduces private fields for classes as a PR against the mega RFC (#242). 

We introduce a different syntax than the last private field PR, emphasize the issues with having just a `private` keyword, and de-emphasize performance/implementation considerations.

Enjoy!

### DIFF
--- a/docs/classes-mega-rfc.md
+++ b/docs/classes-mega-rfc.md
@@ -279,7 +279,7 @@ local threedpoint = ThreeDPoint.new { x = 0, y = 0, z = 0 }
 local erroneous = ThreeDPoint.new { z = 1 } -- Type error.  x and y are uninitialized.  They will be nil at runtime.
 ```
 
-Subclasses are forbidden from redeclaring fields declared in their superclasses. Such a redeclared field would need to type invariantly against the superclass field to maintain soundness anyway. Additionally, this reduces ambiguity for programmers coming from other languages, such as C++, where shadowed fields exist independently (i.e. `A::field` vs. `B::field`, where `B` subclasses `A`). This restriction appies only to public fields.
+Subclasses are forbidden from redeclaring fields declared in their superclasses. Such a redeclared field would need to type invariantly against the superclass field to maintain soundness anyway. Additionally, this reduces ambiguity for programmers coming from other languages, such as C++, where shadowed fields exist independently (i.e. `A::field` vs. `B::field`, where `B` subclasses `A`). This restriction applies only to public fields.
 
 A subclass can redefine a method present in its superclass. However, the method declared in the subclass must be a subtype of the method in the superclass.  That is, a child class can override a parent class method with a function that is more permissive than that of the base class method, but not less.
 
--- a/docs/classes-mega-rfc.md
+++ b/docs/classes-mega-rfc.md
@@ -724,6 +724,8 @@ Roblox itself has used this technique and found itself forced to permanently pre
 
 Methods and fields can both be declared private by using the new `private` keyword. Private property names must be prefixed with “`_`”. A private field can only be accessed by methods of that class. Private methods can only be called by methods of that class. Static methods can also be private.
 
+All fields must be declared with an access modifier, (at this time, either `public` or `private`).
+
 As is the case with public properties, using keywords as the names of private properties is forbidden.
 
 To avoid complications involved with the notion of a private metamethod or metatable and parsing ambiguity, declaring private property that begins with `__` is forbidden. Defining a private property named `_new` is also disallowed to maintain forwards compatibility with the case that we introduce private constructors.
--- a/docs/classes-mega-rfc.md
+++ b/docs/classes-mega-rfc.md
@@ -38,11 +38,11 @@ Defining two classes with the same name in the same module is forbidden.
 
 Within a class block, two declarations are allowed: Fields and methods.
 
-For now, fields are introduced with the new `public` keyword.  Private fields will be described in their own section of this document when they are ready.
+Public fields are introduced with the new `public` keyword.  Private fields are described in their own section of this document.
 
 Class fields are not strictly required to have an annotation, but it is very strongly encouraged.  Unannotated fields will have type `any` and strict mode will flag the field with a warning.
 
-Methods are introduced with the familiar `function` keyword.  `public function f()` is also permitted.
+Methods are introduced with the familiar `function` keyword.  `public function f()` is also permitted. Both syntaxes declare public methods.
 
 Methods defined on class objects can be accessed either via `Class.method()` or `instance:method()` syntax.
 
@@ -279,7 +279,7 @@ local threedpoint = ThreeDPoint.new { x = 0, y = 0, z = 0 }
 local erroneous = ThreeDPoint.new { z = 1 } -- Type error.  x and y are uninitialized.  They will be nil at runtime.
 ```
 
-Subclasses are forbidden from redeclaring fields declared in their superclasses. Such a redeclared field would need to type invariantly against the superclass field to maintain soundness anyway. Additionally, this reduces ambiguity for programmers coming from other languages, such as C++, where shadowed fields exist independently (i.e. `A::field` vs. `B::field`, where `B` subclasses `A`). In the case that private fields are added to classes, we expect this restriction to apply only to public fields.
+Subclasses are forbidden from redeclaring fields declared in their superclasses. Such a redeclared field would need to type invariantly against the superclass field to maintain soundness anyway. Additionally, this reduces ambiguity for programmers coming from other languages, such as C++, where shadowed fields exist independently (i.e. `A::field` vs. `B::field`, where `B` subclasses `A`). This restriction appies only to public fields.
 
 A subclass can redefine a method present in its superclass. However, the method declared in the subclass must be a subtype of the method in the superclass.  That is, a child class can override a parent class method with a function that is more permissive than that of the base class method, but not less.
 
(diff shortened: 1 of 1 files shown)


docs/classes-mega-rfc.md
```diff
@@ -279,7 +279,7 @@ local threedpoint = ThreeDPoint.new { x = 0, y = 0, z = 0 }
 local erroneous = ThreeDPoint.new { z = 1 } -- Type error.  x and y are uninitialized.  They will be nil at runtime.
 ```
 
-Subclasses are forbidden from redeclaring fields declared in their superclasses. Such a redeclared field would need to type invariantly against the superclass field to maintain soundness anyway. Additionally, this reduces ambiguity for programmers coming from other languages, such as C++, where shadowed fields exist independently (i.e. `A::field` vs. `B::field`, where `B` subclasses `A`). In the case that private fields are added to classes, we expect this restriction to apply only to public fields.
+Subclasses are forbidden from redeclaring fields declared in their superclasses. Such a redeclared field would need to type invariantly against the superclass field to maintain soundness anyway. Additionally, this reduces ambiguity for programmers coming from other languages, such as C++, where shadowed fields exist independently (i.e. `A::field` vs. `B::field`, where `B` subclasses `A`). This restriction appies only to public fields.
```
TenebrisNoctua: Small typo here.

TenebrisNoctua:
Honestly, this is so much better than the previous RFC, but still some problem points:

* I believe that instances of a class containing the private fields of the parent class is a bad solution.

Consider this example:

```lua
open class Point
    private _x: number
    private _y: number

    function __eq(self: Point, other)
        return self._x == other._x and self._y == other._y
    end
end

class ThreeDPoint extends Point
    private _z: number
end
```

This should normally fail. `_x` and `_y` should not remain on instances of `ThreeDPoint`, private should always mean private to a specific class, and never to any other class.

Classes are already opt-in to inheritance, so we can use `protected` for this instead:

```lua
open class Point
    protected x: number
    protected y: number

    function __eq(self: Point, other)
        return self.x == other.x and self.y == other.y
    end
end

class ThreeDPoint extends Point
    private _z: number
end
```

* Couldn't really understand from the RFC on what we're planning to do with the ability to not include an access specifier keyword for POD-like classes like this:

```lua
class Vector
    x: number
    y: number
end
```

We assume these properties are all `public`, but if we introduce a new field with an access specifier different than `public`, it creates ambiguity.

This could perhaps be solved with a parse-error:

```lua
class Vector -- Error: Must explicitly define every field with an access specifier
    x: number
    y: number
    private z: number
end
```

* Since we know that a class' private fields can never be accessed by its derived classes, as that behavior is reserved for `protected`, in that same fork, this becomes less of an ambiguity:

```lua
open class Base
    private x: number

    function compare(self, other)
        return self.x == other.x
    end
end

class Derived extends Base
    public x: number
end

local a = Base.new(...)
local b = Derived.new(...)

-- We can know that this will compare Base.x to Derived.x.
local res = a:compare(b)
```

In this case, we can know that `Derived` will never have the private field of `Base`, and thus compare method compares the private field of `Base` to the public field of `Derived`.

skberkeley:
Thank your feedback! I will do my best to address all your points, but please let me know if I miss any!
- This may come down to a matter of opinion, but we firmly believe that child class instances should contain the private fields of their parents. This is the norm for other object oriented languages, and is generally a key part of the abstraction barrier provided by inheritance. 
More concretely, if I am the author of the `Point` class, all I should have to care about when you compare an instance of `Point` is that you compare to another instance of `Point`. This comparison should work whether you compare to `ThreeDPoint` or any other class that has `Point` in its inheritance chain. For this to work, instances of these deriving classes need to contain the private fields of `Point`, even if the child classes themselves don't have a way to access them.
- I can add a clarification for this, but all fields must be declared with an access modifier (either `public` or `private` for now).

docs/classes-mega-rfc.md
skberkeley: oops, thanks for pointing that out!

TenebrisNoctua:
> Thank your feedback! I will do my best to address all your points, but please let me know if I miss any!
> 
>    * This may come down to a matter of opinion, but we firmly believe that child class instances should contain the private fields of their parents. This is the norm for other object oriented languages, and is generally a key part of the abstraction barrier provided by inheritance.
>       More concretely, if I am the author of the `Point` class, all I should have to care about when you compare an instance of `Point` is that you compare to another instance of `Point`. This comparison should work whether you compare to `ThreeDPoint` or any other class that has `Point` in its inheritance chain. For this to work, instances of these deriving classes need to contain the private fields of `Point`, even if the child classes themselves don't have a way to access them.
> 

You can achieve the same exact system using `protected`. No one is stopping you there. It's just a matter of explicitness, separating this behavior to make it explicit using `protected` resolves the ambiguity, and allows you to not reserve any kind of glyph for any field.

It also allows the derived class to gain control over every one of its fields, which is what should be. Adding hidden fields will confuse users, and will create weird scenarios just like the one we're trying to solve in this RFC.

TenebrisNoctua:
>    * I can add a clarification for this, but all fields must be declared with an access modifier (either `public` or `private` for now).

Yeah I don't mind this to be honest. Maybe some folks who like POD-classes may disagree, and perhaps a solution like in the fork I've mentioned can be made, but for now a restriction seems fine.

andyfriesen:
> This should normally fail. _x and _y should not remain on instances of ThreeDPoint, private should always mean private to a specific class, and never to any other class.

This proposal works nearly identically to how it works in C++, Java, and everything that's flowed from that.  If private fields weren't available (but hidden) in derived classes, inheritance wouldn't even work for any class that has a private field.

```luau
open class BasePoint
    private _x: number
    private _y: number
    function __init(self, x, y) ... end
    function length(self)
        return math.sqrt(self._x * self._x + self._y * self._y)
    end
end

class NamedPoint extends BasePoint
    public name: string
    function __init(self, x, y, name) ... end
end

local np = NamedPoint.new(..)
print(np:length()) -- how could this work if np didn't have _x or _y?
```

TenebrisNoctua:
> > This should normally fail. _x and _y should not remain on instances of ThreeDPoint, private should always mean private to a specific class, and never to any other class.
> 
> This proposal works nearly identically to how it works in C++, Java, and everything that's flowed from that. If private fields weren't available (but hidden) in derived classes, inheritance wouldn't even work for any class that has a private field.
> 
> ```luau
> open class BasePoint
>     private _x: number
>     private _y: number
>     function __init(self, x, y) ... end
>     function length(self)
>         return math.sqrt(self._x * self._x + self._y * self._y)
>     end
> end
> 
> class NamedPoint extends BasePoint
>     public name: string
>     function __init(self, x, y, name) ... end
> end
> 
> local np = NamedPoint.new(..)
> print(np:length()) -- how could this work if np didn't have _x or _y?
> ```

With `protected`. This example would work quite well with it. 

Personally, I don't complicate things by wanting to create a hidden field, to me, it is a worthless workaround for the fragile base class problem. In almost all of the cases I desired to use a hidden field, I only saw people (especially in Roblox codebases!) to find workarounds to continue violating it anyway.

InfraredGodYT:
> With `protected`. This example would work quite well with it.
> 
> Personally, I don't complicate things by wanting to create a hidden field, to me, it is a worthless workaround for the fragile base class problem. In almost all of the cases I desired to use a hidden field, I only saw people (especially in Roblox codebases!) to find workarounds to continue violating it anyway.

I agree with Noctua here. I also believe that instances of a class containing the private fields of the parent class is a bad idea when `protected` is usually used in this case. Inheritance without `protected` only amplifies the dissent against adding it to Luau in the first place as opposed to simply supporting composition.

nnullcolumn:
> Inheritance without `protected` only amplifies the dissent against adding it to Luau in the first place as opposed to simply supporting composition.

what does 'simply supporting composition' mean to you?

InfraredGodYT:
> > Inheritance without `protected` only amplifies the dissent against adding it to Luau in the first place as opposed to simply supporting composition.
> 
> what does 'simply supporting composition' mean to you?

Allowing class fields that are instances/objects of other classes and not supporting inheritance?

InfraredGodYT:
> I also believe that instances of a class containing the private fields of the parent class is a bad idea when `protected` is usually used in this case.

I misinterpreted what was going on in this RFC and thought `private` was going to function like `protected` does in other languages. My mistake.

Kenji-Shore:
> > > This should normally fail. _x and _y should not remain on instances of ThreeDPoint, private should always mean private to a specific class, and never to any other class.
> > 
> > 
> > This proposal works nearly identically to how it works in C++, Java, and everything that's flowed from that. If private fields weren't available (but hidden) in derived classes, inheritance wouldn't even work for any class that has a private field.
> > ```luau
> > open class BasePoint
> >     private _x: number
> >     private _y: number
> >     function __init(self, x, y) ... end
> >     function length(self)
> >         return math.sqrt(self._x * self._x + self._y * self._y)
> >     end
> > end
> > 
> > class NamedPoint extends BasePoint
> >     public name: string
> >     function __init(self, x, y, name) ... end
> > end
> > 
> > local np = NamedPoint.new(..)
> > print(np:length()) -- how could this work if np didn't have _x or _y?
> > ```
> 
> With `protected`. This example would work quite well with it.
> 
> Personally, I don't complicate things by wanting to create a hidden field, to me, it is a worthless workaround for the fragile base class problem. In almost all of the cases I desired to use a hidden field, I only saw people (especially in Roblox codebases!) to find workarounds to continue violating it anyway.

With what you are saying, wouldn't this mean that private fields would only be allowed to be accessed through private methods (if the method wasn't private, then it would be inherited by derived classes, except those derived class instances do not contain the private base class fields, hence error)? 

Also, what you are suggesting goes against the principle of encapsulation. Developers already use encapsulation in Lua with closures, so this is not anything new. Derived classes should not be able to manipulate data that is private to their base classes (because it creates opportunities for a derived class to interfere with its base class private fields in a way that the base class cannot anticipate). Base classes can offer methods to interface with its own private data and control the terms of the interaction.

TenebrisNoctua:
> > > > This should normally fail. _x and _y should not remain on instances of ThreeDPoint, private should always mean private to a specific class, and never to any other class.
> > > 
> > > 
> > > This proposal works nearly identically to how it works in C++, Java, and everything that's flowed from that. If private fields weren't available (but hidden) in derived classes, inheritance wouldn't even work for any class that has a private field.
> > > ```luau
> > > open class BasePoint
> > >     private _x: number
> > >     private _y: number
> > >     function __init(self, x, y) ... end
> > >     function length(self)
> > >         return math.sqrt(self._x * self._x + self._y * self._y)
> > >     end
> > > end
> > > 
> > > class NamedPoint extends BasePoint
> > >     public name: string
> > >     function __init(self, x, y, name) ... end
> > > end
> > > 
> > > local np = NamedPoint.new(..)
> > > print(np:length()) -- how could this work if np didn't have _x or _y?
> > > ```
> > 
> > 
> > With `protected`. This example would work quite well with it.
> > Personally, I don't complicate things by wanting to create a hidden field, to me, it is a worthless workaround for the fragile base class problem. In almost all of the cases I desired to use a hidden field, I only saw people (especially in Roblox codebases!) to find workarounds to continue violating it anyway.
> 
> With what you are saying, wouldn't this mean that private fields would only be allowed to be accessed through private methods (if the method wasn't private, then it would be inherited by derived classes, except those derived class instances do not contain the private base class fields, hence error)?
> 
> Also, what you are suggesting goes against the principle of encapsulation. Developers already use encapsulation in Lua with closures, so this is not anything new. Derived classes should not be able to manipulate data that is private to their base classes (because it creates opportunities for a derived class to interfere with its base class private fields in a way that the base class cannot anticipate). Base classes can offer methods to interface with its own private data and control the terms of the interaction.

Any field you'd need to use in that case would just be protected. Or you wouldn't make it `protected`, and just keep it to the class. 

`protected` doesn't violate encapsulation. It's a matter of how you'd use it. I personally think that base class private fields existing under derived classes is not a healthy relationship in inheritance, from my personal experience. 

A better argument would be to say that this behavior is quite the norm in many languages, and developers would expect to see it in Luau in one way or form. I propose it isn't necessary to make `private` contain this behavior, and we should remove it, or we implement it under a different access specifier to make it more explicit.

Kenji-Shore:
> `protected` doesn't violate encapsulation. It's a matter of how you'd use it. I personally think that base class private fields existing under derived classes is not a healthy relationship in inheritance, from my personal experience.

`protected` does violate encapsulation, though, in the sense that it allows for derived classes to modify base class fields in ways that the base class cannot anticipate. The whole point of encapsulation is allowing a class to provide a set interface and control access to its data. Getters and setters are a good example of this. There may be specific details of implementation that a base class wishes to keep tight control over, and does not want any derived classes to know or worry about. `protected` does not enable that, and with your suggestion, any functionality involving `private` would not work for inherited subclasses.

InfraredGodYT:
After revisiting this RFC, I don't think the need for leading underscores has been established, especially given prior art from languages such as C++ and C#. I’ll use the examples in the **Alternatives** section to explain why.

```luau
open class Base
    private x: number

    function compare(self, other)
        return self.x == other.x
    end
end

class Derived extends Base
    public x: number
end

local a = Base.new(...) -- :(
local b = Derived.new(...)

local res = a:compare(b)      -- Compares the private x declared by Base on both values (Base.x to Base.x)
local res2 = a:compare({x=5}) -- Type error
```

Because `compare` is declared in `Base`, both `self.x` and `other.x` should be lexically bound to the private `x` declared by `Base`. Accessing `other.x` therefore requires or infers that `other` is a `Base`.

Passing a `Derived` is valid because it is also a `Base` object, so `compare` reads the private `x` belonging to the `Base` portion of each object. This does not give methods declared in `Derived` access to the private `x` declared by `Base`; only methods declared in `Base` have that authority. The unrelated public `x` declared by `Derived` never participates in this lookup.

Passing `{x = 5}` is a type error because a table with a public property named `x` is not a `Base`. If the type error is ignored or the call comes from nonstrict code, the private field operation should perform a runtime brand check and fail rather than falling back to the table’s public `"x"` property.

The RFC’s second example demonstrates a problem with runtime selection, not with ordinary private field names:

```luau
open class BaseClass
    private x: number

    function __init(self, x)
        self.x = x
    end

    function equalTo(self, other)
        return self.x == other.x
    end
end

class ThirdClass -- extends BaseClass
    public x: number

    function test(self, other: BaseClass)
        if other:equalTo(self) then
            print("equal")
        else
            print("inequal")
        end
    end
end

local b = BaseClass.new(4)
local third = ThirdClass.new { x = 2 }

third:test(b)
```

The maintenance bug arises if `other.x` is interpreted dynamically using logic resembling:

```
if other is an instance of BaseClass then
    read the private x declared by BaseClass
else
    read the ordinary public property "x"
end
```

Under those semantics, changing `ThirdClass` to extend `BaseClass` changes what the same expression denotes at runtime. Before the change, `other.x` reads the public property; afterward, it reads the private field declared by `BaseClass`.

That runtime fallback is the source of the surprising behavior. Private field lookup does not need to work that way.

With lexical binding, both `.x` accesses in `BaseClass.equalTo` always denote the private `x` declared by `BaseClass`. This effectively constrains `other` to `BaseClass`:

```luau
function equalTo(self: BaseClass, other: BaseClass)
    return self.x == other.x
end
```

Before `ThirdClass` extends `BaseClass`, passing `self` to `equalTo` is invalid because `self` is not a `BaseClass`. After `ThirdClass` extends `BaseClass`, the call becomes valid and accesses the private `x` belonging to the `BaseClass` portion of both operands. Changing the inheritance relationship changes the call from invalid to valid; it never changes the meaning of `other.x`.

The compiler can lower a lexically resolved access directly:

```
GETPRIVATE receiver=other, expected_class=BaseClass, private_field="x"
```

The expected class and private field slot are determined at compile time. At runtime, the instruction only verifies that the receiver is a valid `BaseClass` instance; it does not decide whether `x` means a private field, public field, or table key.

If ordinary structural property access is intended instead, it can be written explicitly:

```luau
return self.x == other["x"]
```

Therefore, leading underscores and the special `._` operator are unnecessary. The enclosing class and its private declarations already provide enough information to bind private accesses lexically. The `private` modifier should establish access control, while ordinary field names and ordinary dot syntax should remain usable:

```luau
private x: number
self.x
other.x
```

skberkeley:
Hello! It sounds like you're proposing a similar point in the design space, where `.<privatefieldname>` is essentially the "glyph" that the compiler uses to infer that a private field access is happening. This is an entirely valid approach as well, but there are a few reasons why we're choosing not to take it:
- _All_ field accesses on objects (ie, instances of other classes and tables) with fields named `privatefieldname` will now have to do `obj[privatefieldname]`. This seems more restrictive than writing `._<fieldname>` and occasionally when a table has that key, `tbl["_<fieldname>"]`.
- Prefixing private fields with `_` makes it easier to tell when skimming code which fields are private and which are public, which can be useful when a class grows large.
- Adding new private fields has more downstream consequences, since naming collisions may require refactoring to use the bracket indexing notation.

TenebrisNoctua:
> Hello! It sounds like you're proposing a similar point in the design space, where `.<privatefieldname>` is essentially the "glyph" that the compiler uses to infer that a private field access is happening. This is an entirely valid approach as well, but there are a few reasons why we're choosing not to take it:
> 
> * _All_ field accesses on objects (ie, instances of other classes and tables) with fields named `privatefieldname` will now have to do `obj[privatefieldname]`. This seems more restrictive than writing `._<fieldname>` and occasionally when a table has that key, `tbl["_<fieldname>"]`.

Would it be possible to make this a restriction only in that enclosed function? Perhaps a syntax error could be raised that prompts the user to write `obj["field"]`. And in the other cases, it would be `obj.field`.

> * Prefixing private fields with `_` makes it easier to tell when skimming code which fields are private and which are public, which can be useful when a class grows large.

While this is a pretty good point, I would argue that it doesn't make a lot of difference when maintaining a class.

andyfriesen:
> > * Prefixing private fields with `_` makes it easier to tell when skimming code which fields are private and which are public, which can be useful when a class grows large.
> 
> While this is a pretty good point, I would argue that it doesn't make a lot of difference when maintaining a class.

I strongly disagree.

If private fields can have any name, adding a new private field to an existing class is quite the chore: You need to first search the whole class for any case where the field is also a public field of anything and rewrite that code to use bracket notation.  This is a bunch of work and results in a much larger diff than a single new field should require.  Further, if you forget to do this, it's not certain that type checking would fail!

You'd be well-served by prefixing all of your private fields to minimize the chance of such a conflict whether the language requires it or not.

TenebrisNoctua:
> > > * Prefixing private fields with `_` makes it easier to tell when skimming code which fields are private and which are public, which can be useful when a class grows large.
> > 
> > 
> > While this is a pretty good point, I would argue that it doesn't make a lot of difference when maintaining a class.
> 
> I strongly disagree.
> 
> If private fields can have any name, adding a new private field to an existing class is quite the chore: You need to first search the whole class for any case where the field is also a public field of anything and rewrite that code to use bracket notation. This is a bunch of work and results in a much larger diff than a single new field should require. Further, if you forget to do this, it's not certain that type checking would fail!
> 
> You'd be well-served by prefixing all of your private fields to minimize the chance of such a conflict whether the language requires it or not.

Well, this looks like a downside of a potential implementation rather than a concern with private vs public fields in a more general aspect. I never had a problem with this in C++, C# or any other object oriented language.

Let's say we had to swallow the downside of that potential implementation, I'd agree. It would be a chore to use the bracket notation. But I'm more or less talking from a viewpoint that we may not need to swallow such downside. 

Inheritance makes this really difficult though, with the "hidden" behavior of private. My solution to this on a local copy was to remove this behavior all-together and replace it with `hidden` access specifier for inheritance specifically.

skberkeley:
Hello! I think we've made clear why we prefer the current approach to the suggested alternatives, and I don't see any new points being raised at this time, so I'm going to go ahead and merge this proposal into the mega-RFC. If you have any further ideas you'd like to raise, please feel free to do so on the main mega-RFC thread!

PR #249, opened 2026-08-23
table.drop & linter improvements for table.remove

InfraredGodYT:
[Rendered](https://github.com/InfraredGodYT/rfcs/blob/table-drop/docs/table-drop.md)


InfraredGodYT:
My mistake, used an old fork. Ignore the previous commits.

docs/table-drop.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+However, `table.find` returns `nil` when it cannot find the element specified in the second argument of the function and
+`table.remove` defaults to removing the last element of a table when its second argument is either missing or `nil`. This creates the footgun of unintentionally removing the last element of the table unless you use an `if` statement.
+
+## Design
+
+First, I am proposing a standard library function for tables to remove elements by value rather than index:
+
+```table.drop(t: {V}, value: V, max: number?)```
```
nnullcolumn: someone might interpret `max` as a search distance, rather than a number of occurrences

docs/table-drop.md
```diff
@@ -0,0 +1,53 @@
...
+
+`t` refers to the table you're using, `value` is the value you want removed, and `max` is an optional argument specifying the limit of how many occurrences of `value` you want removed.. If `max` is `nil`, every occurrence of `value` is removed from the table. This function does not return anything due to the existence of `table.remove`.
+
+This function traverses front-to-back (index 1 to `#t`) to find indices where the table element at that index is equal to `value` (using `__eq` semantics), breaking early if `max` is specified and `max` elements have already been found. The indices of elements to remove are pushed in a last-in first-out manner (adding later matches in the table closer to the front of the stack), so that we can directly iterate the stack to remove indices back-to-front so that elements are removed in the correct order. This ensures that we remove the correct indices without worrying about shifting during array removal.
```
nnullcolumn: > This function traverses front-to-back (index 1 to `#t`)

is there a reason front-to-back was chosen over back-to-front or unspecified ordering?

EDIT: right, it needs to mirror `table.find`, which is known to linear search

> The indices of elements to remove are pushed in a last-in first-out manner (adding later matches in the table closer to the front of the stack), so that we can directly iterate the stack to remove indices back-to-front so that elements are removed in the correct order.

does the RFC need to specify this? it seems somewhat in the weeds on implementation details

docs/table-drop.md
```diff
@@ -0,0 +1,53 @@
...
+table.drop(t, "banana", 1) -- t is now {"apple", "banana"}
+```
+
+Next, I am proposing to add a linter rule warning the user about the `table.remove` footgun if their second argument is an optional variable and isn't explicitly `nil`.
+
+```luau
+local t = {"apple", "banana", "orange"}
+
+table.remove(t, nil) -- OK
+table.remove(t, table.find("banana")) -- Warning: Using optional variables as the 2nd arg for table.remove is a footgun. Use table.drop instead.
```
nnullcolumn: `table.find("banana")` is a type error, you need to pass the first argument

docs/table-drop.md
```diff
@@ -0,0 +1,53 @@
...
+```
```
nnullcolumn: couldn't a lint be done without the new library function? with this code:

```luau
table.remove(t, table.find(t, "banana"))
```

the linter could suggest a fix:

```luau
table.remove(t, assert(table.find(t, "banana")))
```

i find it likely this is what a lot of people *want* to do to begin with from this pattern

docs/table-drop.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Drawbacks
+
+While there are no compatibility concerns, this does add a new standard library function to the language.
+
+## Alternatives
+
+- Simply adding the linter rule would stop the footgun, but it would leave the verbose use of `if` statements as the only solution.
```
nnullcolumn: see earlier comment, you can do this with `assert`

nnullcolumn (commented):
oops forgot to submit lol

docs/table-drop.md
InfraredGodYT: My fault, typo

docs/table-drop.md
InfraredGodYT: What would be the best alternative name for that argument then? My mind goes to `limit`.

docs/table-drop.md
InfraredGodYT: Technically no, I just wanted to include `table.drop` as an alternative to the pattern. `table.remove(t, assert(table.find(t, "banana")))` looks pretty verbose.

docs/table-drop.md
InfraredGodYT: You're probably right.

docs/table-drop.md
InfraredGodYT: I'll address that in the next commit.

docs/table-drop.md
nnullcolumn: > Technically no, I just wanted to include `table.drop` as an alternative to the pattern. `table.remove(t, assert(table.find(t, "banana")))` looks pretty verbose.

in the past ive been known to shorten the transcript to `tremove(t, ass(tfind(t, key)))`

docs/table-drop.md
nnullcolumn: > What would be the best alternative name for that argument then? My mind goes to `limit`.

one option is to actually just add a search distance, so it'd be like `table.drop(t, value, count, start, finish)`

InfraredGodYT: Added the `start` and `finish` optional args as an alternative while renaming `max` to `count`.

PR #250, opened 2026-08-25
Stack library

IFMNB:
A library for managing the context of the corroutine execution. Available in English and Russian (on the different stack_lib_ru branch*).

[Rendered](https://github.com/IFMNB/rfcs/blob/stack_library/docs/stack-lib.md)


SeloTheGreat:
AFAIK, i am pretty sure all RFCs should be exclusively in english and not have alternate language counterparts. i don't think there is a written rule about it? but for the sake of conformity, you could remove the Russian counterpart of the English document.

IFMNB:
I checked it, and the English file is corrupted. I’ll fix it soon. I didn’t notice it while I was translating it (fixed now)

InfraredGodYT:
You should leave a link titled "Rendered" in the original message to the .md file of the RFC.

IFMNB:
I've been thinking about the library's security invariants for a while and came to the conclusion that it's not necessary to set a condition for `stack.mark()` to prohibit redefinition, but only if the label has been used. On the other hand, for this to work correctly, we need to enter 1 more function `stack.revive(markname: string)`. This still protects the code from implicit overrides and recursions, but if we use `stack.revive()` incorrectly, we may lose control. Therefore, most likely there will be a limit of 80 redefinitions per 1 label.

Example:

```luau

local f1
local function f0 ()
stack.mark("a0")
stack.revive("a0") 

f1()
end

f1 = function ()
stack.goto("a0")
end

f0()

```

This will lead to incorrect behavior. Such behavior must be stopped. In addition, I assume that only the part of the stack in which the label was declared has the right to call `stack.revive()` for this label.

nnullcolumn:
i wonder if there couldn't be a better name for this. i worry people might conflate 'stack' with a native stack datastructure

IFMNB:
> i wonder if there couldn't be a better name for this. i worry people might conflate 'stack' with a native stack datastructure

this library could be part of the coroutine API, but it is responsible for managing the execution context, not the coroutine object itself.

sooown:
i'm not sure if this RFC is human-written, but there are multiple issues with it regardless:
1. you are proposing an overengineered implementation of Lua's `goto`, which is explicitly marked as "won't implement" on the [Compatibility](https://luau.org/compatibility/) page.
2. the document leans in a lot on implementation details, which should be omitted because RFCs should only propose new features -- what files to modify is only the language developer's concern.
3. you yourself covered enough drawbacks that the cost outweighs the benefits.

IFMNB:
> i'm not sure if this RFC is human-written, but there are multiple issues with it regardless:
> 
>     1. you are proposing an overengineered implementation of Lua's `goto`, which is explicitly marked as "won't implement" on the [Compatibility](https://luau.org/compatibility/) page.

stack intentionally shares the same broad category of non-linear control flow as goto, but it has different semantics and a fundamentally different target model. Luau's decision not to implement lexical, statically-resolved goto does not imply that dynamic runtime addressing of execution contexts is undesirable or out of scope.

>     2. the document leans in a lot on implementation details, which should be omitted because RFCs should only propose new features -- what files to modify is only the language developer's concern.

Yes, it's in the document, but I find it useful for understanding a huge part of the disadvantages.

>     3. you yourself covered enough drawbacks that the cost outweighs the benefits.

This library solves complex issues. Paradoxically, the more complicated the operation, the more useful it can be. I also conducted surveys with regular scriptwriters, and some of them actually confirm that they do not need these operations. But this does not prove that the library itself is not needed, `bit32` solves a class of problems that could be implemented directly in luau and without adding it inside the language as a regular library.

This RFC was originally written in Russian, and ChatGPT was used to translate it.

sooown:
> stack intentionally shares the same broad category of non-linear control flow as goto, but it has different semantics and a fundamentally different target model. Luau's decision not to implement lexical, statically-resolved goto does not imply that dynamic runtime addressing of execution contexts is undesirable or out of scope.

doesn't matter, still makes control flow unnecessarily complex which makes certain optimizations either impossible or much harder to implement. making it dynamic is even worse than static


> Yes, it's in the document, but I find it useful for understanding a huge part of the disadvantages.

it's not useful and would have to be removed either way



> This library solves complex issues. Paradoxically, the more complicated the operation, the more useful it can be. I also conducted surveys with regular scriptwriters, and some of them actually confirm that they do not need these operations. But this does not prove that the library itself is not needed, `bit32` solves a class of problems that could be implemented directly in luau and without adding it inside the language as a regular library.

can you name some of those issues if it's so useful

PR #252, opened 2026-09-09
Move back to call syntax for class constructors

skberkeley:
Hi folks! After discussing with the team, we've come around to the perspective that requiring the use of `.new` to construct class instances is not worth the cost of reserving the name. We initially felt that forcing class instantiation to be consistent with existing Luau conventions was worth this cost, but now recognize that this is not the case. This approach still allows class authors to make `.new()` the happy path for construction by making the default constructor private (RFC for private constructors to come later).


InfraredGodYT:
**Our prayers have been answered**

gaymeowing:
It'd be nice if user defined constructors could allow for just using regular function call syntax, but otherwise no issues with this change as it'll make things less confusing as you're not defining an `__init` constructor to define a static `.new()` field.

skberkeley:
> It'd be nice if user defined constructors could allow for just using regular function call syntax, but otherwise no issues with this change as it'll make things less confusing as you're not defining an `__init` constructor to define a static `.new()` field.

Hello! Apologies if the wording was unclear, but the intention was for classes which define `__init` to also be instantiated with function call syntax.

medkipz:
It's to my understanding that, despite resembling a table call, this syntax isn't intended to _actually_ use a `NEWTABLE` instruction once its implemented, right?

skberkeley:
> It's to my understanding that, despite resembling a table call, this syntax isn't intended to _actually_ use a `NEWTABLE` instruction once its implemented, right?

I can't make guarantees about implementation details, but the current strategy is to add a `__call` metamethod to the class object itself.

PR #253, opened 2026-09-10
Vector library axis constants

gaymeowing:
[Rendered](https://github.com/gaymeowing/rfcs/blob/vector-library-axis-vectors/docs/vector-library-axis-constants.md)


whoamimaybeme:
Could you add a vector library constant that is `vector.create(1,1,0)` since it would be useful for working with 2d stuff.

I was told that I should ask you to add this to your RFC instead of making my own RFC because there was a math constant RFC that got updated with new stuff after it was initially made.

nnullcolumn:
> Could you add a vector library constant that is `vector.create(1,1,0)` since it would be useful for working with 2d stuff.
> 
> I was told that I should ask you to add this to your RFC instead of making my own RFC because there was a math constant RFC that got updated with new stuff after it was initially made.

this is fine, but you could always make a PR to the rfcs fork, and what would this be named?

gaymeowing:
Could this get merged? It’s been 2 weeks and its fairly straightforward.

vegorov-rbx:
No, it cannot be merged until it has been reviewed and approved (with potentially some new steps we plan to introduce).

PR #255, opened 2026-09-15
Propose match syntax

076c:
This is my first contribution/RFC.

[Rendered](https://github.com/076c/rfcs/blob/master/docs/match-syntax.md)


InfraredGodYT:
I would also like `match` statements, but there are a few issues I have with this RFC mostly involving the proposed syntax.

> Range pattern: Consists of two literals seperated with a hashtag (or the until keyword) and indicates a range of numbers to match

`#` is already used as the boundary operator, and using it to indicate a range doesn't read well. You could use guard patterns instead.

Fat arrows are also something I'd like to avoid since Luau usually avoids symbols for control flow in favor of keywords. This line looks especially ugly:

```luau
else if exp <= 60 => "Super Experienced",
``` 

Honestly, I'd prefer syntax more along the lines of:

```luau
local status = match playerHealth
    case 0 then "Dead",
    case playerHealth >= 1 and playerHealth <= 25 then do
        logStatusChange("Critical")
        "Critical"
    end,
    else "Overflow"
end
```

*You also left the "Feature name" title from the RFC template untouched....*

076c:
Yeah, I realized that shortly after I sent the RFC 🙂 For the range operator, `until` would be a great fit aswell since it's already a keyword and fits Luau's style. Your example fits pretty well but the 'case' keyword isn't really needed in my opinion. For alternatives to the fat arrow, you could either do `then` (as you said), `:` or `->` (which are all great fits in my opinion).

xHeptc:
idk if its just me, but the syntax doesn't fit luau at all

InfraredGodYT:
> idk if its just me, but the syntax doesn't fit luau at all

It's not just you. Fat arrows are probably something Luau will never introduce in its syntax.

076c:
> idk if its just me, but the syntax doesn't fit luau at all

I took my inspiration from Rust using fat arrows (because that's what made the most sense to me), but @InfraredGodYT is completely right here; Luau will most likely use a keyword instead or a symbol we're already familiar with instead of introducing the fat arrow.

076c:
I've also updated the syntax to use the `if` keyword instead: it would allow an expression of any type to be matched and it wouldn't be as confusing to learn. It also fits the Luau syntax a little more in my opinion.

xHeptc:
now its just an if expression with the support of until keyword? doesn't really make sense, cause then it would imply a statement has same rules

I feel like implementing if syntax is bad due to how conditions are checked:
```luau
if 0 then
if 1 then
``` just looks really weird and anyone initially would assume first condition always run
not the best syntax designer but something like:
```lua
local value = 0
local result = in value do
    match 0 then "success"
    match 1 then "warning"
    match 2, 3 then "fatal"
end
```
or something else

076c:
> now its just an if expression with the support of until keyword? doesn't really make sense, cause then it would imply a statement has same rules
> 
> 
> 
> I feel like implementing if syntax is bad due to how conditions are checked:
> 
> ```luau
> 
> if 0 then
> 
> if 1 then
> 
> ``` just looks really weird and anyone initially would assume first condition always run
> 
> not the best syntax designer but something like:
> 
> ```lua
> 
> local value = 0
> 
> local result = in value do
> 
>     match 0 then "success"
> 
>     match 1 then "warning"
> 
>     match 2, 3 then "fatal"
> 
> end
> 
> ```
> 
> or something else

The problem with this is that you have an unnecessary keyword in the beginning of every match, but I find the `in` keyword pretty useful here.

xHeptc:
> > now its just an if expression with the support of until keyword? doesn't really make sense, cause then it would imply a statement has same rules
> > I feel like implementing if syntax is bad due to how conditions are checked:
> > ```luau
> > if 0 then
> > 
> > if 1 then
> > 
> > ``` just looks really weird and anyone initially would assume first condition always run
> > 
> > not the best syntax designer but something like:
> > 
> > ```lua
> > 
> > local value = 0
> > 
> > local result = in value do
> > 
> >     match 0 then "success"
> > 
> >     match 1 then "warning"
> > 
> >     match 2, 3 then "fatal"
> > 
> > end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or something else
> 
> The problem with this is that you have an unnecessary keyword in the beginning of every match, but I find the `in` keyword pretty useful here.

using `if` creates syntax problems with if statements and if expressions, due to how conditions are checked
i prefer a new keyword just for that

PR #256, opened 2026-09-18
Non-returning functions

lucaskiforishin:
[Rendered](https://github.com/lucaskiforishin/luau-rfcs/blob/non-returning-functions/docs/non-returning-functions.md)

Proposal to treat calls with a resulting return type of `never` as non-returning in control-flow analysis.


Vighnesh-V:
Would this behaviour also apply to functions that have their return types inferred as never? A possible issue is that during type inference, control flow no longer exists, so this information is never usable.

lucaskiforishin:
Yeah, I intended for the rule to be based on the resulting return type of the call, regardless of whether `never` was explicitly annotated or inferred, but i didn't consider that the inferred return type might not be available when control flow is being analyzed. Would that mean the behavior would need to be limited to cases where the never return type is already available, rather than needing to be inferred from the function body?

PR #257, opened 2026-09-18
Require top-level annotations in strict mode

andyfriesen:
[Rendered](https://github.com/luau-lang/rfcs/blob/top-level-annotations/docs/require-annotations-in-strict-mode.md)


docs/require-annotations-in-strict-mode.md
```diff
@@ -0,0 +1,89 @@
...
+
+The benefits of top-level annotations tend to trickle downward into the program: top-level types act as seeds that help us to infer the interior of each function with much higher fidelity.
+
+## Design
+
+In strict mode, we add a simple extra check when we encounter functions.
+
+We will warn on missing annotations on a function if it is declared using a function statement (eg `function foo`) or if it is a method or field of a class.
+
+We will *not* warn on unannotated parameters for lambdas or functions that appear within nested scopes.
```
nnullcolumn, line 65: this feels a tad jank, i think the rule should be `export function bar` and `function foo.bar()` with no ties to scoping etc. this keeps it easy to substitute lambdas with regular named functions for debugging while keeping most of the 'oo' cases pretty clear

docs/require-annotations-in-strict-mode.md
```diff
@@ -0,0 +1,89 @@
...
+
+We will also not warn on a missing return type annotation if inference indicates that a function returns 0 values.
```
nnullcolumn, line 67: what about functions which return `nil`?

docs/require-annotations-in-strict-mode.md
```diff
@@ -0,0 +1,89 @@
...
+
+Note that we leave developers with a simple way to explicitly opt out if they would truly prefer not to annotate something:
+
+```luau
+export const myFunction = function(arg, arg, arg)
```
nnullcolumn, line 72: i would vastly prefer this feature to just be a lint.

docs/require-annotations-in-strict-mode.md
```diff
@@ -0,0 +1,89 @@
...
+end
+```
+
+Luau already offers capable type inference, so it's very easy for Luau to include the inferred type of the function in the warning message.  This makes it straightforward for the developer (or their text editor) to fill in the needed annotation.
+
+A future improvement would be to extrapolate a suggestion from the inferred type.  For instance, if Luau witnesses an inferred type along the lines of `{ read find: ..., read gsub: ... }`, inference could suggest `string`.
+
+## Drawbacks
+
+The obvious one is that we're asking developers to put a bit of extra work in.  To shore this up, we'd like to offer editor tooling to automatically fill type annotations in for the user.
```
nnullcolumn, line 83: luau-lsp already has click to fill inline type hints, would this be any different from that? (aside from being available in roblox)

nnullcolumn (commented):
i like this rfc, i've left a few small nits & questions

docs/require-annotations-in-strict-mode.md
SeloTheGreat: i agree with this feature being a lint.
IMO it would be easier to manage, especially `.config.luau`/`.luaurc` and `--!nolint NAME` <sub>(...possibly a `@nolint` function attribute in the future 👀)</sub> would let you manage this feature in a fine-grained way without having to compromise on how you structure your top level functions.

though a type error would be fine, but i am most certainly in favor of it being a lint

docs/require-annotations-in-strict-mode.md
nnullcolumn: been thinking about this a little more; there are some really annoying cases with return annotations. for example, for functions which simply return functions, you're forced to annotate the function twice. although this can be fixed trivially in a way left as an exercise to the reader, these lead me to think required return annotations should be explored separately. plus, they shouldn't even have big blowup with generics or anything like that.

docs/require-annotations-in-strict-mode.md
aatxe: The entire point being considered here is to make annotating top-level functions part of the specification of strict mode. Making it a lint would miss the point IMO, and would probably just lead to a whole lot of people doing `--!nolint RequiredTypeAnnotations` or whatever in every file. To this point, we have not been designing a type system with configurable error messages, and I don't see a compelling argument to depart from that.

docs/require-annotations-in-strict-mode.md
aatxe: That's exactly the feature that is being described, yeah.

docs/require-annotations-in-strict-mode.md
nnullcolumn: In the RFC:
> Note that we leave developers with a simple way to explicitly opt out if they would truly prefer not to annotate something

this gave me the impression its something you guys want people to be able to opt out of. If that's incorrect I'm sorry for the confusion, but also like.. I want to be able to opt out of this. It's annoying enough as-is to switch really old games from nonstrict to strict, this just adds another barrier to that

edit: of course i support top level annotations, but like.. i'd rather a given script be strict rather than nonstrict. annotations are added gradually in circumstances like that

aatxe: It's more a consequence of the particularities of the rules here than a "desired" methodology for opting out of it.

